PR #4, opened 2022-07-27
Reading initial mass from headers in singlestar grids

kkovlakas:


ssbvr (approved):
I tested the PR on single grids and it solves the issue.

devimisra (approved):

PR #20, opened 2022-11-25
wrap MD5 verification in data_download in a try statement

ssbvr:
the rea read() function seems to fail on macOS but not on linux


tassos25 (approved):
Seems good to me.

PR #23, opened 2022-12-07
GZIP support for files in MESA grids

kkovlakas:
In this PR I include changes in routines that open MESA files, so that they support GZIP format. Notes:
- All GZIP-ed files should follow the same conventions but with an additional extension (`.gz`), e.g., `history.data.gz`. 
- The code expects that only the plain text or the `.gz` file exist in the folder. There is no guarantee in the behavior of the code if both files coexist in the same folder (but we can test to avoid surprised).

Testing:
- It has been tested after zipping two types of files with the following command:
```find mesagrid_gz \( -name 'history.data' -o -name 'binary_history.data' \) -exec gzip -1 --verbose {} \;```
- The HDF5 files after creating the grid from the original or compressed files are identical at the end (tested with `diff`).
- Note that the HDF5 files might differ if you use separate folders for the runs for testing purposed (because the original MESA run dirs are also saved in the PSyGrid objects).
- A more comprehensive test with the full grids (single or binary), and their usage, would show if the grids are as expected.

### CODE REVIEW

File: posydon/grids/io.py [hunk 1]
```diff
@@ -119,45 +126,51 @@ def read_posydon_format(self):
...
-            elif file == "out.txt" and self.binary:
-                self.out_txt_path = os.path.join(self.path, file)
-            elif file == "out_star1_formation_step0.txt" and not self.binary:
-                self.out_txt_path = os.path.join(self.path, file)
+            fullpath = os.path.join(self.path, file)
+            if file in ["binary_history.data", "binary_history.data.gz"]:
+                self.binary_history_path = fullpath
+            elif file in ["final_star1.mod", "final_star1.mod.gz"]:
+                self.final_star1_path = fullpath
+            elif file == ["final_star2.mod", "final_star2.mod.gz"]:
```
astroJeff: `file in ["final_star2.mod", "final_star2.mod.gz"]`?

posydon/grids/io.py
```diff
@@ -119,45 +126,51 @@ def read_posydon_format(self):
...
+                self.final_star1_path = fullpath
+            elif file == ["out.txt", "out.txt.gz"] and self.binary:
```
astroJeff: `file in ["out.txt", "out.txt.gz"]`?

astroJeff (commented):
I found a couple minor things

astroJeff:
We need to test this with either v1 grids or v2 grid samples - goes hand in hand with data transfer

astroJeff (approved):

ssbvr (approved):

PR #30, opened 2023-01-31
Save profiles

astroJeff:
I have added the option to the .ini file to save mesa profiles. In reality it wraps the various rm `LOGS/profiles*` commands in an `if`-statement that is by default set to false. This can be set to true by adding the line `keep_profiles = True` to the .ini file, at the very end under the `[run_parameters]` setting.


astroJeff:
I have tested this on my laptop, and I don't see how it would change for runs on the cluster. It also is built to be backwards compatible, so there should be no problem with the runs currently on the cluster (i.e., if one does not specify the `keep_profiles` argument, it by default is set to `False`, which is the current behavior.

PR #35, opened 2023-02-20
isolated evolution step 

ezapartas:
New step to have evolution of isolated single stars (i.e. initially single stars, disrupted companions and mergers) based on detached step


posydon/binary_evol/DT/step_detached.py
```diff
@@ -599,188 +621,91 @@ def get_mist0(self, star, htrack):
...
-                )
-                if (np.abs(sol.fun) > tolerance_mist_integration
+            '''
+            ####This is for He-star 
+            '''
+                if (np.abs(sol.fun) > tolerance_matching_integration
                         or not sol.success):
-                    sol = minimize(
-                        lambda x: (
-                            (sc_he_core_mass_He.transform(
```
ezapartas, line 825: Hello @ZepeiX . With @kasdaglie we need you input on this change of trying to slowly implement a step of stars that are doing an isolated evolution (initially single stars, disrupted binaries or mergers) based on the detached step. For now we are only focusing on the disrupted binaries.

We tried to generalize the detached step to accept as an input the parameters that will be used for the matching, and not hard coded. We saw that you had some alternative matching trials if the first match is not working. I know that you have tested it and you have come up with them as optimal at least at that point. For now we do not do it, we only try to do the first matching trial that was there. Do you think we should implement the alternative ones? Specifically, we saw that you sometime change from matching the absolute mass to the relative mass, as in this case (dividing by the mass value itself). Can you suggest which of these to keep and why?

ZepeiX: Now the known issues of failed matching include:
1. MS stars after MT has a smaller radius so that we match it again without radius
2. An almost stripped post-MS star cannot be matched within the single H-grid while it is more like a He star and we match it to the He-grid
3. Some He stars cannot be matched within single He-grid because when they got stripped, they have evolved a bit inside which cannot be reproduced with He-MS. But we found they are more similar to evolved H stars so that we match them in H-grid.

mkruckow:
The new steps look to be still incomplete:
- They, e.g. step_disrupted, don't do anything
- Loading a lot of stuff, which isn't needed, e.g. orbital_period_from_separation for a single star?
- Is there a reason, why the `path` is excluded from some of the `super().__init__()`?

ezapartas:
Let me explain a bit more here the general structure for the isolated step, I think @mkruckow it will answer some of your points.

The idea is to create an isolated_step which will use step_detached as a super class. In isolated_step, we manually make the binary very wide and we want to evolve it as a detached binary but without evolving the binary orbit for mass loss, tides, magnetic braking and GW radiation, only for the evolution of the isolated star itself (change of spin due to mass loss and expansion/contraction). That's the reason that in the isolated_step:
       do_wind_loss=False,
        do_tides=False,
        do_gravitational_radiation=False,
        do_magnetic_braking=False
        (but  do_stellar_evolution_and_spin_from_winds=True as in detached_step)
        
Now the isolated_step serves itself as a super class for 3 different evolutionary substeps:
- Disrupted_step
- Initially_single_star_step
- Merged_step
This means that the isolated step is never invoked in the flow. The above sub-steps call the isolated_step which calls the detached_step.

In this PR, we only worked in the disrupted_step which seems the easiest to be already implemented in the structure of POSYDON. Initially_single and merged do not work yet, indeed. The idea is to work on them after that PR, they are already added in the flow, but if people have concerns we can take them out of the flow in this PR and re-introduce them in another PR.

Now I know that right now disrupted_step is not doing much more compared to its super class, the isolated step, but it is needed to be an extra, different step in order to be able to **change the single star grid assigned for this step** during the initialization of the POSYDON population, as in the future we aim to have other grids specifically made for mass accretors. In the same way, mergers could also have a different single star grid in the future.

posydon/binary_evol/DT/step_detached.py
```diff
@@ -599,188 +621,91 @@ def get_mist0(self, star, htrack):
...
-                        if i not in self.root_keys:
-                            raise Exception("Expected matching parameter not "
-                                            "added in the MIST model options.")
-                    x0 = get_root0(MESA_label, posydon_attribute,
+                    list_for_matching = self.list_for_matching_HMS
+            elif star.state in LIST_ACCEPTABLE_STATES_FOR_postMS:
+                list_for_matching = self.list_for_matching_postMS
+                
+            elif star.state in LIST_ACCEPTABLE_STATES_FOR_HeStar:
+                list_for_matching = self.list_for_matching_HeStar
```
ZepeiX, line 699: Because for each kind of state the alternative matching require different `list_for_matching` and `htrack`, we create a function with input variable 'list_for matching' and 'htrack' to return the final solution 'sol'. 
And we call it here for different star state. If  np.abs(sol.fun) > tolerance_matching_integration or not sol.success, we change the list_for_matching and htrack.

ZepeiX (dismissed):

posydon/binary_evol/DT/step_detached.py
ezapartas: @kasdaglie Maybe we can implement the above suggestion from Zepei for the alternative matchings

astroJeff (dismissed):

ssbvr:
The PR is changing the `grid_params/POSYDON-MESA-INLISTS` submodule tagged commit. Could you unstage those changes?

posydon/binary_evol/DT/step_disrupted.py
```diff
@@ -0,0 +1,84 @@
...
+LIST_ACCEPTABLE_STATES_FOR_HMS = ["H-rich_Core_H_burning"]
+LIST_ACCEPTABLE_STATES_FOR_HeMS = ["stripped_He_Core_He_burning"]
+
+LIST_ACCEPTABLE_STATES_FOR_POSTMS = STAR_STATES_H_RICH.copy()
+[LIST_ACCEPTABLE_STATES_FOR_POSTMS.remove(x) for x in LIST_ACCEPTABLE_STATES_FOR_HMS]
+
+LIST_ACCEPTABLE_STATES_FOR_POSTHeMS = STAR_STATES_HE_RICH.copy()
+[LIST_ACCEPTABLE_STATES_FOR_POSTHeMS.remove(x) for x in LIST_ACCEPTABLE_STATES_FOR_HeMS]
+
+class disrupted_step(isolated_step):
```
ka-rocha: I would suggest changing the naming of the classes to match CamelCase convention so we are consistent throughout the code (e.g. `DisruptedStep`, `InitiallySingleStarStep`)

posydon/popsyn/binarypopulation.py
```diff
@@ -830,46 +837,94 @@ def draw_initial_binary(self, **kwargs):
...
-        star2_params = dict(
-            mass=m2,
-            state="H-rich_Core_H_burning",
-            metallicity=Z,
-            center_h1=X,
-            center_he4=Y,
-        )
+
+
+        if random.random()<self.binary_fraction:
```
ka-rocha: I would suggest a different way to generate binaries given a binary fraction. For this PR maybe we can use it and change later but I would like to have a more detailed discussion. And another separate note - using a random process like this which is not using the global seed with numpy will not be reproducible.

ka-rocha: And to suggest an actual fix if you want to keep it as is:
```python
if self.RNG.uniform() < self.binary_fraction:
```

posydon/popsyn/population_params_default.ini
```diff
@@ -116,6 +116,12 @@
     # True False
   verbose = False
     # True False
+  
+[step_disrupted]
+  import = ['posydon.binary_evol.DT.step_disrupted','disrupted_step']
+  import = ['posydon.binary_evol.DT.step_isolated',isolated_step']
```
ssbvr: This line should be deleted, as one should only import one step at a time. Also, does this step have inputs? Does it take the same one as of the detached step? If yes, comment here in needed, else you should define them both in the step initialisation and here.

From the changes done in the flow chart, the isolated_step is not part of the default setup, right?

posydon/binary_evol/DT/step_detached.py
```diff
@@ -605,328 +676,219 @@ def get_mist0(self, star, htrack):
...
+            def square_difference(x):
+                result = 0.0
+                for (MESA_label, posydon_attr,
+                     colscaler, scale_of_that_MESA_label) in zip(
+                         MESA_labels, posydon_attributes, colscalers, scales):
+                    single_track_value = scale_of_that_MESA_label.transform(
+                        get_track_val(MESA_label, htrack, *x))
+                    posydon_value = scale_of_that_MESA_label.transform(
+                        posydon_attr)
+                    if MESA_label == "center_he4":
```
ZepeiX: Do we still keep the relative value for center_he4?

ezapartas: I will comment it out

posydon/binary_evol/DT/step_detached.py
```diff
@@ -605,328 +676,219 @@ def get_mist0(self, star, htrack):
...
-                            - star.total_moment_of_inertia
-                        ) / star.total_moment_of_inertia
+                    elif star.state in LIST_ACCEPTABLE_STATES_FOR_postMS:
+                        star.htrack = False
+                    htrack  = star.htrack
+                    x0 = get_root0(
+                        MESA_label, posydon_attribute, htrack, rs=rs)
+                    # bnds = ([m_min_H, m_max_H], [0, None])
+                    sol = minimize(square_difference,
+                                   x0, method="TNC", bounds=bnds)
```
ZepeiX: Since we change the grids, we should update **list_for_matching** for new MESA_label, posydon_attribute,  bnds as well. Just like the code above in 2nd alternative matching.

posydon/binary_evol/DT/step_detached.py
```diff
@@ -936,79 +898,114 @@ def __repr__(self):
...
-            secondary = binary.star_2
-            secondary.htrack = False
-            primary.htrack = False
-            primary.co = False
-        else:
-            raise Exception("States not recognized!")
-
-        if not primary.co:
-            m1, t1 = get_mist0(primary, primary.htrack)
-        m2, t2 = get_mist0(secondary, secondary.htrack)
```
ZepeiX, line 1004: This is actually to update htrack if the binary goes to the 3rd alternative matching. 
For the function get_star_data below, the input htrack is not updated if we call the matching function inside it. 
There must be a better way to deal with the binaries where we change the htrack.

ezapartas: I try to solve it in the next commit [901590a]

posydon/binary_evol/CE/simulationproperties.py
astroJeff, line 1: This file should not exist here. Let's remove it and then the PR should be ready to accept

astroJeff (approved):
Let's get this thing accepted!

ka-rocha (approved):

PR #39, opened 2023-02-21
Fixes to orbital_kick in step_SN

chase-kimball:
The current version of orbital_kick has an inconsistent coordinate system. It mostly follows  *Wong, T.-W., Valsecchi, F., Fragos, T., & Kalogera, V. 2012, ApJ, 747, 111*, which references the coordinate system in *Kalogera, V. 1996, ApJ, 471, 352*. However, the latter is restricted to circular preSN orbits. In places, the current code makes the assumption that the preSN relative velocity lies along the Y axis (in this reference frame, possible only if the orbit is circular), and treats theta as both the angle between the kick and the Y axis and the angle between the kick and the preSN relative velocity. This change redefines the kick angles to explicitly be relative to the preSN relative velocity, and propagates the changes to computing Apost, epost, Vsys, and tilt. Note: *the old code performed correctly for circular preSN orbits*


chase-kimball:
@mkruckow My latest commit fixes the issue with cos_psi that we discussed

mkruckow (dismissed):

tassos25 (dismissed):
Seems good to me.

astroJeff (approved):

ezapartas (approved):

PR #45, opened 2023-03-15
Small changes in common function and matching post-MS to He in detached step

ZepeiX:


ZepeiX:
BH spin evolution in step_MESA

posydon/binary_evol/DT/step_detached.py
```diff
@@ -696,6 +696,49 @@ def get_mist0(self, star, htrack):
...
+                        if i not in self.root_keys:
+                            raise Exception("Expected matching parameter not added"
+                                            " in the MIST model options.")
+                    x0 = get_root0(
+                        MESA_label, posydon_attribute, star.htrack, rs=rs)
+                    bnds = ([m_min_He, m_max_He], [0, None])
+                    sol = minimize(
+                        lambda x: (
+                            (sc_he_core_mass_He.transform(
+                                get_track_val(MESA_label[0], htrack, *x))
```
mkruckow: Should it be "star.htrack" instead of "htrack" here?

posydon/binary_evol/DT/step_detached.py
```diff
@@ -696,6 +696,49 @@ def get_mist0(self, star, htrack):
...
+                             - sc_he_core_mass_He.transform(
+                                 posydon_attribute[0]))
+                            / sc_he_core_mass_He.transform(
+                                 posydon_attribute[0])) ** 2
+                        + (sc_center_he4_He.transform(
+                            get_track_val(MESA_label[1], htrack, *x))
```
mkruckow: Should it be "star.htrack" instead of "htrack" here?

posydon/binary_evol/DT/step_detached.py
```diff
@@ -696,6 +696,49 @@ def get_mist0(self, star, htrack):
...
+                           - sc_center_he4_He.transform(
+                                 posydon_attribute[1]))
+                        ** 2 + (sc_log_R_He.transform(
+                            get_track_val(MESA_label[2], htrack, *x))
```
mkruckow: Should it be "star.htrack" instead of "htrack" here?

posydon/utils/common_functions.py
```diff
@@ -575,7 +575,7 @@ def bondi_hoyle(binary, accretor, donor, idx=-1, wind_disk_criteria=True,
                 slope = (3.7 - 3.25) / (-2.5 + 5.25)
             else:
                 slope = (3.25 - 3.75) / (-5.25 + 7.25)
-            v_wind = 10 ** (slope * lg_mdot[i] + 3.25 + 5.25 * slope) * 1000
+            v_wind[i] = 10 ** (slope * lg_mdot[i] + 3.25 + 5.25 * slope) * 1000
```
mkruckow, line 578: This won't work, because "v_wind" is a float not a list.
In the old version, it take the value of the last index, where the condition (surface_h1[i] < 0.4 and Teff[i] > 1.0e4) is fulfilled.

devimisra (dismissed):

posydon/utils/common_functions.py
ZepeiX: v_wind is an np.array. m and radius in v_esc are array

posydon/utils/common_functions.py
```diff
@@ -933,20 +933,22 @@ def inspiral_timescale_from_orbital_period(star1_mass, star2_mass,
     return T_merge
 
 
-def spin_stable_mass_transfer(star_mass_preMT, star_mass_postMT):
+def spin_stable_mass_transfer(spin_i, star_mass_preMT, star_mass_postMT):
```
mkruckow, line 936: On a quick search, I didn't found where in this function is called in the code. But I guess, you'd need to update all calls of that function when you add a new argument.

ZepeiX: yes, this function is only called in my latest change above in mesa_step

ezapartas:
@ZepeiX this is highly connected with #35 where with @kasdaglie to generalize the matching by not hard-coding each time the parameters used for matching (e.g. mass, log_R, center_he4, etc) but providing a list with them for each case (MS, postMS, HeStar). I know that you have tested and implemented alternative matching sequences if it fails. I think it would be cleaner to try to implement these alternative matching in the generalized case, and not hard-coded. What do you think?

posydon/binary_evol/DT/step_detached.py
ZepeiX: There must be a better way to organize it. Right now the htrack here doesn't affect the matching because the function is called twice latter. The first time star.track becomes False and the returned solution is not used. Then the second  time htrack becomes the same with star.htrack and the solution is used.

mkruckow: Because we have two independent variables `htrack` and `star.htrack`, there should be a reason for them to not be the same. Hence, please check when they are not to same and which one is the one which should be used here in the future.
I get to the question, because in the rest of your changes you use `star.htrack`, thus in terms of consistency I was wondering, whether there is a reason to not use it here but use `htrack` instead.

mkruckow (approved):

astroJeff (approved):

PR #46, opened 2023-03-22
Fixes in psy grid

mkruckow:
Main task:
- [x] fix adding of abundance values (lines 927-931 -> 927-931): don't write on all values, only on current run

Side tasks:
- [x] modify the copy of the initial and final values to be done in a safer way (lines 1099-1104 -> 1099-1112)
- [x] fix typo (line 1307->1315): `profile1` -> `profile2`
- [x] fix issue with inconsistent properties introduced in PR#36 (line 1966-1967 -> 1974-1975): add `accept_missing_profile` to list


tassos25 (approved):
looks good to me.

astroJeff (approved):

PR #57, opened 2023-04-14
Updating modules to newest versions for python 3.11

kasdaglie:
Starting the migration to Python 3.11. We updated setup.py installation was successful but couldn't run a population.  It failed when loading the interpolation files in pickle.load() 
` self.interpolators = pickle.load(f)`
`ModuleNotFoundError: No module named 'sklearn.neighbors.regression'`
This migration will require us to retrain the interpolators because sklearn change their module structure.

### DIFF
--- a/posydon/binary_evol/SN/step_SN.py
+++ b/posydon/binary_evol/SN/step_SN.py
@@ -1236,7 +1236,7 @@ class StepSN(object):
                     # if key is 'nearest_neighbour_distance':
                     #     setattr(binary, key, ['None', 'None', 'None'])
                 binary.separation = new_separation
-                if binary.state is not "disrupted":
+                if binary.state != "disrupted":
                     binary.state = "detached"
                 binary.event = None
                 binary.time = binary.time_history[-1]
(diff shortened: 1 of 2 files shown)


astroJeff:
Check to see if pytables is used in POSYDON. If so it should be added to setup.py

kasdaglie:
MESA grids run successfully. Run a small  population of 100 binaries that had no issues. To use this updated version of POSYDON one will need to have the new interpolators that are trained with the new version of python and a new conda enviroment with python 3.11.

kasdaglie:
> Check to see if pytables is used in POSYDON. If so it should be added to setup.py

We checked and didn't found any instances of pytables in the POSYDON except from some .yaml files

astroJeff:
This PR should be ready to accept.

ssbvr (approved):
I merged the development branch in this PR, fixed some "is not" Syntax Warnings in step SN and tested the populaiton synthesis code maiking sure that it outputs the same results as before. Also, we might want to consider in the future to add the `requirements.txt` and the `enviromnet.yml` file to allow the use to set up pip env or conda env directly after cloning the repo. Once this PR is accepted I will move the `POSYDON_3.11/POSYDON_data` to the `POSYDON_GRIDS/POSYDON_popsynth_data/v2/POSYDON_data` directory.

astroJeff (approved):

PR #58, opened 2023-04-15
adding profile interpolation

elizabethteng:
see posydon/interpolation/profile_interpolation.py. Includes funcitonality for Density and H mass fraction profiles.


posydon/interpolation/profile_interpolation.py
```diff
@@ -0,0 +1,553 @@
...
+        """Extracts profile data from '.h5' grid files and saves to file.
+        Args:
+            train_path (str) : path/name of '.h5' file for training data.
+            valid_path (str) : path/name of '.h5' file for testing/validation data.
+            profile_names (array-like) : list of profile quantities to be extracted.
+        """
+        self.names = profile_names
+        
+        try:
+            print(valid.final_values["S1_state"][0])
```
mkruckow: Did I missed it, or is the variable `valid` not defined before that try at all? It will get set in line 45.

elizabethteng: yes, you're right! thanks for catching that

posydon/interpolation/profile_interpolation.py
```diff
@@ -0,0 +1,553 @@
...
+        Returns:
+            scalars (array-like) : dictionary containing initial m1, m2, p and final s1_state, final_mass.
+            profiles (array-like) : all N specified profiles, shape (N,200).
+        """
+        # open individual run as a DataFrame
+        df = pd.DataFrame(grid[ind]['final_profile1'])
+        df = df.sort_values(by="mass")
+        df = df.reset_index(drop=True)
+
+        # grab input values, final star 1 state, final star 1 mass
```
mkruckow, line 95: Minor: If you need the values only for putting them in the dictionary, there is no need to define other variables first (this only wastes resources).

posydon/interpolation/profile_interpolation.py
```diff
@@ -0,0 +1,553 @@
...
+        scalars = {"m1":m1,
+                   "m2":m2,
+                   "p":p,
+                   "s1_state":state,
+                   "final_mass":final_mass}
+
+        # grab output vectors, interpolate to normalize
+        profiles=np.zeros([len(self.names),200])
+        for i,prof in enumerate(self.names):
+            f = interp1d(df['mass']/final_mass,df[prof],fill_value="extrapolate")
```
mkruckow: You may want to check first, that the `df` contains the column `prof` and add a catch if it is not there.

posydon/interpolation/profile_interpolation.py
```diff
@@ -0,0 +1,553 @@
...
+        self.profiles = []
+        training_failed = []
+        
+        for i in range(len(train)):
+            try:
+                scalars,profiles = self.scrape(train,i)
+                self.scalars = self.scalars.append(scalars,ignore_index=True)
+                self.profiles.append(profiles)
+            except:
+                training_failed.append(i)
```
mkruckow, line 75: `training_failed` and similarly `testing_failed` is created, but never used. I guess, you planed to raise a warning at the end of the initialization, telling which runs in the psygrid failed. Please, add this warning output.

posydon/interpolation/profile_interpolation.py
```diff
@@ -0,0 +1,553 @@
...
+            self.valid_profiles (array-like) : M profiles for each of N testing binaries, shape (N,M,200).
+            self.valid_scalars (array-like) : DataFrame containing initial and final conditions for N testing binaries.
+        """
+        with open(filename, 'rb') as f:
+            myattrs = pd.read_pickle(f)
+            for key in myattrs:
+                setattr(self, key, myattrs[key])
+        self.profiles = np.array(self.profiles)
+        self.valid_profiles = np.array(self.valid_profiles)
+        return self.profiles,self.scalars,self.valid_profiles,self.valid_scalars
```
mkruckow: I'm not sure, whether it is a good idea to return here this 4 parts. Wouldn't is be better to return `self`, in this way you wouldn't discard other key (you may add in the future). And you won't need to create new objects on the fly.
But I see, where you called it you don't use the return value at all.

posydon/interpolation/profile_interpolation.py
```diff
@@ -0,0 +1,553 @@
...
+        m1 = grid.initial_values["star_1_mass"][ind]
+        m2 = grid.initial_values["star_2_mass"][ind]
+        p = grid.initial_values["period_days"][ind]
+        state = grid.final_values["S1_state"][ind]  
+        final_mass = grid.final_values["star_1_mass"][ind]
+        scalars = {"m1":m1,
+                   "m2":m2,
+                   "p":p,
+                   "s1_state":state,
+                   "final_mass":final_mass}
```
mkruckow: Should we call `final_mass` better `reference_mass` or `unit_mass`? The term `final_mass` won't tell which star's final mass it is, while we do have two final star masses.

mkruckow:
I talked to Simone, he said, that we need the angular momentum profile to be well for calculating the spins. Could you please think about adding a feature to get a new profile from other profiles.
The problem is the following: Let's define a function `f(a,b)`, which calculates a quantity based on `a` and `b`. Furthermore, we have the interpolation mapping `i(x)` which interpolates one quantity. Now f(i(a),i(b))!=i(f(a,b)), while it should be equal if we would have detailed calculations instead of interpolation. Because of the inequality, we need to think about, which quantities should be interpolated (usually, the ones we need the best accuracy on).
Another example of this is the interpolation of X, Y and Z, because 1=X+Y+Z should always hold.

posydon/interpolation/profile_interpolation.py
```diff
@@ -0,0 +1,553 @@
...
+            profiles (array-like) : N profiles to be post-processed for monotonicity
+        Returns:
+            profiles_mono (array-like) : post-processed profiles
+        """
+        
+        def mono_renorm(arr):
+            arr_copy = arr.copy()
+            # force rising points down
+            for i in range(1,len(arr_copy)):
+                if arr_copy[i]-arr_copy[i-1]>0:
```
mkruckow: Minor: Why don't you write it as `if arr_copy[i]>arr_copy[i-1]:`? There shouldn't be a need to request an additional difference to be calculated.

posydon/interpolation/profile_interpolation.py
```diff
@@ -0,0 +1,553 @@
...
+                    arr_copy[i]=arr_copy[i-1]
+            # cut off points below surface value
+            low = np.where(arr_copy<arr[-1])[0]
+            arr_copy[low] = arr[-1]
+            return arr_copy
```
mkruckow: Minor: Couldn't lines 237-239 be better written as `return np.where(arr_copy<arr[-1], arr[-1], arr_copy)`?

mkruckow:
In general, whenever you use an imported function, could you please write a comment there, what this function is supposed to do (and hopefully is doing in all cases).

posydon/interpolation/profile_interpolation.py
```diff
@@ -0,0 +1,553 @@
...
+                             'H-rich_Central_He_depleted',
+                             'H-rich_non_burning',
+                             'H-rich_Core_H_burning',
+                             'H-rich_Core_He_burning',
+                             'stripped_He_Central_He_depleted',
+                             'stripped_He_non_burning',
+                             'NS',
+                             'stripped_He_Central_C_depletion',
+                             'H-rich_Central_C_depletion',
+                             'BH']
```
mkruckow: How do you selected those states? E.g. why is there no "WD"?

posydon/interpolation/profile_interpolation.py
```diff
@@ -0,0 +1,553 @@
...
+        Returns:
+            b_models (array-like) : dictionary containing boundary models.
+        """
+        b_models = {}
+        for state in ["H-rich_Shell_H_burning",
+                  'H-rich_Core_H_burning',
+                  'H-rich_Core_He_burning', 
+                  "H-rich_Central_He_depleted",
+                  'H-rich_Central_C_depletion',
+                  "NS","BH"]:
```
mkruckow: Minor: You shouldn't mix different string types (single vs. double quote).

ssbvr:
@elizabethteng I would like to check the quality of a sample of interpolated profiles in the black hole spin calculation context. Could you please find a couple of profiles in the grid you are interpolating where `grid.final_values["Patton&Sukhbold20-engineN20_state"] == 'BH'` and `grid.final_values["Patton&Sukhbold20-engineN20_state_spin"]` is large, say > 0.5 (if you are interpolating CO-HeMS grids) or 0.25 (if you are interpolating HMS-HMS grids)? I will later test the spin prediction accuracy from the `posydon/binary_evol/SN/profile_collapse.py` calculation on the ORIGINAL, LITE, and interpolated profiles.

posydon/interpolation/profile_interpolation.py
elizabethteng: I just used the ones that showed up in the grid I used to develop this; do you know if others are missing besides "WD"?

posydon/interpolation/profile_interpolation.py
elizabethteng: I didn't know you could do that! will definitely be very helpful

posydon/interpolation/profile_interpolation.py
mkruckow: from plot_defaults.py:
`'H-rich_Core_C_burning'`
`'stripped_He_Core_He_burning'`
`'stripped_He_Core_C_burning'`
There are some more strange states, which should never happen:
`'stripped_He_Core_H_burning'`
`'stripped_He_Shell_H_burning'`
`'undetermined_evolutionary_state'`
By the way, I'd suggest you to bring them in a more logical order.

mkruckow (dismissed):
Now, the code looks fine to me.
I didn't run it, hence I suggest some more test runs if not done already.

docs/interpolation/Profile-Interpolator/ProfileInterpolator.rst
```diff
@@ -0,0 +1,106 @@
...
+===============================
+
+To load a pretrained interpolator we need to pass the optional
+``load_interpolator`` argument to the ``ProfileInterpolator`` 
+instance which specifies the path to a .pkl file where the 
+pretrained interpolator can be loaded from. 
+
+.. code-block:: python
+
+  model = ProfileInterpolator(load_interpolator = "/path/to/profile/interpolator.pkl")
```
philipp-rajah, line 51: we should be consistent with how we load pertained interpolators across all of our interpolators. In the IFInterpolator we have a method called load that we call after constructing the class and the argument that points to where the interpolator is saved is called filename.

posydon/interpolation/profile_interpolation.py
```diff
@@ -0,0 +1,572 @@
...
+# POSYDON
+from posydon.grids.psygrid import PSyGrid
+from posydon.interpolation.IF_interpolation import IFInterpolator
+
+# Math and ML
+import numpy as np
+import pandas as pd
+import tensorflow as tf
+tf.get_logger().setLevel('ERROR')
+from tensorflow.keras import layers, losses, models, optimizers
```
philipp-rajah, line 20: In the future we need to decide on either TensorFlow or PyTorch

posydon/interpolation/profile_interpolation.py
```diff
@@ -0,0 +1,572 @@
...
+        Args:
+            grid (obj) : PSyGrid object.
+            ind (int) : index of run to be scraped.
+        Returns:
+            scalars (array-like) : dictionary containing initial 
+                                   m1, m2, p and final s1_state, final_m1.
+            profiles (array-like) : all N specified profiles, shape (N,200).
+        """
+        # open individual run as a DataFrame
+        df = pd.DataFrame(grid[ind]['final_profile1'])
```
philipp-rajah, line 91: So I like the idea of using pandas df, but in IF mostly numpy structured arrays are used. I don't think this matters, but it may be worth discussing in the future, and deciding on one.

posydon/interpolation/profile_interpolation.py
```diff
@@ -0,0 +1,572 @@
...
+        DONT_SAVE = []
+        myattrs = {key: SAVE_ATTRS[key]
+                   for key in SAVE_ATTRS if key not in DONT_SAVE}
+        with open(filename, 'wb') as f:
+            pickle.dump(myattrs, f)
+            
+            
+class ProfileInterpolator:
+    
+    def __init__(self,load_interpolator=None):
```
philipp-rajah: name this argument filename to keep it consistent with the IFInterpolator and have a separate load method

posydon/interpolation/profile_interpolation.py
```diff
@@ -0,0 +1,572 @@
...
+            loss (object) : loss function for training.
+        """
+        print("training on PCA weights...")
+        
+        self.model_prof.compile(optimizers.Adam(clipnorm=1),loss=loss)
+        callback = tf.keras.callbacks.EarlyStopping(monitor="loss",patience=200)
+        history = self.model_prof.fit(self.initial,self.weights,
+                                      epochs=3000,callbacks=[callback],verbose=0,
+                                      validation_data=(self.valid_initial,
+                                                       self.valid_weights))
```
philipp-rajah: maybe name this PCA weights, pretty minor, but I was initially confused about the possibility of passing network weights since I don't work with tf. Also it would be useful to be able to see the model train and to see the loss history. Might make astro people less reluctant to use it since it's not just this black box. Look into programs like tensorboard.

posydon/interpolation/profile_interpolation.py
```diff
@@ -0,0 +1,572 @@
...
+        """Trains NN models. 
+        Args: 
+            loss (object) : loss function for training.
+        """
+        print("training on PCA weights...")
+        
+        self.model_prof.compile(optimizers.Adam(clipnorm=1),loss=loss)
+        callback = tf.keras.callbacks.EarlyStopping(monitor="loss",patience=200)
+        history = self.model_prof.fit(self.initial,self.weights,
+                                      epochs=3000,callbacks=[callback],verbose=0,
```
philipp-rajah: I would have more of these parameters be accessible for tweaking by the user. Depending on the data, it may take more than 3000 epochs to converge, and a different user may want to be more patient. I think that since this whole DL thing is new to most astro people, there will definitely be those curious in what's going on under the hood and some may want to play around with things, which is something I think we should encourage. This goes for all calls of the fit function.

posydon/interpolation/profile_interpolation.py
```diff
@@ -0,0 +1,572 @@
...
+            # identify input/output testing data for class
+            valid_indices = self.valid_sort_ind[state]
+            valid_inputs = self.valid_initial[valid_indices]
+            valid_prof = self.valid_profiles[valid_indices]
+            
+            # calculate boundary points for training and testing data 
+            if "burning" in state:  # these classes' profile shapes have 2 boundary points
+                outs = 2
+                bounds = []
+                nonflat = []  # TODO: explain this
```
philipp-rajah: what is nonflat?

docs/interpolation/Profile-Interpolator/ProfileInterpolator.rst
```diff
@@ -0,0 +1,106 @@
...
+where the interpolator will be saved to.
+
+.. code-block:: python
+
+ model.save("path/for/profile/interpolator.pkl")
+   
+Evaluating on Testing Data
+==========================
+
+To evaluate the interpolator on the testing grid that was used as validation data in
```
philipp-rajah, line 94: Validation set should not be the same as the testing set, this is a bigger issue I'd say. It's probably ok in this case because we won't ever see data far outside of the distribution of the testing and training sets. I'll double check with people here though.

philipp-rajah (commented):
Made a lot of minor comments. I'm assuming that most of the logic of the code works and that this had been tested. Need to check with Aggelos about the validation and testing set.

posydon/interpolation/profile_interpolation.py
elizabethteng: I'll make a note to bring this and tf/torch up in a dev meeting

posydon/interpolation/profile_interpolation.py
elizabethteng: just added this comment: # ensures that training data only has the correct "non-flat" shape
\# to avoid issues with calculating the boundary points

docs/interpolation/Profile-Interpolator/ProfileInterpolator.rst
elizabethteng: yeah we should bring this up when I present on Monday

posydon/interpolation/profile_interpolation.py
elizabethteng: I made the parameters for the PCA weights and H mass fraction boundary points customizable. I don't think anyone could ever need more for the minimum density, since it's such a simple problem. usually only takes a few seconds to train.

posydon/interpolation/profile_interpolation.py
elizabethteng: adding this to end of next PR task list

elizabethteng:
did tests on these changes, everything works as expected

philipp-rajah (commented):
I meant to mention this yesterday but forgot, what about GPU acceleration? I. don't really see it in the code

mkruckow (approved):

elizabethteng:
> I meant to mention this yesterday but forgot, what about GPU acceleration? I. don't really see it in the code

I didn't use GPU acceleration because the training doesn't take that long. like 1 hour per grid maximum. tbh don't think it's worth the effort

philipp-rajah (approved):

ssbvr (approved):

PR #60, opened 2023-04-20
some grid fixes

mkruckow:
- [x] in case there are no history files, try to get metallicity from directory names
- adding the metallicity to be read from directory names (will be used when joining grids as well to match masses, period and metallicity)
- read the metallicity value from directory name when there is no history.data, while X and Y keep nan.
- [x] add initial RLO fix when joining grids, to avoid non-detections on slice boundary or in reruns
- fix updating of the initial RLO list in termination_flags.py, too
- [x] add to reset the display, when creating plots via slurm (ssh -X, may cause otherwise troubles with the display settings)
- [x] add step number to slurm job-name

### DIFF
--- a/bin/setup-pipeline
+++ b/bin/setup-pipeline
@@ -175,6 +175,10 @@ def slurm_job(job_name,
     path_to_csv_file = os.path.join(PATH,f"{job_name}.csv")
 
     with open(f'{job_name}.slurm', 'w') as f:
+        if 'step' in job_name:
+            STEPID = job_name.split("_")[1]
+        elif job_name == 'rerun':
+            STEPID = 'R'
         f.write("#!/bin/bash\n")
         f.write(f"#SBATCH --account={ACCOUNT}\n")
         f.write(f"#SBATCH --partition={PARTITION}\n")
@@ -182,7 +186,7 @@ def slurm_job(job_name,
         f.write("#SBATCH --cpus-per-task 1\n")
         f.write("#SBATCH --ntasks-per-node 1\n")
         f.write(f"#SBATCH --time={WALLTIME}\n")
-        f.write("#SBATCH --job-name=psygrid\n")
+        f.write(f"#SBATCH --job-name=psygrid{STEPID}\n")
         f.write("#SBATCH --mem-per-cpu=4G\n")
 
         if EMAIL is not None:
--- a/bin/setup-pipeline
+++ b/bin/setup-pipeline
@@ -175,10 +175,12 @@ def slurm_job(job_name,
     path_to_csv_file = os.path.join(PATH,f"{job_name}.csv")
 
     with open(f'{job_name}.slurm', 'w') as f:
-        if 'step' in job_name:
+        if ('step' in job_name) and (len(job_name.split("_")) > 1):
             STEPID = job_name.split("_")[1]
         elif job_name == 'rerun':
             STEPID = 'R'
+        else:
+            STEPID = ''
         f.write("#!/bin/bash\n")
         f.write(f"#SBATCH --account={ACCOUNT}\n")
         f.write(f"#SBATCH --partition={PARTITION}\n")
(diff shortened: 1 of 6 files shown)


bin/setup-pipeline
```diff
@@ -175,14 +175,18 @@ def slurm_job(job_name,
     path_to_csv_file = os.path.join(PATH,f"{job_name}.csv")
 
     with open(f'{job_name}.slurm', 'w') as f:
+        if 'step' in job_name:
+            STEPID = job_name.split("_")[1]
+        elif job_name == 'rerun':
+            STEPID = 'R'
```
kkovlakas: If we are sure about the format, sure. I would also add the case of using the raw `job_name` if the `split` does not return 2 items. In the sense:
if step in job_name and len(job_name.split("_")) > 1:
     ...
elif jon_name == "rerun":
    ....
else:
    STEPID = job_name

This avoids two possible runtime errors (also the missing `else`). Just in case.

posydon/grids/psygrid.py
```diff
@@ -2049,6 +2053,27 @@ def say(something):
...
 
+    if (newconfig["initial_RLO_fix"]):
+        say("Determine initial RLO boundary from all grids")
+        detected_initial_RLO = []
+        colnames = ["termination_flag_1", "termination_flag_2", "interpolation_class"]
+        valtoset = ["forced_initial_RLO", "forced_initial_RLO", "initial_MT"]
+        for grid in grids:
+            new_detected_initial_RLO = get_detected_initial_RLO(grid)
+            for new_sys in new_detected_initial_RLO:
+                e = False
```
kkovlakas: Just for the sake of clarity, can you use a more descriptive variable name instead of `e`?

posydon/grids/psygrid.py
```diff
@@ -2049,6 +2053,27 @@ def say(something):
...
+                for i, sys in enumerate(detected_initial_RLO):
+                    # check whether there are double entries
+                    if (abs(sys["star_1_mass"]-new_sys["star_1_mass"])<1.0e-5 and
+                        abs(sys["star_2_mass"]-new_sys["star_2_mass"])<1.0e-5):
```
kkovlakas, line 2068: I would use a general EPS constant or something for such cases, showing that we expect equality rather than accuracy. Does it make sense?

posydon/grids/termination_flags.py
```diff
@@ -260,20 +260,20 @@ def get_detected_initial_RLO(grid):
             period = grid.initial_values[i]["period_days"]
             e = False
             #check for already existing entries of same mass combination
-            for d in detected:
+            for j, d in enumerate(detected):
                 if (abs(d["star_1_mass"]-mass1)<1.0e-5 and
                     abs(d["star_2_mass"]-mass2)<1.0e-5):
                     e = True
```
kkovlakas: Same as before? Variable name (if it's easy).

posydon/grids/psygrid.py
mkruckow: I renamed it.
We should either use long descriptive variable names or using the 80 characters per line limit. Doing both at the same time usually causes too many unnecessary line breaks. When I look at the code, it more looks like we want to go for the 80 character limit, hence I used a short name for an auxiliary variable.

posydon/grids/psygrid.py
mkruckow: I'm not sure, whether this a good idea here. Be aware, that the values we take here as equal are subject to rounding error from a different code, here form MESA. Thus, the EPS constant might be too restrictive.

kkovlakas: OK, I am fine by that. It's a minor comment!

kkovlakas (approved):

ssbvr (approved):
LGTM!

astroJeff (approved):

PR #64, opened 2023-05-03
Add slurm dependency

mkruckow:
- [x] create run-pipeline.sh to run all jobs with dependency
- [x] automatically kill dependent jobs if a previous failed
- [x] add logic, which steps can run in parallel


ssbvr (approved):
LGTM!

astroJeff (approved):

PR #70, opened 2023-05-10
Add support for post processing an arbitrary number of core-collapse models + introduce H envelope option to step_SN + v2 data pop synth debugged

ssbvr:
This PR introduces changes to the step SN and post-processing script.

- [x] step_SN now allows accounting for H envelope in the formation of the compact object
- [x] post-processing allows for an arbitrarily large list of the core-collapse model assumption
- [x] compute and store for interpolation m_disk_accreted and m_disk_radiated (needed for LGRB energy estimations)
- [x] test code on a subsample of single HMS and HMS-HMS grids
- [x] large-scale grid test and training of the interpolators
- [x] support the new interpolated quantities in step_SN for pop synth
- [x] run a large pop synth model to test the processed v2 data

By default, we support 10 core-collapse MODELs, 5 mechanisms with and without the H envelope.

TL;DR:
WD --> m_co_core (no change in assumptions)
NS --> prescription (based on m_co_core, f_fb accounts for m_h_core or m_he_core)
BH (below PPISN) --> prescription (based on m_co_core, m_BH and f_fb accounts for m_h_core or m_he_core)
BH (PPISN) --> prescription (based on m_he_core, never retain any H envelope !!!)
BH (PISN) --> disruption (prescription based on m_he_core)
BH (above PISN) —> direct collapse (prescription based on m_he_core, m_BH accounts for m_h_core or m_he_core)

Example of two primary HMS-HMS forming a BH above the PISN and in the PPISN.
Screenshot 2023-06-07 at 13 12 02

Here is an example of the BH mass as a function of ZAMS for the single star grids at Zsun and 10^-4Zsun for the 10 model variations we support (direct collapse, Fryer rapid, Fryer delayed, Sukhbold, Patton&Sukhbold) x 2 (without and with H-rich envelope).
1e+00_Zsun__no_H-rich_envelope
1e+00_Zsun__with_H-rich_envelope
1e-04_Zsun__no_H-rich_envelope
1e-04_Zsun__with_H-rich_envelope

Here is an example of the SN_type, (CO_type) compact object type, mass, spin as predicted on a mass ratio slice of the HMS-HMS 10^-4 Zsun binary grid. 

Without H-rich envelope:
S1_MODEL05_q_0 7_SN_type_legend
S1_MODEL05_q_0 7_CO_type_legend
S1_MODEL05_q_0 7_mass
S1_MODEL05_q_0 7_spin

Without H-rich envelope:
S1_MODEL10_q_0 7_SN_type_legend
S1_MODEL10_q_0 7_CO_type_legend
S1_MODEL10_q_0 7_mass
S1_MODEL10_q_0 7_spin


ezapartas:
Should we introduce, as we discussed yesterday, the possibility for a user to allow (part of) the H-rich envelope to collpase inside the BH? Either with a boolean flag, or with a [0,1] parameter of the relative mass of the H-rich envelope that is going in. @ssbvr

ssbvr:
> Should we introduce, as we discussed yesterday, the possibility for a user to allow (part of) the H-rich envelope to collpase inside the BH? Either with a boolean flag, or with a [0,1] parameter of the relative mass of the H-rich envelope that is going in. @ssbvr

I forgot about this feature. Let's discuss it at tomorrow's dev meeting what this feature would look like and if we want to support it (note that this feature will not be accessible in population synthesis given the fact that these quantities are preprocessed and interpolated).

mkruckow (changes requested):
I'd suggest to not put all to He core mass hard coded.
We should introduce a tunable parameter to decide how much of the H envelope should be considered. Let's say `f_SN_H` and use `m_rembar=f_SN_H*m_star+(1-f_SN_H)*m_He_core`.
Especially above the PISN gap, using the He-core mass is a bad idea, because the won't be a SN expected at all, there it should be a real direct collapse.

mkruckow:
> > Should we introduce, as we discussed yesterday, the possibility for a user to allow (part of) the H-rich envelope to collpase inside the BH? Either with a boolean flag, or with a [0,1] parameter of the relative mass of the H-rich envelope that is going in. @ssbvr
> 
> I forgot about this feature. Let's discuss it at tomorrow's dev meeting what this feature would look like and if we want to support it (note that this feature will not be accessible in population synthesis given the fact that these quantities are preprocessed and interpolated).

By the way, what is the reason, why we calculate those values in the post processing and not in the pop-syn?

mkruckow:
Another issue we should may have in mind: shouldn't we consider the conversion from baryonic to gravitational mass, too?

EDIT: Yes, I found it. I was just not expecting that this is done in a different function.

ezapartas:
> Another issue we should may have in mind: shouldn't we consider the conversion from baryonic to gravitational mass, too?

I thought we did, following Lattimer & Yahil, 1989 and through the **max_neutrino_mass_loss** parameter

ssbvr:
> > > Should we introduce, as we discussed yesterday, the possibility for a user to allow (part of) the H-rich envelope to collpase inside the BH? Either with a boolean flag, or with a [0,1] parameter of the relative mass of the H-rich envelope that is going in. @ssbvr
> > 
> > 
> > I forgot about this feature. Let's discuss it at tomorrow's dev meeting what this feature would look like and if we want to support it (note that this feature will not be accessible in population synthesis given the fact that these quantities are preprocessed and interpolated).
> 
> By the way, what is the reason, why we calculate those values in the post processing and not in the pop-syn?

Because we do the core collapse of the ORIGINAL stellar profiles to compute the BH spin in the disk accretion calculation. In v1 experiments, we found that the same calculation on downsampled profiles is less accurate. This is because the code assumes the profile is finally divided into shells, and the code computes disk accretion shell by shell.

ssbvr:
> > Another issue we should may have in mind: shouldn't we consider the conversion from baryonic to gravitational mass, too?
> 
> I thought we did, following Lattimer & Yahil, 1989 and through the **max_neutrino_mass_loss** parameter

Yes, we have a revisited version of that formula which accounts for continuity between NS and BH. We also assume *up to* 0.5 Msun mass loss due to neutrinos, see v1 paper.

mkruckow:
> > By the way, what is the reason, why we calculate those values in the post processing and not in the pop-syn?
> 
> Because we do the core collapse of the ORIGINAL stellar profiles to compute the BH spin in the disk accretion calculation. In v1 experiments, we found that the same calculation on downsampled profiles is less accurate. This is because the code assumes the profile is finally divided into shells, and the code computes disk accretion shell by shell.

Hm, would it work, if we would provide e.g. 3 values of all: no H, 50% H, all H? In this way, we could make some simple interpolation for the pop-syn, if the considered H fraction is in between the three? So, that we get not the exact value, but a free scalable one which should be close.

posydon/binary_evol/SN/step_SN.py
```diff
@@ -70,6 +70,7 @@
     "engine": 'N20',
     "PISN": "Marchant+19",
     "ECSN": "Podsiadlowksi+04",
+    "conserve_hydrogen_envelope" : False,
```
mkruckow, line 75: Just a small naming thing: I wouldn't call it `conserve_hydrogen_envelope`, because we don't conserve the hydrogen-rich envelope. The NS/BH won't have a hydrogen-rich envelope. We include/consider it in the collapse. Thus, I'd suggest names like `include_hydrogen_envelope` or `collapse_hydrogen_envelope`.

posydon/binary_evol/MESA/step_mesa.py
```diff
@@ -788,31 +789,22 @@ def update_properties_NN(self, star_1_CO=False, star_2_CO=False,
...
+                        for key in ['state', 'SN_type', 'f_fb', 'mass', 'spin',
+                                    'm_disk_accreted', 'm_disk_radiated']:
+                            if key == 'state': 
+                                key = 'CO_type'
+                            values[key] = cb.final_values[f'S{i+1}_{MODEL_NAME}_{key}']
+                        state = cb.final_values[f'S{i+1}_{MODEL_NAME}_CO_type'] 
                         if state is None or state == 'None':
                             # privent to any quantities for star that did not
                             # reach core collpase
                             setattr(star, key, None)
```
mkruckow: Here `key` is not well defined, because it is outside the for loop

posydon/grids/post_processing.py
```diff
@@ -25,28 +27,45 @@
...
-    "mechanism": None,
-    "engine": None,
-    "PISN": "Marchant+19",
-    "ECSN": "Podsiadlowksi+04",
-    "max_neutrino_mass_loss": 0.5,
-    "kick": True,
-    "sigma_kick_CCSN_NS": 265.0,
-    "sigma_kick_CCSN_BH": 265.0,
-    "sigma_kick_ECSN": 20.0,
-    "max_NS_mass": 2.5,
```
mkruckow, line 38: A question: I haven't seen the kick values in the new version anymore. Is it correct, that we can leave them out or are they missing now?

posydon/binary_evol/MESA/step_mesa.py
ssbvr: Good catch, I fixed the bug. This portion of the code was not tested as I did not run a population synthesis model with the first nearest neighbour.

posydon/grids/post_processing.py
ssbvr: The post-processing script uses the `collapse_star` method of the `step_SN`. Kicks arguments are used only by the `orbital_kick` method which is not part of the `collapse_star` method. When running the `__call__` method of the StepSN object, e.g. during population synthesis, then they will be called one after the other.

mkruckow (dismissed):

mkruckow (dismissed):

ezapartas:
@simone @mkruckow  
Suggested small addition to the PR, to fix cases of mgrav being <~2.5 Msun (max_NS_mass) after the neutrinos losses calculations
At line 685 of the new version:
we should add
`m_grav = max(m_grav,self.max_NS_mass)`

mkruckow:
> @simone @mkruckow Suggested small addition to the PR, to fix cases of mgrav being <~2.5 Msun (max_NS_mass) after the neutrinos losses calculations At line 685 of the new version: we should add `m_grav = max(m_grav,self.max_NS_mass)`

I think, that's not the correct way. There is already a check for m_grav>self.max_NS_mass in line 693 and 723. In your case you got to line 750, because you neither have a profile nor use the core masses.
We should check `compute_m_rembar`, why it gives a state NS for a star getting m_grav>self.max_NS_mass.

mkruckow (dismissed):

ezapartas (dismissed):

mkruckow (approved):

astroJeff (approved):

PR #71, opened 2023-05-10
add subclass of reverseMT and DoubleCE

ZepeiX:


mkruckow:
I'd suggest, to move your logic from `infer_interpolation_class` to `get_mass_transfer_flag`, thus modifying the TF2, which is aimed to store the mass transfer information. And getting the new interpolation class based on the new values of TF2.
(Be aware, when you get a new TF2, you need to add this in other files of the visualization class.)

posydon/grids/termination_flags.py
```diff
@@ -231,8 +232,52 @@ def infer_interpolation_class(tf1, tf2):
...
+            rl_overflow1 = binary_history['rl_relative_overflow_1']
+            rl_overflow2 = binary_history['rl_relative_overflow_2']
+            center_h1 = history1['center_h1']
+            forward = np.logical_and(rl_overflow1>RL_RELATIVE_OVERFLOW_THRESHOLD,
+                                     rl_overflow2<RL_RELATIVE_OVERFLOW_THRESHOLD)
+            reverse = np.logical_and(rl_overflow1<RL_RELATIVE_OVERFLOW_THRESHOLD,
+                                     rl_overflow2>RL_RELATIVE_OVERFLOW_THRESHOLD)
+            reverse_postMS = np.logical_and(reverse,
+                                           center_h1<THRESHOLD_CENTRAL_ABUNDANCE)
+            if (np.max(forward) == True and np.max(reverse_postMS) == True):
```
mkruckow, line 245: Why do you request to be post MS?

posydon/grids/termination_flags.py
```diff
@@ -231,8 +232,52 @@ def infer_interpolation_class(tf1, tf2):
...
+                return "stable_reverse_MT"
+            else:
+                return "stable_MT"
         return "stable_MT"
     if tf1 in TF1_POOL_UNSTABLE:
+        if (binary_history is not None and history1 is not None):
```
mkruckow, line 251: all below: could you please use different variable names, to indicate that they are only from the last entry in the history and not a list/numpy-array like in the stable RLO part.

mkruckow:
I think we should discuss in a meeting, under which conditions we like to get into a double CE.
In the POSYDON paper, there it is call "For the case in which the companion star also has a giant-like structure with a distinct core-envelope separation (i.e., anything but an MS star, an He star in its He-MS, or a CO), then its envelope also **may** contribute in a (double) CE." I'd say here as well, in a contact case with unstable RLO -> double CE, but for forward/reverse unstable MT with a non-MS nor CO companion, it is unclear whether we should go into a double CE. Yes, some of the companion's envelope might be lost as well, but all?
I'm not sure whether it may only remove a minority, this material will get started to be lost from the companion only because of the interaction with the first CE, hence much later and it will probably take a thermal timescale of the companion's envelope to transfer it to become part of the CE, which is potentially much longer than the CE timescale.

posydon/grids/termination_flags.py
ZepeiX: I want to exclude the ones with very close orbits where they have MT from both stars when they are MS

mkruckow: In which way are they different, that they need to be excluded? Do we have many of forward and backward MT on MS, which is not going into a contact system?

ZepeiX: The reason we separate reverse_MT is that the MT would cause spin up of the star_1 and star_2 would be a partially stripped postMS star, which are very different from forward_MT cases.  If the reverse_MT happaned when the star1 is still in MS, the binary is more like the forward_MT case.

mkruckow: The spin is only one reason, the masses are different as well, because with reverse MT star1 gained mass, while without reverse MT star 1 never gained mass. This holds for the MS case, too.
I don't really get, why does the reverse MT on MS don't lead to critical rotation like you are suggesting?

ZepeiX: The mass gained through reverse MT should be little because of rotation. And MS stars will lose AM after because of forward MT and winds so it will end with a low spin anyway. While for the reverse MT cases I am talking about here, the primary star is spun up close to SN and it will end with a moderate spin.

mkruckow: OK, for initial-final interpolation, you are correct.
But we need to be aware, when we would like to get interpolation during the evolution in the future, this will cause trouble there, without anybody remembering this to be the reason.

PR #88, opened 2023-06-05
Hdf5 refactor - data types for population output

ka-rocha:
Working with @celiotine:

Data types have been flexible in the past but are causing issues with IO for binary population output. To fix this, we propose explicit declaration of data types of all output variables in posydon. We expect users to specify data types for any _extra_ columns they are adding to posydon that are not part of the default dictionaries in source. 

Some motivation:

- We expect explicit data type handing to help with the combining of batch files at the end of population runs. This was the main issue raised by many people running standard populations and when adding extra features to posydon.
- Users adding hooks and extra columns (calculating physics things) must add many conditions to clean binary history removing random `None`s in order to use `numpy` operations without erroring.

Some things to add before this is complete:

- [x] Functionality for users to pass dictionaries (instead of lists) with `extra_columns` names and dtypes as arguments for `BinaryPopulation` (with actual implementation in `SimulationProperties`, `EvolveHooks`, `BinaryStar`, and `SingleStar`)
- [x] Add these options and default dictionaries for the `oneline` `DataFrames`. 

As a consequence of this change, we would recommend developers respect data types of output variables (e.g. `S1_mass`) throughout evolutionary steps by not mixing `None` in with variables that are naturally `float` typed for example (use `np.nan` instead). For columns containing strings (e.g. `event`) we should always use strings instead of `None` objects mixed in. This would go in line with changes @celiotine & @astroJeff proposed to add more descriptive `event` names for example.

Also note: we tried implementing `pandas` builtin `string` data type for saving to HDF [but this apparently is not implemented](https://github.com/pandas-dev/pandas/issues/31199). Our work-around is to cast things as the builtin `string` data type to coerce non valid data to be represented as strings and then go back to `object` for saving. According to the docs, the performance between the two types for strings is similar.

Syntax changes:
The parameter `extra_columns` in `BinaryStar`'s `to_df` and `to_oneline_df` methods take dictionaries instead of lists with column names. The dictionaries now specify the name of the column and the data type. 
```python
binarypop_kwargs = {
    ...
    extra_columns = {'step_names':'string', 'step_times':'float64'}
}
```


ssbvr (approved):
I tested the PR in comination with PR#122 in a multi metallicity pop sinth of 2k binaries and everything works fine. I added the change to the default ini file required by this PR directly in PR122.

kasdaglie (approved):

PR #89, opened 2023-06-06
Allow grid end at He depletion (for massive stars)

mkruckow:
- [x] new option for pysgrid creation `stop_before_carbon_depletion` similar to `start_at_RLO`
- additionally it requires an initial mass to be at least 100Msun
- [x] new function `keep_till_central_abundance_He_C` similar to `keep_after_RLO`
- It takes upper limits of the He and C abundance to stop as arguments. Function defaults 10^-5 and 1, respectively. Called with our common `THRESHOLD_CENTRAL_ABUNDANCE` used for helium and 0.1 for carbon if psygrid is created with `stop_before_carbon_depletion=True`.
- [x] add of `get_flags_from_MESA_run` a way to set a TF1 (needed to overwrite the TF taken from the MESA output files in case we crop some late evolution)

### DIFF
--- a/posydon/grids/scrubbing.py
+++ b/posydon/grids/scrubbing.py
@@ -208,14 +208,22 @@ def keep_till_He_depletion(bh, h1, h2, Ystop=1.0e-5, XCstop=1.0):
         return bh, h1, h2, ''
     
     if depleted1:
-        where_conditions_met1 = np.where((h1["center_he4"]<Ystop) and (h1["center_c12"]<XCstop))[0]
+#        where_conditions_met1 = np.where((h1["center_he4"]<Ystop) and (h1["center_c12"]<XCstop))[0]
+        where_conditions_met1 = []
+        for i in range(len(h1["center_he4"])):
+            if (h1["center_he4"][i]<Ystop) and (h1["center_c12"][i]<XCstop):
+                where_conditions_met1 += [i]
         if len(where_conditions_met1) == 0:
             warnings.warn("No He depletion found in h1, while expected.")
             return bh, h1, h2, ''
         last_index = where_conditions_met1[0]
         newTF1 = 'Primary has depleted central helium'
     if depleted2:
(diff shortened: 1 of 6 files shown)


posydon/grids/scrubbing.py
```diff
@@ -152,3 +153,101 @@ def keep_after_RLO(bh, h1, h2):
         new_h2["star_age"] -= age_to_remove
 
     return new_bh, new_h1, new_h2
+
+
+def keep_till_He_depletion(bh, h1, h2, Ystop=1.0e-5, XCstop=1.0):
```
ezapartas: My main comment is that I am not sure that the  "He depletion" term in the 'stop_at_He_depletion' and 'keep_till_He_depletion' is the most appropriate term, as in the case THRESHOLD_CENTRAL_ABUNDANCE  = Xc<~0.3-0.4, we probably go BEYONG He depletion as we have almost depleted all Carbon in the center too. We used to use another term, sth like "Close_to_carbon_depletion", do you think it makes sense here?

ezapartas (changes requested):

posydon/grids/scrubbing.py
mkruckow: Yes, the name was taken when we were thinking, that we should get all the way back to He-deletion. With later changes it turns out, that the name might not be that well anymore.

Hence, I'd more go to change it to what it is: the function keeps stuff till both requirements are meet, hence I suggest `keep_till_central_abundance_He_C`, the function is aimed to be more general, that one can use it for any combination of He and C abundance specified.
For the flag, we can take something like `stop_before_carbon_depletion`, because otherwise it would stop at carbon depletion, hence the name should represent the difference to that.
Would you agree with that?

ezapartas (dismissed):

ssbvr (dismissed):

ezapartas (approved):

ssbvr (approved):

PR #92, opened 2023-06-08
Initially single stars and merged isolated evolution

kasdaglie:
This PR creates and evolves single stars in POSYDON and also effectively evolves stars that were produced by mergers.


posydon/utils/common_functions.py
```diff
@@ -44,7 +44,7 @@
                   "Shell_H_burning", "Central_He_depleted",
                   "Central_C_depletion"]
 RICHNESS_STATES = ["H-rich", "stripped_He"]
-COMPACT_OBJECTS = ["WD", "NS", "BH"]
+COMPACT_OBJECTS = ["WD", "NS", "BH","massless_remnant"]
```
ezapartas, line 47: Let's be consistent throughout the code to have COMPACT_OBJECTS = ["WD", "NS", "BH"] and NOT_NORMAL_STARS = ["WD", "NS", "BH","massless_remnant"]

posydon/popsyn/binarypopulation.py
```diff
@@ -915,11 +914,11 @@ def draw_initial_binary(self, **kwargs):
                 center_he4=Y,
             )
             star2_params = dict(
-                mass=m2,
-                state="BH",
+                mass=0,
```
ezapartas: reminder that we should connect this with the function of massless_remnant

posydon/utils/common_functions.py
astroJeff: massless_remnant seems very supernova oriented language. Can we use a different term? Maybe "massless" or "not_a_star" or something else?

posydon/popsyn/binarypopulation.py
```diff
@@ -883,7 +882,7 @@ def draw_initial_binary(self, **kwargs):
         #If binary_fraction not default a initially single star binary is created.
         else:
             formation_time = output['time'].item()
-            separation = output['separation'].item()
+            separation = 1e99
```
astroJeff: Instead of setting the separation to 1e99, can we set it to np.nan? Or maybe a negative number? Someone looking for a bound system might select systems using the condition separation>0 and be confused by the results.

ezapartas:
We have already implemented new isolated step which includes the isolated evolution of  initially single stars, ejected companions in disrupted systems from a previous SN and also Merger products.


Isolated step is linked to detached step, where the isolated star is matched to a single star grid. In principle we can assign physically motivated single star grids, better matched for each case (e.g. potentially single star grids spanning a big combination of envelope to core mass ratio for merger products, for disrupted we can link to a grid of binary mass accretor grids \citep[e.g.,][]{Renzo+2022}).

Assume an extremely large orbit, and no calculation of orbital evolution. Only spin evolution of the isolated star due to winds and moment of inertia change  (assuming solid body rotation).

For merged or initially single stars, the companion is considered a new state called massless_remnant (the name may change potentially to sth like massless_star or no_star (?)).

Many new assumptions during the merged process:
\item similar to \citet{Hurley+2002}. Core with heavier elements sink in the center, added from both stars (e.g. if both stars have a helium core, we add their mass for the value of the he core of the merged object. Central abundances are then decided as the mass weighted average of both stars if they both contribute to the core, else only for the star that contributes to the core. Also we assume mass weighted mixing on the surface abundances of the lighter element envelope, including  whole comp with the envelope of base star. 
\item assumption in mass loss during merging of two MS stars \citep[e.g.][]{Izzard+, de Mink+}
\item ad hoc spin

ezapartas (dismissed):

posydon/popsyn/defaults.py
```diff
@@ -15,10 +15,11 @@
 
     # Size of the population
     'number_of_binaries': 100,
-    #Eirini's changes 
+    #Eirini's changes
```
ssbvr: Avoid this type of comment. The repository is now public. With `git blame` or the git extension for VS code, you can see directly who is the author of each line of code.

ssbvr:
Do not forget to untag the commit change to the MESA-INLIST submodule.

astroJeff (approved):

ezapartas (approved):

PR #103, opened 2023-06-23
Introducing GRB properties and rate calculations

ssbvr:
This is a PR draft showing the new functionalities I am bringing to POSYDON to compute LGRB properties and rates.

- [x] embed a method to `SyntheticPopulation` to compute the GRB properties
- [x] add GRB rate calculations
- [x] add GRB plotting methods
- [x] test code integration with dco rate calculation
- [x] generalise plotting methods
- [x] post a tutorial on slack

Load synthetic population
```py
MODEL = {
    'delta_t' : 100, # Myr
    'SFR' : 'IllustrisTNG',
    'sigma_SFR' : None,
    'Z_max' : 1.,
    'Zsun' : 0.0142,
    'compute_GRB_properties' : True,
    'GRB_beaming' : 'Goldstein+15', # e.g., 0.5, 
    'GRB_efficiency' : 0.01, # e.g., 0.01
    'E_GRB_iso_min' : 1e51, # e.g., 1e51 erg 
}

pop.get_dco_at_formation(S1_state='BH', S2_state='BH', formation_channels=True,
                         oneline_cols=['S1_natal_kick_array_0', 'S2_natal_kick_array_0',
                                       'interp_class_HMS_HMS'])
```

Compute GRB intrinsic and observable (beamed) population and rate densities
```py
pop.get_grb_rate_density() 
```
```txt
GRB (beamed) rate density in the local Universe (z=0.00): 0.14 Gpc^-3 yr^-1
```

Compute BBH rate density and detection rate
```py
pop.get_dco_merger_rate_density()
pop.get_dco_detection_rate()
```
```txt
DCO merger rate density in the local Universe (z=0.00): 93.74 Gpc^-3 yr^-1
DCO detection rate at design_H1L1V1 sensitivity: 5368.19 yr^-1
```

Plot rate densities
```py
kwargs = dict(ylim=(1e-2,4e2), zmax=10., show=True, path=None, channels=True,
              GWTC3=True, Perley16=True, grb_components=False)
pop.plot_rate_density(DCO=True, GRB=True, **kwargs)
```
Screenshot 2023-07-04 at 12 58 40

Plot population property distributions
```py
kwargs = dict(bins=100, xlog=True)
pop.plot_hist_properties(['S1_E_GRB_iso','S2_E_GRB_iso'], intrinsic=True, observable=True, pop='GRB', **kwargs)
pop.plot_hist_properties('S2_f_beaming', intrinsic=True, observable=True, pop='GRB', **kwargs)
kwargs = dict(bins=100, xlog=False)
pop.plot_hist_properties('m_chirp', intrinsic=True, observable=True, pop='DCO', **kwargs)
pop.plot_hist_properties('chi_eff', intrinsic=True, observable=True, pop='DCO', **kwargs)
pop.plot_hist_properties(['S1_spin','S2_spin'], intrinsic=True, observable=True, pop='DCO', **kwargs)
pop.plot_hist_properties('chi_eff', intrinsic=True, observable=True, pop='DCO', channel='ZAMS_oRLO1_CC1_oRLO2_CC2_CO_contact_END', **kwargs)
pop.plot_hist_properties('chi_eff', intrinsic=True, observable=True, pop='DCO', channel='ZAMS_oRLO1_CC1_oRLO2_oCE2_CC2_CO_contact_END', **kwargs)
```
Screenshot 2023-07-04 at 13 20 35
Screenshot 2023-07-04 at 13 20 42

... refer to the tutorial on slack for the other plots.


posydon/popsyn/GRB.py
```diff
@@ -0,0 +1,119 @@
...
+    if E_GRB_iso_min != 0.:
+        warnings.warn("We only consider GRBs with isotropic equivalent "
+                        f"energy larger than {E_GRB_iso_min} erg!")
+        
+    # define some methods
+    
+    def compute_eta(i, sel):
+        """Compute GRB energy conversion efficiency factor."""
+        if (not isinstance(GRB_efficiency, float) and
+            GRB_efficiency >= 0. and GRB_efficiency <= 1.):
```
mkruckow: It looks like a bracket is missing here, it should read `(not (isinstance(GRB_efficiency, float) and GRB_efficiency >= 0. and GRB_efficiency <= 1.))` or `(not isinstance(GRB_efficiency, float) or GRB_efficiency < 0. or GRB_efficiency > 1.)`

posydon/popsyn/GRB.py
```diff
@@ -0,0 +1,119 @@
...
+    
+    # reminder the user about the threshold cut
+    if E_GRB_iso_min != 0.:
+        warnings.warn("We only consider GRBs with isotropic equivalent "
+                        f"energy larger than {E_GRB_iso_min} erg!")
+        
+    # define some methods
+    
+    def compute_eta(i, sel):
+        """Compute GRB energy conversion efficiency factor."""
```
mkruckow: At the moment this function is not computing an efficiency, it only sets it to the value given by `GRB_efficiency`

posydon/popsyn/GRB.py
```diff
@@ -0,0 +1,119 @@
...
+        # 'Fong+15', 'Ghirlanda+16', 'Lamb+19'?
+        
+        df[f'S{i}_f_beaming'] = [np.nan] * df.shape[0]
+        if isinstance(GRB_beaming, float):
+            df.loc[sel, f'S{i}_f_beaming'] = GRB_beaming
+        elif GRB_beaming == 'Goldstein+15':
+            n = df.loc[sel,f'S{i}_f_beaming'].shape[0]
+            # generate a opening angle from a lognormal distribution 
+            # as in Table 3 of https://arxiv.org/pdf/1512.04464.pdf
+            theta = 10**(np.random.normal(0.77,0.37,n))/180.*np.pi # radian
```
mkruckow, line 50: I guess, we should truncate the function to the interval [0,pi], currently it is unlikely but not impossible to get a value outside that range.

mkruckow: As suggested by the next lines, we should may even restrict the range to [0,pi/2], because we automatically assume a symmetry for both hemispheres.

posydon/popsyn/GRB.py
```diff
@@ -0,0 +1,119 @@
...
+            # compute the solid angle, factor 2 to accounts for the two emispheres
+            solid_angle = 4*np.pi*(1.-np.cos(theta)) 
+            df.loc[sel,f'S{i}_f_beaming'] = solid_angle/(4*np.pi)
+        elif GRB_beaming == 'Lloyd-Ronning+19':
+            if z is None:
+                raise ValueError('Redshift not specified!')
+            else:
+                z = df.loc[sel,'z_formation']
```
mkruckow: This looks strange: if `z is None` we raise an error, otherwise we set `z` to a new value (data frame with values), hence if it had a value this gets overwritten.

posydon/popsyn/GRB.py
```diff
@@ -0,0 +1,119 @@
...
+            n = df.loc[sel,f'S{i}_f_beaming'].shape[0]
+            # generate a opening angle from a lognormal distribution
+            # as in Table 3 of https://arxiv.org/pdf/1512.04464.pdf
+            theta =10**(np.random.normal(0.77*(1+z)**(-0.75),0.37,n))/180.*np.pi # radian
```
mkruckow: again, this angle should not be allowed to have any value in (0,inf)

posydon/popsyn/GRB.py
```diff
@@ -0,0 +1,119 @@
...
+            # compute the solid angle, factor 2 to accounts for the two emispheres
+            solid_angle = 4*np.pi*(1.-np.cos(theta))
+            df.loc[sel,f'S{i}_f_beaming'] = solid_angle/(4*np.pi)
+        elif GRB_beaming == 'Pescalli+15':
+            def f_B(E,eta):
+                xi=0.5
+                t=25.
+                C = (7e46)**xi
+                fb = np.ones(len(E))
+                fb[E>0.] = (C*(eta*E[E>0.]/t)**(-xi))**(1./(1-xi))
```
mkruckow, line 79: Please double check, that it is 1/(1-xi), from a quick look at the paper I had the impression it should be C*(1-xi)*L^(-xi)

posydon/popsyn/GRB.py
```diff
@@ -0,0 +1,119 @@
...
+            theta =10**(np.random.normal(0.77*(1+z)**(-0.75),0.37,n))/180.*np.pi # radian
+            # compute the solid angle, factor 2 to accounts for the two emispheres
+            solid_angle = 4*np.pi*(1.-np.cos(theta))
+            df.loc[sel,f'S{i}_f_beaming'] = solid_angle/(4*np.pi)
+        elif GRB_beaming == 'Pescalli+15':
+            def f_B(E,eta):
+                xi=0.5
+                t=25.
+                C = (7e46)**xi
+                fb = np.ones(len(E))
```
mkruckow, line 69: Please double check, that you like to have a beaming fraction of 1, in case of a negative energy. I'd more expect a fraction of 0 to cut off negative energies.

posydon/popsyn/GRB.py
```diff
@@ -0,0 +1,119 @@
...
+        df[f'S{i}_E_GRB_iso'] = np.zeros(df.shape[0])
+        E_GRB = df.loc[sel,f'S{i}_E_GRB']
+        f_beaming = df.loc[sel,f'S{i}_f_beaming']
+        df.loc[sel,f'S{i}_E_GRB_iso'] = E_GRB/f_beaming # erg
+        
+    def compute_L_GRB_iso(i, sel):
+        """Compute the GRB isotropic equivalet luminosity."""
+        df[f'S{i}_L_GRB_iso'] = np.zeros(df.shape[0])
+        # compute the GRB isotropic equivalent luminosity assuming a
+        # constant average timescale of 2.5s
```
mkruckow: Could you give a reference for the 2.5s? Pescalli+15 had a t=25s, which is a factor 10 different.
May implement it as a free parameter for that function, where you set the default, e.g. to the 2.5s.

posydon/popsyn/GRB.py
```diff
@@ -0,0 +1,119 @@
...
+        # compute the GRB isotropic equivalent luminosity assuming a
+        # constant average timescale of 2.5s 
+        # TODO: improve this assumption
+        E_GRB_iso = df.loc[sel,f'S{i}_E_GRB_iso']
+        df.loc[sel,f'S{i}_L_GRB_iso'] = E_GRB_iso/2.5 # erg/s
+    
+    def flag_GRB_systems(i, sel):
+        # flag all systems emitting at least one GRB
+        df[f'GRB{i}'] = [False] * df.shape[0]
+        df.loc[sel,f'GRB{i}'] = df.loc[sel,f'S{i}_E_GRB_iso'] > E_GRB_iso_min
```
mkruckow: Should that be a `>=` to include the allowed minimum?

posydon/popsyn/rate_calculation.py
```diff
@@ -31,11 +32,16 @@
     'Z_max' : 1.,
     'select_one_met' : False,
     'dlogZ' : None, # e.g, [np.log10(0.0142/2),np.log10(0.0142*2)]
-    'Zsun' : Zsun
+    'Zsun' : Zsun,
+    'compute_GRB_properties' : False,
+    'GRB_beaming' : None, # e.g., 0.5, 'Goldstein+15'
+    'GRB_efficiency' : None, # e.g., 0.01
+    'E_GRB_iso_min' : None, # e.g., 1e51 erg
```
mkruckow: Why do you use unsupported `None` for default values? I'd suggest to use
`'GRB_beaming' : 1.0,
'GRB_efficiency' : 0.0,
'E_GRB_iso_min' : 0.0,`

posydon/popsyn/rate_calculation.py
```diff
@@ -657,18 +673,16 @@ def f(z,sensitivity):
...
         ----------
         w_ijk : array doubles
             Cosmological weights computed with Eq. B.8 of Bavera et at. (2020).
         z_event : array doubles
             Cosmolgocial redshift of the event you are tracking.
-        Type : string
+        observable : string
             Event you are tracking, available:
             'DCOs': merger event of a DCO system
             'GRBs': coming soon.
```
mkruckow: Please update the lines above

posydon/popsyn/rate_calculation.py
```diff
@@ -657,18 +673,16 @@ def f(z,sensitivity):
         return sp.integrate.quad(f, z_hor_i, z_hor_f, args=(sensitivity))[0] # Gpc^3
 
 
-    def compute_merger_rate_density(self, w_ijk, z_event, observable='DCOs', sensitivity='infiite', index=None):
-        """Compute the DCOs merger rate density.
-
-        TODO: add support for GRBs.
+    def compute_rate_density(self, w_ijk, z_event, observable='DCO', sensitivity='infiite', index=None):
```
mkruckow: shouldn't it read `'infinite'`?

posydon/popsyn/rate_calculation.py
```diff
@@ -689,20 +703,230 @@ def compute_merger_rate_density(self, w_ijk, z_event, observable='DCOs', sensiti
...
+        if extention == 'npz':
+            if self.verbose:
+                print('Formatting the data ....')
+            for s in [1,2]:
+                data_to_save = [[],[],[],[]]
+                for i, dict in enumerate([data[key] for key in data.keys() if f'{s}' in key]):
+                    for key in dict.keys():
+                        data_to_save[i].extend(dict[key].tolist())
+                        if i == 0:
+                            data_to_save[3].extend(np.ones(len(dict[key]))*float(key))
```
mkruckow: I'd suggest to use `[-1]` instead of `[3]`, this would allow future extensions in a more easy way. (the same applies then to line 894)

posydon/popsyn/synthetic_population.py
```diff
@@ -230,6 +239,10 @@ def parse(self, S1_state=None, S2_state=None, binary_state=None,
                     df_sel_met = pd.concat([df_sel_met, df_tmp])
                     del df_tmp
 
+            if k != 0:
```
mkruckow: Like for PR 104, I'd suggest to use `k>0` instead of `k!=0` when you like have something executed from the second iteration on. (I won't repeat this for each single case following.)

posydon/popsyn/synthetic_population.py
```diff
@@ -343,6 +375,45 @@ def get_dco_at_formation(self, S1_state, S2_state, oneline_cols=None):
...
+                            warn=False)
+                last_index = -1
+                time = []
+                for index, value in self.df.loc[logic,'time'].items():
+                    if index != last_index:
+                        time.append(value)
+                    last_index = index
+                if len(time) != self.df_synthetic.shape[0]:
+                    raise ValueError('Missing values in time_{event}!')
+                self.df_synthetic[f'time_{event}'] = np.array(time)*1e-6 # Myr
```
mkruckow, line 418: I'm a bit confused: In `rate_calculation.py` you convert the column `time` with 10^-3 to Gyr, suggesting that this column in the data frame has the unit of Myr. Here, you convert it via 10^-6 to Myr, suggesting, that this `time` column is in units of yr. What is correct? In the case the data frames use different units, we need to make this clear by adding the unit to the column name, otherwise this will cause trouble in the future.

mkruckow:
The comments I put up for PR #104 apply here, too. Hence, please check them.

posydon/popsyn/GRB.py
mkruckow: I'd put the scaling with z outside the original calculation for theta, because they say in the paper hat theta ~ (1+z)^-0.75; otherwise we would need to think about a z-dependence on sigma.
beside that, they show a straight line in log space, hence it should even be outside the power with base 10.

posydon/popsyn/rate_calculation.py
```diff
@@ -689,20 +703,230 @@ def compute_merger_rate_density(self, w_ijk, z_event, observable='DCOs', sensiti
...
                                      /self.get_shell_comovig_volume(z_hor[i-1],
                                                                     z_hor[i],
                                                                     sensitivity))
-                return Rate_DCOs # Gpc^-3 yr^-1
+                return Rate_DCO # Gpc^-3 yr^-1
             else:
                 raise ValueError('Unsupported sensitivity!')
+        
+        elif observable in ['GRB1','GRB2']:
+            s = observable[-1]
```
mpgalleg, line 724: It is unclear to me from here and the raise ValueError that the user will know observable should be one string

posydon/popsyn/rate_calculation.py
```diff
@@ -689,20 +703,230 @@ def compute_merger_rate_density(self, w_ijk, z_event, observable='DCOs', sensiti
...
+            'CC1', 'CC2'.
+
+        Returns
+        -------
+        t_BRB : double
+            Cosmic time of the GRB event in Gyr.
+
+        """
+        if event not in ['CC1', 'CC2']:
+            raise ValueError(f'Unknown event {event}!')
```
mpgalleg, line 786: in the for loop on line 837 you specify `for s, event in enumerate(['CC1', 'CC2'])` why is this ValueError need here?

posydon/popsyn/rate_calculation.py
```diff
@@ -689,20 +703,230 @@ def compute_merger_rate_density(self, w_ijk, z_event, observable='DCOs', sensiti
...
+                    
+                    # sort out system not emitting GRBs
+                    bool_GRB = np.logical_and(t_CC < cosmology.age(1e-08).value*0.9999999 , self.df[f"GRB{s+1}"])
+                    
+                    # if there are no system emitting any GRB, continue
+                    if len(index[bool_GRB]) == 0:
+                        data[f'index_{s+1}'][str(z_birth[i])] = np.array([])
+                        data[f'z_grb_{s+1}'][str(z_birth[i])] = np.array([])
+                        data[f'weights_{s+1}'][str(z_birth[i])] = np.array([])
+                        continue
```
mpgalleg, line 853: would it be helpful to the user to have a statement here "No system emitting any GRB"?

posydon/popsyn/rate_calculation.py
```diff
@@ -689,20 +703,230 @@ def compute_merger_rate_density(self, w_ijk, z_event, observable='DCOs', sensiti
...
+                    data[f'index_{s+1}'][str(z_birth[i])] = index[bool_GRB]
+                    data[f'z_grb_{s+1}'][str(z_birth[i])] = z_GRB
+
+                    # compute and store cosmological rate weights as in eq. B.8 in Bavera et al. (2020)
+                    z_b = np.ones(len(index[bool_GRB]))*z_birth[i]
+                    w_ijk = self.merger_rate_weight(z_b, z_GRB, p_det, index[bool_GRB])
+                    data[f'weights_{s+1}'][str(z_birth[i])] = w_ijk
+
+                else:
+                    raise ValueError('Unknnown sensitivity!')
```
mpgalleg: change to `Unknown`

posydon/visualization/plot_pop.py
```diff
@@ -0,0 +1,254 @@
...
+            if not xlog:
+                plt.xlabel(x[0])
+            else:
+                plt.xlabel('log10_'+x[0])
+    if isinstance(x, list):
+        plt.legend(loc=1)
+    if path:
+        plt.savefig(path)
+    if show:
+        plt.show()
```
mpgalleg, line 254: Sometimes I like to return the ax object (when there is one) so that I can make edits or overplot something. Something to consider eventually maybe

posydon/popsyn/GRB.py
ssbvr: In Python, if the first condition is `False` the other two `and` conditions will never be executed as the logic directly evaluates to `False`. So in this case the first condition will check for the variable not being a float, if it is a float it will continue to evaluate the other two, one by one.

posydon/popsyn/GRB.py
ssbvr: The code needs to be set up to use this beaming option as `z_formation` is specified by the rate calculation only after the GRB properties are evaluated. I just put the option here for completeness but there is some major work to make usable. So I will not act on this.

posydon/popsyn/GRB.py
ssbvr: agreed adding a TODO

posydon/popsyn/GRB.py
ssbvr: added a TODO in the code

posydon/popsyn/GRB.py
ssbvr: I don't remember where I took it, probably Bromberg, I am updating to 25s for the moment and citing Pescalli+15. Again all of this needs to be tested and improved by someone doing an active science project.

posydon/popsyn/synthetic_population.py
ssbvr: Yes, the rate calculation class uses `Myr` while POSYDON `yr`. The code is converting the units only for the exported datasets where we get DCO at formation (this dataset is no longer following POSYDON pop synth indexing or columns, so I think we are ok but I understand the confusion). From this point, having time in Myr is very convenient when inspecting the data frame by eye. I will leave a TODO in the code to address this.

posydon/popsyn/GRB.py
ssbvr: I will comment out the option altogether, with a note to fix it.

posydon/visualization/plot_pop.py
ssbvr: Makes sense, for the other plotting libraries I implemented it when it was needed. I suggest that the first time someone needs it should implement it.

posydon/popsyn/rate_calculation.py
ssbvr: This is part of a large loop, that will execute every delta T ~ 100 Myr so we do not want to spam the user. This information will become clear when plotting rate densities.

posydon/popsyn/rate_calculation.py
ssbvr: I now have updated the docstring accordingly. Note that this method will not be accessible by the user, the user will use the `SyntheticPopulation` method `get_grb_rate_density` that will take care of computing the GRB rate density from `star_1` and `star_2` and return the corresponding rate densities for both stars and total, as well as the intrinsic and observable data frames (similar to the DCO population).

posydon/popsyn/rate_calculation.py
ssbvr: To be safe, I am just preventing future misuse of this method which could be renamed to a more genric name.

posydon/popsyn/rate_calculation.py
mkruckow: What you'd usually do in such cases is to define a flag before the loop and set it during the loop, when it happened at least once and make a printout after the loop has finished or only if the flag is still unset.

posydon/popsyn/rate_calculation.py
mkruckow: Here I'd like to point you to the document Konstantinos pinned in the developers slack channel:
In section 6 it recommends to start all private functions in a class with an underscore `_` to show, that this function should not be called from outside.

posydon/popsyn/GRB.py
mkruckow: The point here was not about the order, it is about the logic:
1. if it is not a float
2. if it is larger or equal to 0
3. if it is smaller or equal to 1

We should get the error, when 1. is not fulfilled, and when it is not in the range [0,1].
So here a third way to write that: `not isinstance(GRB_efficiency, float) or not (GRB_efficiency >= 0. and GRB_efficiency <= 1.)`

posydon/popsyn/rate_calculation.py
ssbvr: Appending empty arrays is only done for booking when saving and exporting the `index, z_formation, z_grb, w_ijk`. The end product or the user will not care about this dictionary and which redshift bin has empty arrays as they are dropped in the concatenation when exporting the full `index, z_formation, z_grb, w_ijk` arrays.

posydon/popsyn/rate_calculation.py
ssbvr: Correct, unfortunately, we never enforced this condition. Something I plan to do, we could do it together over the summer, is to enforce a shared Python formatting and syntax and have a GitHub action rejecting any unformatted PR. Then we will request everyone to set up the code editor to run the formatted on save for each file. I was playing around with the VS Code Python formatted and tested the `yapf` with `pep8` but is not matching the v1 syntax. I am waiting for @kkovlakas to get back to me with the exact setting used to format the v1 code. Once this is set we parse all files to set the proper formatting and eventually homogenize class and method names.

posydon/popsyn/GRB.py
ssbvr: done

mkruckow (dismissed):
There are a few todos left, otherwise fine.

mkruckow (approved):
There are a few todos left, otherwise fine.

kkovlakas (approved):

PR #104, opened 2023-06-23
add support to compute DCO formation channels

ssbvr:
This is a draft PR that discuss the changes we plan to do to study DCO formation channels
- [x] add a method to compute the formation channel
- [x] add a method to compute the channelwise merger efficiency 
- [x] add a method to display the channelwise merger rate density
- [x] add a method to display the channelwise intrinsic/detectable observable distributions

Code demo:

get DCO at formation and compute the formmation channel column.
```py
pop.get_dco_at_formation(S1_state='BH', S2_state='BH', formation_channels=True,
                         oneline_cols=['S1_natal_kick_array_0', 'S2_natal_kick_array_0'])
``` 

Compute and display the channelwise merger efficiency.
```py
pop.get_dco_merger_efficiency()
pop.plot_merger_efficiency(channels=True)
```
Screenshot 2023-06-23 at 12 00 17

Compute and display the channelwise merger rate density.
```py
pop.get_dco_merger_rate_density()
pop.plot_merger_rate_density(channels=True)
```
Screenshot 2023-06-23 at 12 00 39

Plot channelwise observable property distributions.
```py
# CE channel
channel = 'ZAMS_oRLO1_CC1_oRLO2_oCE2_CC2_CO_contact_END'
pop.plot_hist_dco_properties('m_chirp', intrinsic=True, detectable=True, channel=channel)
pop.plot_hist_dco_properties(['S1_spin','S2_spin'], intrinsic=True, detectable=True, channel=channel)
```
Screenshot 2023-06-23 at 14 11 07
Screenshot 2023-06-23 at 14 11 17

### DIFF
--- a/posydon/popsyn/synthetic_population.py
+++ b/posydon/popsyn/synthetic_population.py
@@ -558,7 +558,7 @@ class SyntheticPopulation:
         if "channel" in self.df_synthetic:
             channels = np.unique(self.df_synthetic['channel'])
             for ch in channels:
-                sel = (self.df_synthetic.loc[index,'channel'] == ch)
+                sel = (self.rates.get_data('channel', index) == ch)
                 rate = self.rates.compute_merger_rate_density(w_ijk[sel], z_merger[sel], observable='DCOs', sensitivity=sensitivity)
                 self.rate_density[ch] = rate
             
--- a/posydon/popsyn/synthetic_population.py
+++ b/posydon/popsyn/synthetic_population.py
@@ -6,6 +6,7 @@ e.g. with multiple metallicities
 __authors__ = [
     "Simone Bavera <Simone.Bavera@unige.ch>",
     "Kyle Akira Rocha <kylerocha2024@u.northwestern.edu>",
+    "Monica Gallegos-Garcia <monicagallegosgarcia2024@u.northwestern.edu>",
 ]
 
 import warnings
--- a/posydon/popsyn/synthetic_population.py
+++ b/posydon/popsyn/synthetic_population.py
@@ -549,8 +549,17 @@ class SyntheticPopulation:
 
         # compute rate density weights
         self.z_rate_density = self.rates.get_centers_redshift_bins()
-        self.rate_density = self.rates.compute_merger_rate_density(w_ijk, z_merger, observable='DCOs', sensitivity=sensitivity)
-        print(f'DCO merger rate density in the local Universe (z={self.z_rate_density[0]:1.2f}): {round(self.rate_density[0],2)} Gpc^-3 yr^-1')
+        self.rate_density = {}
+        total_rate = self.rates.compute_merger_rate_density(w_ijk, z_merger, observable='DCOs', sensitivity=sensitivity)
+        self.rate_density['total'] = total_rate
+        if "channel" in self.df_synthetic:
+            channels = np.unique(self.df_synthetic['channel'])
+            for ch in channels:
+                sel = (self.df_synthetic.loc[index,'channel'] == ch)
+                rate = self.rates.compute_merger_rate_density(w_ijk[sel], z_merger[sel], observable='DCOs', sensitivity=sensitivity)
+                self.rate_density[ch] = rate
+            
+        print(f'DCO merger rate density in the local Universe (z={self.z_rate_density[0]:1.2f}): {round(total_rate[0],2)} Gpc^-3 yr^-1')
 
         # export the intrinsic DCO population
         self.df_intrinsic = self.resample_synthetic_population(index, z_formation, z_merger, w_ijk, export_cols=export_cols)
--- a/posydon/popsyn/synthetic_population.py
+++ b/posydon/popsyn/synthetic_population.py
@@ -303,7 +303,7 @@ class SyntheticPopulation:
         else:
             raise ValueError('You already have a population stored in memory!')
 
-    def get_dco_at_formation(self, S1_state, S2_state, oneline_cols=None):
+    def get_dco_at_formation(self, S1_state, S2_state, oneline_cols=None, formation_channels=False):
         """Sort synthetic population, i.e. DCO at formation.
 
         Note: by default this function looks for the symmetric state
@@ -320,6 +320,12 @@ class SyntheticPopulation:
             into the synthetic population.
 
         """
+        # add channel column to oneline dataframe
+        if formation_channels:
+            if self.verbose:
+                print('Computing formation channels...')
+            self.get_formation_channels()
+            
 
         # to avoid the user making mistake automatically check the inverse of
         # the stellar states, since the df is already parsed this will not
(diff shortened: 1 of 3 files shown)


posydon/popsyn/synthetic_population.py
```diff
@@ -230,6 +231,10 @@ def parse(self, S1_state=None, S2_state=None, binary_state=None,
                     df_sel_met = pd.concat([df_sel_met, df_tmp])
                     del df_tmp
 
+            if k != 0:
```
mkruckow, line 234: Do you mean `k>0`? Here both statements are equivalent. But I'd say `k>0` is more understood as after the first iteration, while `k!=0` means not in the first iteration.

posydon/popsyn/synthetic_population.py
```diff
@@ -243,6 +248,11 @@ def parse(self, S1_state=None, S2_state=None, binary_state=None,
             df_sel_met_oneline = pd.read_hdf(file,  key='oneline')
             df_sel_met_oneline = df_sel_met_oneline.loc[sel_met]
             df_sel_met_oneline['metallicity'] = met
+
+            if k != 0:
```
mkruckow, line 252: same as above: consider to use `k>0` instead

posydon/popsyn/synthetic_population.py
```diff
@@ -392,7 +410,24 @@ def get_dco_merger_efficiency(self):
             efficiencies.append(eff)
             print(f'DCO merger efficiency at Z={met:1.2E}: {eff:1.2E} Msun^-1')
         self.met_merger_efficiency = np.array(self.met_merger_efficiency)
-        self.merger_efficiency = np.array(efficiencies)
+        self.merger_efficiency = {}
+        self.merger_efficiency['total'] = np.array(efficiencies)
```
mkruckow, line 414: The two lines of creating a dictionary and adding an entry there can be combined in one statement/command:
`self.merger_efficiency = {'total': np.array(efficiencies)}`

posydon/popsyn/synthetic_population.py
```diff
@@ -392,7 +410,24 @@ def get_dco_merger_efficiency(self):
...
+        # if the channel column is present compute the merger efficiency per channel
+        if "channel" in self.df_synthetic:
+            channels = np.unique(self.df_synthetic['channel'])
+            for ch in channels:
+                efficiencies = []
+                for met in self.met_merger_efficiency:
+                    sel = (self.df_synthetic['metallicity'] == met) & (self.df_synthetic['channel'] == ch)
+                    count = self.df_synthetic[sel].shape[0]
+                    if count > 0.:
```
mkruckow, line 423: Shouldn't `count` be an integer? Why is it compared to a float here?

posydon/popsyn/synthetic_population.py
```diff
@@ -392,7 +410,24 @@ def get_dco_merger_efficiency(self):
...
+                        underlying_stellar_mass = self.df_synthetic.loc[sel,'underlying_mass_for_met'].values[0]
+                        eff = count/underlying_stellar_mass
+                    else:
+                        eff = np.nan
```
mkruckow, line 427: I'm not sure whether putting `eff` to `nan` is a good idea here. Having a `count` of 0 only means, your efficiency is below `1/underlying_stellar_mass`, but you don't know where. I'd prefer if there would be a way to conserve this information of the upper limit.
I'd make two suggestions here:
- use `0.5/underlying_stellar_mass`, which is the mean in the unknown range
- use `-1/underlying_stellar_mass`, which gives the value, where the negativeness shows, that it is the upper limit instead of a value

posydon/popsyn/synthetic_population.py
```diff
@@ -515,8 +552,17 @@ def get_dco_merger_rate_density(self, export_cols=None,  working_dir='./', load_
 
         # compute rate density weights
         self.z_rate_density = self.rates.get_centers_redshift_bins()
-        self.rate_density = self.rates.compute_merger_rate_density(w_ijk, z_merger, observable='DCOs', sensitivity=sensitivity)
-        print(f'DCO merger rate density in the local Universe (z={self.z_rate_density[0]:1.2f}): {round(self.rate_density[0],2)} Gpc^-3 yr^-1')
+        self.rate_density = {}
+        total_rate = self.rates.compute_merger_rate_density(w_ijk, z_merger, observable='DCOs', sensitivity=sensitivity)
+        self.rate_density['total'] = total_rate
```
mkruckow, line 557: Again, define and assign the dictionary value at the same time, this is more efficient in python.

posydon/popsyn/synthetic_population.py
```diff
@@ -556,6 +602,35 @@ def get_dco_detection_rate(self, sensitivity='design_H1L1V1', export_cols=None,
...
+    def get_formation_channels(self):
+        """Get formation channel and add to df and df_online."""
+        
+        # loop through each binary
+        unique_binary_index = np.unique(self.df.index)
+        for index in unique_binary_index:
+
+            # get event column and information from interpolated classes
+            df_binary = self.df.loc[index,['event']].dropna()
+            df_binary_online = self.df_oneline.loc[index,['interp_class_HMS_HMS','interp_class_CO_HMS_RLO', 'interp_class_CO_HeMS']].dropna()
```
mkruckow, line 614: Consider to not hard code the interpolation classes, e.g. first generate a list with all which contain "interp_class".

mkruckow:
In general, I'd suggest to not use `met` as a shortcut of metallicity; better use `metal`
(at least I think of `met` being the past tense of `meet`, thus it is confusing to a condition is met)

posydon/popsyn/synthetic_population.py
```diff
@@ -230,6 +231,10 @@ def parse(self, S1_state=None, S2_state=None, binary_state=None,
                     df_sel_met = pd.concat([df_sel_met, df_tmp])
                     del df_tmp
 
+            if k != 0:
+                shift_index = max(np.unique(df_sel.index)) + 1 
+                df_sel_met.index += shift_index
```
mpgalleg, line 236: is it important to keep the old index as a separate column?

mpgalleg: Also, is there an easy way to reindex them from 0-n_binaries in self.df where there are duplicate index? Right now the binary index can be very large which is unnecessary but convenient. I couldn't find something straightforward shifting the index

posydon/popsyn/synthetic_population.py
ssbvr: This is not set to `0` or a negative number because this value is plotted in log scale, and at the moment, we have residual channels that might have discontinuous values and did not want to display sharp spikes. I am closing the issue for the moment and will address in the future when these values are used in an active science project.

posydon/popsyn/synthetic_population.py
ssbvr: No, it is not important to have the old index, but it might be useful in the feature for debugging purposes. Let's keep what we have for the moment and if needed we will change the code later.

mkruckow:
Please check whether this PR is still needed or whether all is already included in PR#103.

ssbvr:
Merged to PR#103.

PR #110, opened 2023-07-05
Single stars updates

kasdaglie:
This PR has been created in order to allow the initially single stars to have their own IMF. By doing this we allow the abillity of changing the IMF of the single stars and the binaries separately and also the mass of the single stars is not longer limited by the mass range limits of the primary binary mass and can now extend as low as the single star grids allow.


posydon/popsyn/independent_sample.py
```diff
@@ -358,3 +358,48 @@ def generate_secondary_masses(primary_masses,
...
+        if m1 <= 5 and m1 > 2 :
+            binary_fraction = 0.59
+        if m1<=9 and m1 > 5 :
+            binary_fraction = 0.76
+        if m1<= 16 and m1 > 9  : 
+            binary_fraction = 0.84
+        if m1 > 16:
+            binary_fraction = 0.94
+    else: 
+        pass
```
mkruckow: I'd suggest to better set here a default value instead of just putting a `pass`. Usually, the code shouldn't get here, but if it will, you'll crash on the next line, because the variable is not defined.

astroJeff: `pass` will be fine, so long as you first set a check to make sure that `binary_fraction_scheme` is in `binary_fraction_scheme_options`.

As a separate point, can you make sure all the spacing is consistent? There should not be a space before the `:` after each if condition above.

astroJeff (dismissed):

posydon/popsyn/independent_sample.py
mkruckow: > `pass` will be fine, so long as you first set a check to make sure that `binary_fraction_scheme` is in `binary_fraction_scheme_options`.

If you want to rely on the check on the `binary_fraction_scheme_options`, you won't need an `else` clause at all. Hence I interpreted the `else` block to be there as a fall back in case the check on the options fail, e.g. because on got disabled, while not removed from the `binary_fraction_scheme_options`.

posydon/popsyn/independent_sample.py
```diff
@@ -358,3 +358,49 @@ def generate_secondary_masses(primary_masses,
...
+        elif m1 <= 2  and m1 > 0.8:
+            binary_fraction = 0.4
+        elif m1 <= 5 and m1 > 2:
+            binary_fraction = 0.59
+        elif m1<=9 and m1 > 5 :
+            binary_fraction = 0.76
+        elif m1<= 16 and m1 > 9:
+            binary_fraction = 0.84
+        elif m1 > 16:
+            binary_fraction = 0.94
```
mkruckow, line 403: Could you raise an Error for m1<0.8, currently this case is currently not covered at all. (Did you tried, e.g. m1=0.7?)
May consider to convert the "pass" in case the scheme is not recognized into an Error raising, too.

kasdaglie: You are right, I will raise an error for this case.

mkruckow: It is better to use an else clause instead of another elif, you now missed the m1=0.8 case.

mkruckow (dismissed):

ssbvr (approved):

mkruckow (approved):

PR #113, opened 2023-07-07
Allowing for relative paths in the pipeline

mkruckow:
So far, the code was always adding a `/` in front, hence assuming, that only absolute paths are given in the ini file of the pipeline. Now it checks, whether the path is absolute or relative and deals with it accordingly.

### DIFF
--- a/bin/compress-mesa
+++ b/bin/compress-mesa
@@ -44,12 +44,12 @@ def set_up_test(args):
                         shutil.copytree(track_dirs[ind], os.path.join(
                             args.test_dir,
                             folder,
-                            track_dirs[ind].split("/")[-1]))
+                            os.path.split(track_dirs[ind])[1]))
                     else:
                         shutil.copy(track_dirs[ind], os.path.join(
                             args.test_dir,
                             folder,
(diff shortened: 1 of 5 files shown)


ssbvr (dismissed):

kkovlakas (dismissed):
Ok for now, but both these lines and the previous ones, are... unnecessarily complicated. First of all there is magic character `/` which is OS specific (e.g., in windows it will fail), and this is why we use `os.path.join` in the first place. So, many operations here could be done with os. E.g., taking the bottom level sir is done with `os.path.basename`.

mkruckow:
> Ok for now, but both these lines and the previous ones, are... unnecessarily complicated. First of all there is magic character `/` which is OS specific (e.g., in windows it will fail), and this is why we use `os.path.join` in the first place. So, many operations here could be done with os. E.g., taking the bottom level sir is done with `os.path.basename`.

Your are correct, we should have used the os functions in the first place. Hence, I changed it accordingly and scanned through the repo, where it should be changed, too.

ssbvr (approved):

kkovlakas (approved):

PR #115, opened 2023-07-17
Plot popsyn values over grid slices

ssbvr:
This PR introduces a powerful visualization tool to over-plot the population synthesis results over the grids slices.

Here is a short tutorial.

```py
from posydon.popsyn.synthetic_population import SyntheticPopulation
pop = SyntheticPopulation(path+'population_params.ini')

# HMS-HMS
# plot all grid slices
pop.plot_popsyn_over_grid_slice('HMS-HMS', 1e-04, plot_dir='./plot_grids/HMS-HMS/all/')
# plot only a subset of slices
pop.plot_popsyn_over_grid_slice('HMS-HMS', 1e-04, slices=[0.7], plot_dir='./plot_grids/HMS-HMS/all/', verbose=True)
# plot only a specific channel (SMT, CE, ...) and show a DCO property (e.g., S1_spin) 
pop.plot_popsyn_over_grid_slice('HMS-HMS', 1e-04, slices=[0.7], channel='ZAMS_oRLO1_CC1_oRLO2_CC2_CO_contact_END', 
                            plot_dir='./plot_grids/HMS-HMS/SMT/', prop='S1_spin', prop_range=[0.,0.5], verbose=True)
pop.plot_popsyn_over_grid_slice('HMS-HMS', 1e-04, slices=[0.7], channel='ZAMS_oRLO1_CC1_oRLO2_oCE2_CC2_CO_contact_END', 
                            plot_dir='./plot_grids/HMS-HMS/CE/', prop='S1_spin', prop_range=[0.,0.5])
# CO-HMS_RLO
pop.plot_popsyn_over_grid_slice('CO-HMS_RLO', 1e-04, slices=[14.66,30.], plot_dir='./plot_grids/CO-HMS_RLO/all/')
# CO-HeMS
pop.plot_popsyn_over_grid_slice('CO-HeMS', 1e-04, slices=[14.66,30.], plot_dir='./plot_grids/CO-HeMS/all/',
                            prop='S2_spin', prop_range=[0.,1.])
```

Screenshot 2023-07-28 at 17 14 42
Screenshot 2023-07-28 at 17 09 13
Screenshot 2023-07-28 at 17 09 26
Screenshot 2023-07-28 at 17 09 37
Screenshot 2023-07-28 at 17 09 52


posydon/visualization/plot_pop.py
```diff
@@ -251,4 +258,137 @@ def plot_hist_properties(x, df_intrinsic=None, df_observable=None,
...
\ No newline at end of file
+        plt.show()
+        
+        
+def plot_popsyn_over_grid_slice(pop, grid_type, met_Zsun, slices=None, channel=None, 
+                                plot_dir='./', prop=None, prop_range=None, 
+                                log_prop=False, alpha=0.3, s=5., 
+                                show_fig=True, save_fig=True, close_fig=True,
+                                verbose=False):
+
```
mkruckow, line 269: Please add doc-strings to all new defined functions, which explain the function, the variables, and the return. Having a tutorial here is nice, but in a year the tutorial will be forgotten, while the code remains.

posydon/visualization/plot_pop.py
```diff
@@ -251,4 +258,137 @@ def plot_hist_properties(x, df_intrinsic=None, df_observable=None,
...
+    # load grid
+    met = convert_metallicity_to_string(met_Zsun)
+    grid_path = os.path.join(PATH_TO_POSYDON_DATA, f'POSYDON_data/{grid_type}/{met}_Zsun.h5')
```
mkruckow, line 272: Shouldn't `PATH_TO_POSYDON_DATA` already include the directory `POSYDON_data`?

posydon/visualization/plot_pop.py
```diff
@@ -251,4 +258,137 @@ def plot_hist_properties(x, df_intrinsic=None, df_observable=None,
...
+            print('Computing formation channels...')
+        pop.get_formation_channels()
+        
+    if 'CO' in grid_path:
+        # compact object mass slices
+        def log_range(x_min,x_max,x_n):
+            return 10**np.linspace(np.log10(x_min),np.log10(x_max), x_n)
+        m_CO_min = 1.
+        m_CO_max = 51.32759630397827
+        m_CO_n = 23
```
mkruckow: Why are those values for the range hard coded? Wouldn't it be better to have them as arguments with default values like those ones, here?

posydon/visualization/plot_pop.py
```diff
@@ -251,4 +258,137 @@ def plot_hist_properties(x, df_intrinsic=None, df_observable=None,
...
+        m_COs = log_range(m_CO_min, m_CO_max, m_CO_n)
+        m_COs_edges = 10**(np.log10(np.array(m_COs)[:-1])+
+                                (np.log10(np.array(m_COs)[1:])-np.log10(np.array(m_COs)[:-1]))/2)
```
mkruckow: I'd simplify this to `m_COs_edges = 10**((np.log10(np.array(m_COs)[1:])+np.log10(np.array(m_COs)[:-1]))/2)`; usually it is even better to use `*0.5` instead of `/2`, to avoid that integer division might be used.

posydon/visualization/plot_pop.py
```diff
@@ -251,4 +258,137 @@ def plot_hist_properties(x, df_intrinsic=None, df_observable=None,
...
+        m2 = [0.]+m_COs_edges.tolist()+[55.]
```
mkruckow: Again, don't hard code values as long as they aren't fixed limits.
So, I'd either expect:
- in the case of having limits: `m2 = [0.]+m_COs_edges.tolist()+[np.finfo(a.dtype).max]`
- in the case of expanding the range: `m2 = [10**(2*np.log10(m_COs_edges[0])-np.log10(m_COs_edges[1]))]+m_COs_edges.tolist()+[10**(2*np.log10(m_COs_edges[-1])-np.log10(m_COs_edges[-2]))]` here it would be good to add an catch in case the array don't have 1 or 0 values.

posydon/visualization/plot_pop.py
```diff
@@ -251,4 +258,137 @@ def plot_hist_properties(x, df_intrinsic=None, df_observable=None,
...
+        vars = m_COs
+        fname = 'grid_m_%1.2f.png'
+        title = '$m_\mathrm{CO}=%1.2f\,M_\odot$'
+        slice_3D_var_str='star_2_mass'
+    elif 'HMS-HMS' in grid_path:
+        # mass ratio slices
+        qs = np.linspace(0.05,1.,20)
+        qs[-1] = 0.99
```
mkruckow: Again, it would be better not not hard code this.

posydon/visualization/plot_pop.py
```diff
@@ -251,4 +258,137 @@ def plot_hist_properties(x, df_intrinsic=None, df_observable=None,
...
+        raise ValueError('Grid type not supported!')
+
+
+    for i, var in enumerate(vars):
+        if slices is not None and round(var,2) not in np.around(slices,2):
+            if verbose:
+                print(f'Skipping {round(var,2)}')
+            continue
+        if 'HMS-HMS' in grid_path:
+            slice_3D_var_range = (var-2.5e-2,var+2.5e-2)
```
mkruckow: This don't need to be hard coded as `2.5e-2` either. By the way, this is already questionable in the case of q=0.99, which would now get an overlap with q=0.95.

posydon/visualization/plot_pop.py
```diff
@@ -251,4 +258,137 @@ def plot_hist_properties(x, df_intrinsic=None, df_observable=None,
...
+            # select popsynth binaries in the given mass ratio range
+            sel = (pop.df['metallicity'] == met_Zsun*Zsun) & (pop.df['event'] == 'ZAMS')
+            q = pop.df['S2_mass'].values/pop.df['S1_mass'].values
+            q[q>1] = 1./q[q>1]
+            mask = sel & (q>=slice_3D_var_range[0]) & (q<=slice_3D_var_range[1])
+        elif 'CO' in grid_path:
+            if i == len(m2):
```
mkruckow, line 318: I'd suggest to change it into `if i >= len(m2):` to although capture unintended cases.

posydon/visualization/plot_pop.py
```diff
@@ -251,4 +258,137 @@ def plot_hist_properties(x, df_intrinsic=None, df_observable=None,
...
+                plt.scatter(log10_m1, log10_p, c=prop_values, s=s, vmin=vmin, vmax=vmax, marker='v', cmap='viridis', alpha=alpha)
+                if prop in DEFAULT_LABELS:
+                    if log_prop:
+                        label = DEFAULT_LABELS[prop][1]
+                    else:
+                        label = DEFAULT_LABELS[prop][0]
+                else:
+                    label = prop
+                    if log_prop:
+                        label = 'log10_'+label
```
mkruckow, line 378: This should read `label = 'log_10('+label+')'`

mkruckow:
I'd suggest to reverse the plotting order: the population in the background and the grid on top, because the grid has guarantied free space, while a population may be so dense, that the grid below is invisible.

posydon/visualization/plot_pop.py
ssbvr: no, it does not. We made this choice in v1 to be the working directory to where the `POSYDON_data` tar file will be downloaded from Zenodo and the uncompressed to the `POSYDON_data` directory. Though I agree this is confusing, once we revisit how to store and import paths from a `config.py` file reading `.env` we can edit this.

posydon/visualization/plot_pop.py
ssbvr: fixed. I am now reading this information from the grid.intial_values

posydon/visualization/plot_pop.py
ssbvr: fixed

posydon/visualization/plot_pop.py
ssbvr: Fixed, I did it this because on the moment it was annoying to write down the lines of code that compute the ranges dinamically as you need to convert from from array to string to do the manipulation and then manyally add the edges.

posydon/visualization/plot_pop.py
ssbvr: Will not address. Properties without specified default legend are always printed with the variable name string or with `log10_` in front.

ssbvr:
> I'd suggest to reverse the plotting order: the population in the background and the grid on top, because the grid has guarantied free space, while a population may be so dense, that the grid below is invisible.

This cannot be easely implemented as the figure object is created by the plot2D method.

posydon/visualization/plot_pop.py
ssbvr: done

posydon/visualization/plot_pop.py
ssbvr: done

posydon/visualization/plot_pop.py
ssbvr: The entire plotting file methods needs docstrings. I think we can postpone this to a later stage as all these methods are evolving quickly and might now be the final product.

mkruckow:
> > I'd suggest to reverse the plotting order: the population in the background and the grid on top, because the grid has guarantied free space, while a population may be so dense, that the grid below is invisible.
> 
> This cannot be easely implemented as the figure object is created by the plot2D method.

It should be easy by simply specifying the zorder. The default zorder of the data should be 2 or 1, hence try values of 1.9 or 0.9 to move the new data below the already existing one.

ssbvr:
> > > I'd suggest to reverse the plotting order: the population in the background and the grid on top, because the grid has guarantied free space, while a population may be so dense, that the grid below is invisible.
> > 
> > 
> > This cannot be easely implemented as the figure object is created by the plot2D method.
> 
> It should be easy by simply specifying the zorder. The default zorder of the data should be 2 or 1, hence try values of 1.9 or 0.9 to move the new data below the already existing one.

done, I updated the plots above to reflect the change.

posydon/visualization/plot_pop.py
mkruckow: somehow, this was not part of your addressing commit.

mkruckow (approved):

astroJeff (approved):

PR #116, opened 2023-07-20
Evaluate if interp

philipp-rajah:
I've added functionality to assess interpolation performance, this includes modularized violin plots, confusion matrices and 2d slices that show where errors are present in the grid. Note that the slices are a work in progress and probably require another hour or two of work.


ssbvr:
This PR is now part of #124

PR #122, opened 2023-08-07
introducing step_CO_HeMS_RLO

ssbvr:
This PR solves an outstanding issue from v1; namely, we want eccentric CO-stripped_He binaries emerging from >Zsun HMS-HMS grids to be redirected to step_detached instead of being evolved with circular CO-HeMS grids, which are supposed to be used only in the case of post-common envelope step.

Since I was working on this code, I also implemented an automatic check detection of grid boundaries in step_MESA for all grids.

Here is an example of how an eccentric BH-stripped_He system is redirected to the detached step.
Screenshot 2023-08-07 at 15 45 09

Since I could not easily find a system that after the detached step would lead to CO-HeMS_RLO. I artificially crafted the initial conditions and show here that the changes to POSYDON indeed evolve such systems.
Screenshot 2023-08-07 at 15 45 51

I tested that the code runs correctly on a 1K binary pop synth model at 8 metallicities. I found the following unexpected, but correct, way to hit the step_CO_HeMS_RLO. After the step_CE we map to a NS-He-star where the secondary star has mass below ou minimum He-star in the grids (this is ok as the star will become a WD, not supported in v2). The code correctly redirects to step_detach which correctly finds RLO as low mass He-star expands mapping to the step_CO-HeMS_RLO, in this case we map to initail RLO probably suggesting that the evolved He-satar is more compact than an evolved ZAHeMS RLO in the grid.
Screenshot 2023-08-08 at 10 39 15

EXTRA:
- [x] fixed bug for `interpolation_method = 'nearest_neighbour'`. The interpolated/classified MODEL properties were not stored correctly in MESA step.

### DIFF
--- a/posydon/binary_evol/MESA/step_mesa.py
+++ b/posydon/binary_evol/MESA/step_mesa.py
@@ -1359,7 +1359,6 @@ class CO_HMS_RLO_step(MesaGridStep):
             if self.binary.star_1.state == 'H-rich_Central_C_depletion':
                 self.binary.event = 'CC1'
                 return
-            # super().__call__(binary)
         elif (state_1 in ["WD", "NS", "BH"] and (state_2 in FOR_RLO_STATES)
                 and event == "oRLO2"):
             self.flip_stars_before_step = True
@@ -1367,7 +1366,6 @@ class CO_HMS_RLO_step(MesaGridStep):
             if self.binary.star_2.state == 'H-rich_Central_C_depletion':
                 self.binary.event = 'CC2'
                 return
-            # super().__call__(binary)
         else:
             raise ValueError(
                 'The star_1.state = %s, star_2.state = %s, binary.state = %s, '
@@ -1446,18 +1444,16 @@ class CO_HeMS_RLO_step(MesaGridStep):
                 and (state_1 in CO_He_STATES) and event == "oRLO1"):
             self.flip_stars_before_step = False
             # catch and redirect double core collapse, this happens if q=1:
-            if self.binary.star_1.state == 'H-rich_Central_C_depletion':
+            if self.binary.star_1.state == 'stripped_He_Central_C_depletion':
                 self.binary.event = 'CC1'
                 return
-            # super().__call__(binary)
         elif (state_1 in ["WD", "NS", "BH"] and (state_2 in CO_He_STATES)
                 and event == "oRLO2"):
             self.flip_stars_before_step = True
             # catch and redirect double core collapse, this happens if q=1:
-            if self.binary.star_2.state == 'H-rich_Central_C_depletion':
+            if self.binary.star_2.state == 'stripped_He_Central_C_depletion':
                 self.binary.event = 'CC2'
                 return
-            # super().__call__(binary)
         else:
             raise ValueError(
                 'The star_1.state = %s, star_2.state = %s, binary.state = %s, '
--- a/posydon/binary_evol/MESA/step_mesa.py
+++ b/posydon/binary_evol/MESA/step_mesa.py
@@ -1461,7 +1461,7 @@ class CO_HeMS_RLO_step(MesaGridStep):
         else:
             raise ValueError(
                 'The star_1.state = %s, star_2.state = %s, binary.state = %s, '
-                'binary.event = %s and not CO - HMS - oRLO1/oRLO2!'
+                'binary.event = %s and not CO - HeMS - oRLO1/oRLO2!'
                 % (state_1, state_2, state, event))
         # redirect if outside grids
         if ((not self.flip_stars_before_step and
--- a/posydon/binary_evol/MESA/step_mesa.py
+++ b/posydon/binary_evol/MESA/step_mesa.py
@@ -787,24 +787,25 @@ class MesaGridStep:
             for key in key_post_processed:
                 setattr(star, key, cb.final_values['S%d_%s' % (k+1, key)])
 
-        # update nearest neighbor core collapse quantites
+        # update nearest neighbor core collapse quantites                            
         if interpolation_class != 'unstable_MT':
             for MODEL_NAME in MODELS.keys():
                 for i, star in enumerate(stars):
-                    if not stars_CO[i]:
+                    if (not stars_CO[i] and 
+                        cb.final_values[f'S{i+1}_{MODEL_NAME}_CO_type'] != 'None'):
                         values = {}
                         for key in ['state', 'SN_type', 'f_fb', 'mass', 'spin',
                                     'm_disk_accreted', 'm_disk_radiated']:
-                            if key == 'state': 
-                                key = 'CO_type'
-                            values[key] = cb.final_values[f'S{i+1}_{MODEL_NAME}_{key}']
-                        state = cb.final_values[f'S{i+1}_{MODEL_NAME}_CO_type'] 
-                        if state is None or state == 'None':
-                            # privent to any quantities for star that did not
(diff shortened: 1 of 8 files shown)


posydon/binary_evol/MESA/step_mesa.py
```diff
@@ -1240,6 +1240,14 @@ def __init__(self, metallicity=1., grid_name=None, *args, **kwargs):
                          *args, **kwargs)
         # special stuff for my step goes here
         # If nothing to do, no init necessary
+
```
ka-rocha, line 1244: Nice removing hard coded limits!

posydon/binary_evol/MESA/step_mesa.py
```diff
@@ -1240,6 +1240,14 @@ def __init__(self, metallicity=1., grid_name=None, *args, **kwargs):
                          *args, **kwargs)
         # special stuff for my step goes here
         # If nothing to do, no init necessary
+        
+        # load grid boundaries
+        self.m1_min = min(self._psyTrackInterp.grid.initial_values['star_1_mass'])
+        self.m1_max = max(self._psyTrackInterp.grid.initial_values['star_1_mass'])
+        self.q_min = 0.05 # can be computed m2_min/m1_min
```
mkruckow, line 1248: Do we need a fixed value here? Because `q_min` will be a function depending on `m1`. Hence, I'm wondering, where it would be better to not specify a `q_min`, but instead provide a function `q_min(m1)`?

posydon/binary_evol/MESA/step_mesa.py
```diff
@@ -1251,29 +1259,39 @@ def __call__(self, binary):
...
             self.flip_stars_before_step = False
             super().__call__(self.binary)
-        elif (state_1 == 'H-rich_Core_H_burning'
-              and state_2 == 'H-rich_Core_H_burning'
-              and event == 'ZAMS' and mass_ratio > 1. and p <= p_max_HMS_grid):
+        elif (state_1 == 'H-rich_Core_H_burning' and 
+              state_2 == 'H-rich_Core_H_burning' and 
+              event == 'ZAMS' and
+              self.m1_min <= m2 <= self.m1_max and
+              1./self.q_min >= mass_ratio > self.q_max and
```
mkruckow: Here we need a `1./self.q_max`, too. Just because we use a defaulted `q_max=1` doesn't mean it will always be that way. Alternatively, we could say `self.q_min <= 1./mass_ratio <= self.q_max`, which I'd find even more intuitive.

posydon/binary_evol/MESA/step_mesa.py
```diff
@@ -1251,29 +1259,39 @@ def __call__(self, binary):
...
+        # redirect if outside grid
+        elif (state_1 == 'H-rich_Core_H_burning' and 
+              state_2 == 'H-rich_Core_H_burning' and 
+              event == 'ZAMS' and 
+              p > self.p_max):
             self.binary.event = 'redirect'
             return
-        elif (state_1 == 'H-rich_Central_C_depletion'):     # redirect if CC1
+        # redirect if CC1
+        elif (state_1 == 'H-rich_Central_C_depletion'):
```
mkruckow, line 1291: Just a question about the stars, where we stop the evolution before carbon depletion: do they still get to the stellar state 'H-rich_Central_C_depletion'?

posydon/binary_evol/MESA/step_mesa.py
```diff
@@ -1251,29 +1259,39 @@ def __call__(self, binary):
...
+        # redirect if CC1
+        elif (state_1 == 'H-rich_Central_C_depletion'):     
             self.binary.event = 'CC1'
             return
-        elif (state_2 == 'H-rich_Central_C_depletion'):     # redirect if CC2
+        # redirect if CC2
+        elif (state_2 == 'H-rich_Central_C_depletion'):     
             self.binary.event = 'CC2'
             return
         else:
```
mkruckow, line 1298: Why don't we send stars with masses outside the grid to redirect as well? Our single star models cover a larger range in masses, especially at the low mass end.

posydon/binary_evol/MESA/step_mesa.py
```diff
@@ -1325,36 +1355,125 @@ def __call__(self, binary):
         if (state_2 in ["WD", "NS", "BH"]
```
mkruckow, line 1357: Could you add a TODO to import the list of compact objects from flow_chart.py as well (please do it for all cases with a static list `["WD", "NS", "BH"]`.

posydon/binary_evol/MESA/step_mesa.py
ssbvr: You are right about what we have in reality. Is a mass range for m1 and one for `m2=m1*q` at `0.5 < m2/Msun <  m1_max*0.99`. Is it better to transform the q condition to an m2 condition?  Else I could add the extra condition `m2 = m1*q > 0.5 Msun`, which is the cut we add in practice.

posydon/binary_evol/MESA/step_mesa.py
ssbvr: Correct, thank you.

posydon/binary_evol/MESA/step_mesa.py
ssbvr: No, they will not. They will likely get `H-rich_Central_He_depleted` as we no longer have a state `near_C_depletion`. But then because we know the binary system reached the end of the MESA evolution successfully (no `initial_MT` or `unstable_MT`), we update the binary event to `CC1` or `CC2`, and the `step_SN` will collapse such binary (see `STAR_STATES_CC` in `binary_evol/flow_chart.py`). I vaguely remember why this redirect condition was introduced. How do we hit HMS-HMS grids while one of the stars was previously evolved? This redirection might be obsolete.

ssbvr: @devimisra @ezapartas do you remember why we introduced this redirection in the HMS-HMS step? It thought it had to do with q=1 binaries in v1 reaching two CC1 one after the other. But a `CO--H_rich_Central_He_depleted` binary would not reach this step, it will be redirected by `step_detached` to `CC2`.

posydon/binary_evol/MESA/step_mesa.py
ssbvr: I though we used to do this, but now I suspect we did remove this as step_detached would like to do matching with star properties that are not yet defined. Maybe with the redesign of `step_detached`, we can now handle mapping to a single star only with respect to the ZAMS mass when t=0. What do you think @ezapartas? @kasdaglie, this is relevant for your project as well, right now a HMS-HMS binary system outside the maximal orbital period is not redirected to the `step_detached` but in principle, it should.

posydon/binary_evol/MESA/step_mesa.py
ssbvr: done

posydon/binary_evol/MESA/step_mesa.py
mkruckow: Yes, the `m2 = m1*q > 0.5 Msun` was, what I had in mind. Just to prevent you from doing a mistake here, there is still no global q_min, which you could derive from m1_min and m1_max, because we need to get a larger q_min in case m2 would get below 0.5Msun.

mkruckow (dismissed):

astroJeff (approved):

mkruckow (approved):

PR #123, opened 2023-08-09
Fix Typo in scrubbing.py

mkruckow:
Use h2_colnames instead of h1_colnames before h2 is checked.


ssbvr (approved):
Do you have any idea how this bug affected processed data in the past?

mkruckow:
> Do you have any idea how this bug affected processed data in the past?

This didn't had any effect on the values of post processed data. The typo is just in the check to avoid KeyErrors. Hence, would would have gotten more key errors in the past.

ezapartas (approved):

astroJeff (approved):

kkovlakas (approved):

PR #129, opened 2023-08-15
Tilt tracking

chase-kimball:
This pull request adds combined tilt tracking to step_SN, as well as the required angles to binarystar and singlestar. 
After the first SN1, the spin-orbit misalignment for both objects is just the tilt from SN1. At SN2, we assume the remaining non-compact object realigns with the orbital angular momentum. After SN2, it's spin-orbit misalignment is the tilt from SN1, but the compact object formed in SN2 has a combined misalignment from both SN:

Taking $\hat{x}$ and $\hat{z}$ to be the X and Z unit vectors of the SN1 coordinate system, and $\hat{L}=\hat{z}$ and $\hat{L}'$ to be the unit vectors in the direction of the orbital angular momentum pre-SN1 and post-SN2 respectively, we can write:

$\hat{L}' =  R_{\hat{x}' }(\theta_{SN2})R_{\hat{x} }(\theta_{SN1})\hat{L}$

where 

$\hat{x}' = R_{\hat{z}'} (f_{SN2} - f_{SN1})\hat{x}$
and

$\hat{z}' = R_{\hat{x}}(\theta_{SN1})\hat{z}$

are the transformed unit vectors in the SN2 coordinate frame, 
$R_{\vec{u}}(t)$ denotes rotation about the vector $\vec{u}$ by an angle $t$, and $f_{SN1}$ and $f_{SN2}$ are the true anomalies of the binary just before the first and second SN, respectively. Then the angle between the first compact object's spin and the binary orbital angular momentum is given by
$\cos(\Omega_{1,SN2}) = \hat{z} \cdot \hat{L}' $

### DIFF
--- a/posydon/binary_evol/SN/step_SN.py
+++ b/posydon/binary_evol/SN/step_SN.py
@@ -1425,7 +1425,10 @@ class StepSN(object):
             # Eq 15, Wong, T.-W., Valsecchi, F., Fragos, T., & Kalogera, V. 2012, ApJ, 747, 111
             # orbital separation at the time of the exlosion
             rpre = Apre * (1.0 - epre * np.cos(E_ma))
-
+            
+            true_anomaly = 2 * np.arctan(
+                np.sqrt((1 + epre) / (1 - epre)) * np.tan(E_ma / 2)
+            )
             # load constants in CGS
             G = const.standard_cgrav
 
@@ -1625,18 +1628,43 @@ class StepSN(object):
             # update the binary object which was bound at least before the SN
             if flag_binary:
                 # update the tilt
-                if binary.event == "CC1":
-                    binary.star_1.spin_orbit_tilt = tilt
-                elif binary.event == "CC2":
-                    binary.star_2.spin_orbit_tilt = tilt
-                else:
-                    raise ValueError("This should never happen!")
-
-                # compute new orbital period before reseting the binary properties
+                #Check if this is the first SN
+                first_SN = binary.true_anomaly_first_SN is None
 
                 for key in BINARYPROPERTIES:
-                    if key != 'nearest_neighbour_distance':
+                    if key != 'nearest_neighbour_distance' and key != 'true_anomaly_first_SN':
                         setattr(binary, key, None)
+
+                if first_SN:
+                    # update the tilt
+                    binary.star_1.spin_orbit_tilt_first_SN = tilt
+                    binary.star_2.spin_orbit_tilt_first_SN = tilt
+                    binary.true_anomaly_first_SN = true_anomaly
+                else:
+                    if binary.event == 'CC2':
+                        # Assume progenitor has aligned with the preSN orbital angular momentum
+                        binary.star_2.spin_orbit_tilt_second_SN = tilt
+                        binary.star_1.spin_orbit_tilt_second_SN = get_combined_tilt(
+                            tilt_1 = binary.star_1.spin_orbit_tilt_first_SN, 
+                            tilt_2 = tilt, 
+                            true_anomaly_1 = binary.true_anomaly_first_SN, 
+                            true_anomaly_2 = true_anomaly
+                            )
+                        binary.true_anomaly_SN2 = true_anomaly
+                    elif binary.event == 'CC1':
+                        # Assume progenitor has aligned with the preSN orbital angular momentum
+                        binary.star_1.spin_orbit_tilt_second_SN = tilt
+                        binary.star_2.spin_orbit_tilt_second_SN = get_combined_tilt(
+                            tilt_1 = binary.star_1.spin_orbit_tilt_first_SN, 
+                            tilt_2 = tilt, 
+                            true_anomaly_1 = binary.true_anomaly_first_SN, 
+                            true_anomaly_2 = true_anomaly
+                            )
+                        binary.true_anomaly_second_SN = true_anomaly
+                    else:
+                        raise ValueError("This should never happen!")
+
+                # compute new orbital period before reseting the binary properties
 
                 binary.state = "detached"
                 binary.event = None
@@ -1650,18 +1678,20 @@ class StepSN(object):
                     binary.separation, binary.star_1.mass, binary.star_2.mass)
                 binary.orbital_period = new_orbital_period
                 binary.mass_transfer_case = 'None'
-            else:
-                # update the tilt
-                if binary.event == "CC1":
-                    binary.star_1.spin_orbit_tilt = np.nan
-                elif binary.event == "CC2":
-                    binary.star_2.spin_orbit_tilt = np.nan
-                else:
-                    raise ValueError("This should never happen!")
 
+            else:
                 for key in BINARYPROPERTIES:
                     if key != 'nearest_neighbour_distance':
                         setattr(binary, key, None)
+                # update the tilt
+                if first_SN:
+                    binary.star_1.spin_orbit_tilt_SN1 = np.nan
+                    binary.star_2.spin_orbit_tilt_SN1 = np.nan
+                else:
+                    binary.star_1.spin_orbit_tilt_SN2 = np.nan
+                    binary.star_2.spin_orbit_tilt_SN2 = np.nan
+
+             
                 binary.state = "disrupted"
                 binary.event = None
                 binary.separation = np.nan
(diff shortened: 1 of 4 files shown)


posydon/binary_evol/binarystar.py
```diff
@@ -162,6 +162,11 @@ def __init__(self, star_1=None, star_2=None, index=None, properties=None,
             self.inspiral_time = None
         if not hasattr(self, 'mass_transfer_case'):
             self.mass_transfer_case = 'None'
+
+        if not hasattr(self, 'true_anomaly_first_SN'):
```
ssbvr, line 166: For consistency, this quantity should be stored only once for each star in the star object as a fith value to the the natal kick array. Given that this will be the fifth value associated to the kicks, to avoid confusions, maybe we should get rid of the natal kick array and have proper names for each variable stored separatly. Also in the future the user might not want to store the entire natal kick array, but say, e.g. only the magnitude of the kick.

posydon/binary_evol/singlestar.py
```diff
@@ -137,8 +137,10 @@ def __init__(self, **kwargs):
         # these quantities are updated in step_SN.py
         if not hasattr(self, 'natal_kick_array'):
             self.natal_kick_array = [None] * 4
-        if not hasattr(self, 'spin_orbit_tilt'):
-            self.spin_orbit_tilt = None
+        if not hasattr(self, 'spin_orbit_tilt_first_SN'):
```
ssbvr, line 140: Is this nomenclature really needed? We do not allow to have references of another star in a single stellar object. The way of keeping track of this is in the step SN, where you access the S1 and S2 thorugh the binary object.

posydon/binary_evol/SN/step_SN.py
```diff
@@ -1623,21 +1626,46 @@ def SNCheck(
                                   verbose=self.verbose)
 
             # update the binary object which was bound at least before the SN
+            #Check if this is the first SN
+            first_SN = binary.true_anomaly_first_SN is None
```
ssbvr: A better way to check if this is the first or second supernova would be to check the stellar.state of the companion star. If the companion is already a compact object, i.e. `star.state in ['massless_remnant', 'WD', 'NS', 'BH']`, then this is the second supernova.

posydon/binary_evol/singlestar.py
ssbvr: oh, maybe I understood it wrong. You want to store the initial tilt and then the updated one after the tilt is changed by the second superonova. But then how would the user know witch one to use, e.g., to compute the chi_eff? Wouldn't it be easier to have just one variable named tilt and then add the tilt history in natal_kick_array?

posydon/binary_evol/SN/step_SN.py
```diff
@@ -1726,7 +1756,82 @@ def generate_kick(self, star, sigma):
             Vkick = 0.0
 
         return Vkick
+    def rotate(axis, angle):
+
+        """Generate rotation matrix to rotate a vector about an arbitrary axis 
+            by a given angle
```
mkruckow: The rotate function is a general rotation. Hence, I'm wondering about two things:
1. Shouldn't there be a similar function already existing? I haven't seen one in numpy, but it looks like scipy has. Thus, may consider to not define this ourselves.
2. If we define it ourselves, is it specific to the SN_step or should we better put it in the common_functions.py in utils?

posydon/binary_evol/SN/step_SN.py
astroJeff: The problem with this is that we use the term `massless_remnant` when referring to both a binary that has been disrupted as well as an initially single star. The ambiguity will need to be resolved in the future. For now though, let's use a different method. @chase-kimball, can you add two flags to the binary object that says whether the star has gone through SN1 and SN2? It's verbose but clear.

posydon/binary_evol/SN/step_SN.py
astroJeff: If Chase has already built the function and it works, I don't mind keeping what we have put together already. But I agree it should go in common_functions.py.

posydon/binary_evol/binarystar.py
astroJeff: I never liked the natal kick array, but until we decide to make a change to it, we are stuck with it. Nevertheless, I think it is fine to keep it as is (with just the four parameters) and add these two new variables to the binary object. We may want to come up with a better option in the future, but for now, I'm fine with this nomenclature.

posydon/binary_evol/singlestar.py
astroJeff: We will need to be explicitly clear in the tutorials/documentation: chi_eff is calculated from spin_orbit_tilt_second_SN.

mkruckow: To be more explicit on the naming, we can may include an after there, hence `spin_orbit_tilt_after_first_SN` and `spin_orbit_tilt_after_second_SN`.
Btw. I just thought of a possible science case for using the the tilt after the first SN: How about mergers of NS/BH with a WD/sdb star? Here we should get only one SN, hence only one tilt. Those could be potential LISA sources beside the WD+WD mergers. Similarly, I can imaging to include AIC in the future, where we have two NS merging, but only one went through a SN.
An other option to make it easy for a user could be to introduce a variable called `final_spin_orbit_tilt`, which gets updated during the evolution.

posydon/binary_evol/SN/step_SN.py
```diff
@@ -1726,7 +1756,82 @@ def generate_kick(self, star, sigma):
...
+            combined_tilt: float
+                Angle, in radians, between the spin and orbital angular momentum
+                after SN2
+        """
+        z_prime = rotate((1,0,0), tilt_1).dot((0,0,1))
+        x_prime = rotate(z_prime, true_anomaly_2-true_anomaly_1).dot((1,0,0))
 
+        cos_tilt = np.dot((0,0,1),rotate(x_prime, tilt_2).dot(z_prime))
+        combined_tilt = np.arccos(cos_tilt)
+        return combined_tilt
```
ZepeiX, line 1795: Here for the calculation of combined tilt, we need to consider the direction of the two tilts. A simple case that the two tilts are in the same plane, if the two tilts are equal, the combined tilt can be 0 or 2 depending on the direction. 
The tilt angle we defined does not have the sign for the direction. The first thought is the information we need is in phi.

mkruckow: To be a bit more precise: the sign of sin(phi) will correspond to the not tracked sign of the tilt. It can be understood via Vkz, which gets its sign from sin(phi) but is only used as Vkz**2 in the tilt calculation.

tassos25:
@chase-kimball @ZepeiX @mkruckow Can you please push on finalizing this? My understanding is that a minor change is required and then we can accept it. This PR is also pending for a looooong time.

chase-kimball:
@tassos25 Yes I will on finalizing this ASAP.

posydon/binary_evol/SN/step_SN.py
```diff
@@ -1624,19 +1627,44 @@ def SNCheck(
...
+                            tilt_2 = tilt, 
+                            true_anomaly_1 = binary.true_anomaly_first_SN, 
+                            true_anomaly_2 = true_anomaly
+                            )
+                        binary.true_anomaly_SN2 = true_anomaly
+                    elif binary.event == 'CC1':
+                        # Assume progenitor has aligned with the preSN orbital angular momentum
+                        binary.star_1.spin_orbit_tilt_second_SN = tilt
+                        binary.star_2.spin_orbit_tilt_second_SN = get_combined_tilt(
+                            tilt_1 = binary.star_1.spin_orbit_tilt_first_SN,
```
mkruckow, line 1662: Shouldn't this be `binary.star_2.spin_orbit_tilt_first_SN`.
In the current version, the tilt of the first SN is the same for both stars, hence it won't change the values.

posydon/binary_evol/SN/step_SN.py
```diff
@@ -1684,11 +1684,11 @@ def SNCheck(
...
                 if first_SN:
-                    binary.star_1.spin_orbit_tilt_SN1 = np.nan
-                    binary.star_2.spin_orbit_tilt_SN1 = np.nan
+                    binary.star_1.spin_orbit_tilt_first_SN = np.nan
+                    binary.star_2.spin_orbit_tilt_first_SN = np.nan
                 else:
-                    binary.star_1.spin_orbit_tilt_SN2 = np.nan
-                    binary.star_2.spin_orbit_tilt_SN2 = np.nan
+                    binary.star_1.spin_orbit_tilt_first_SN = np.nan
+                    binary.star_2.spin_orbit_tilt_first_SN = np.nan
```
mkruckow: Shouldn't those be `spin_orbit_tilt_second_SN` in the else part?

posydon/binary_evol/singlestar.py
chase-kimball: I've updated this as @mkruckow suggested with spin_orbit_tilt_after_first_SN etc.

posydon/binary_evol/SN/step_SN.py
chase-kimball: I've added a flag binary.first_SN_already_occurred to binarystar.py

posydon/binary_evol/SN/step_SN.py
chase-kimball: As suggested, I've moved rotate() this to common_functions.py

posydon/binary_evol/SN/step_SN.py
chase-kimball: Thanks for catching this! I've explicitly tracked the sign of tilt according to Vkz, with Vkz < 0 corresponding to -pi< tilt < 0, Vkz >0 corresponding to 0<tilt<pi. get_combined_tilt should now add the tilts correctly

posydon/binary_evol/SN/step_SN.py
chase-kimball: @mkruckow Thanks for catching this. Fixed.

mkruckow:
@chase-kimball thanks for taking care of the open issues.
Before we merge it I'd suggest to merge development into your branch to resolve any merge conflicts, which might be there because development has evolved significantly since the branch out. Currently the diff between the branches shows a lot of files (more than touched in this PR).

We should may check a cleaner comparison as the one linked on top of this PR, may check https://github.com/POSYDON-code/POSYDON/compare/development...chase-kimball:POSYDON:development

astroJeff:
Update (Feb 1): @chase-kimball  has told me the PR is finally ready for review. However, as @mkruckow  pointed out above, it needs to be rebased with the latest in development.

chase-kimball:
Will do @astroJeff @mkruckow

mkruckow (dismissed):

astroJeff (approved):

mkruckow (approved):

PR #132, opened 2023-08-18
Konst removing init imports

kkovlakas:
This PR
- removes imports from all `__init__.py` files, in hope of resolving future cyclic imports
- corrects import statements where aliases were used (e.g., `from posydon.binary_evol import SimulationProperties` is not valid anymore, since `SimulationProperties` is in `posydon.binary_evol.simulationproperties`)
- removes a significant fraction of unused imports in all POSYDON scripts
- adds a TODO note to a possible mistake

### DIFF
--- a/posydon/interpolation/__init__.py
+++ b/posydon/interpolation/__init__.py
@@ -1,2 +1,0 @@
-from .interpolation import GRIDInterpolator, fscale
-from .IF_interpolation import IFInterpolator
--- a/posydon/tests/utils/test_configfile.py
+++ b/posydon/tests/utils/test_configfile.py
@@ -57,7 +57,7 @@ class TestConfigFile(unittest.TestCase):
 
         # check `print` and `str` methods
         print(str(self))
-        print(len(str(self).split("\m")))
+        print(len(str(self).split("\m")))   # TODO: Should this be r"\m"?
 
     def test_getters(self):
         """Test the two ways to get entries."""
--- a/posydon/binary_evol/DT/step_detached.py
+++ b/posydon/binary_evol/DT/step_detached.py
@@ -24,7 +24,7 @@ from scipy.optimize import root
 from posydon.utils.data_download import PATH_TO_POSYDON_DATA
 from posydon.binary_evol.binarystar import BINARYPROPERTIES
 from posydon.binary_evol.singlestar import STARPROPERTIES
-from posydon.interpolation import GRIDInterpolator
+from posydon.interpolation.interpolation import GRIDInterpolator
 from posydon.interpolation.data_scaling import DataScaler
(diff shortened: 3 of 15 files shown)


posydon/tests/utils/test_configfile.py
```diff
@@ -57,7 +57,7 @@ def test_IO(self):
 
         # check `print` and `str` methods
         print(str(self))
-        print(len(str(self).split("\m")))
+        print(len(str(self).split("\m")))   # TODO: Should this be r"\m"?
```
mkruckow, line 60: I'd say this depends on whether the `str()` call on the object reruns a raw string or not. For me it's not that obvious what it returns, but I'd guess the standard wouldn't be a raw string, hence `"\m"` should be correct.

In fact `"\m"` and `r"\m"` return the same object, hence it won't make a difference. But e.g. `r"\t"` is equivalent to `"\\t"` instead of `"\t"`, thus here it would be important to know, whether the string object is a raw string or not when using the split function on it.

posydon/interpolation/__init__.py
```diff
@@ -1,2 +0,0 @@
-from .interpolation import GRIDInterpolator, fscale
```
mkruckow, line 1: I think, this requires another change in `posydon/binary_evol/DT/step_detached.py`:
the old: `from posydon.interpolation import GRIDInterpolator`
needs to read now: `from posydon.interpolation.interpolation import GRIDInterpolator`

At the moment all the following imports will cause an error:
`import posydon.binary_evol.DT.step_detached`
- `import posydon.binary_evol.DT.step_isolated`
  - `import posydon.binary_evol.DT.step_disrupted`
  - `import posydon.binary_evol.DT.step_initially_single`
  - `import posydon.binary_evol.DT.step_merged`
- `import posydon.tests.binary_evol.DT.test_step_detached`

ka-rocha (dismissed):
This looks good to me assuming @mkruckow's comment is resolved!

posydon/tests/utils/test_configfile.py
kkovlakas: I got linter warning on that - so maybe there are cases where this is important. I suggest to change it to r"\m" if this is what we want to be the string exactly. I was just not sure about the intention in this part of the code.

astroJeff: Let's resolve this with a separate PR. The TODO statement allows us to find it in the future.

mkruckow (approved):

posydon/interpolation/__init__.py
astroJeff: @kkovlakas has fixed this now.

astroJeff (approved):

PR #137, opened 2023-09-06
TF2 remake and support reverse MT flag in TF12

ssbvr:
This PR updates the termination flag 2 and the combination of TF1 and 2.

- [x] TF2 now displays the index of the donor star after the mass transfer case, hence supporting reverse mass transfer episodes, e.g., `case_A1/B1/B2`. 
- [x] TF2 is now computed also for unstable MT binaries
- [x] combined TF12 now displays reverse stable or unstable MT 
- [x] test the changes on the entire dataset
- [x] add support to classify and save mt_history in pop synth for each mesa step
- [x] add option to disentangle formation channel with respect to mt_history
- (optional: add an option to the parser to sort all revere MT binaries and compute LGRB rates)

Examples of the updated TF2 plots for HMS-HMS Zsun v2 grid.
q_0 95
q_0 99

Example of the updated TF12 plots for HMS-HMS Zsun v2 grid.
q_0 95
q_0 99

Example of a synthetic population data frame with formation channel column differentiating oRLO1/oRLO2 from oRLO1-contact/oRLO2-contact. The new information can be leveraged by the `get_dco_at_formation` method with the option `mt_history=True`. Short tutorial
```py
pop.get_dco_at_formation(S1_state='BH', S2_state='BH',   formation_channels=True, mt_history=True)
cols = ['metallicity','time','t_delay','S1_state','S2_state','S1_mass','S2_mass','orbital_period','eccentricity','channel']
pop.df_synthetic[cols]
```
Screenshot 2023-09-18 at 14 28 50

### DIFF
--- a/posydon/grids/termination_flags.py
+++ b/posydon/grids/termination_flags.py
@@ -99,20 +99,16 @@ def get_mass_transfer_flag(binary_history, history1, history2,
     Returns
     -------
     flag_system_evolution_history : string
-        Possible flags are: "contact during MS", "no_RLOF", a cumulative MT
-        flag + "_from_star1/2", or in case of missing binary history data,
-        "undetermined_flag_mass_transfer_with_no_binary_history".
+        Possible flags are: "None", "initial_RLOF", "contact_during_MS", 
+        "no_RLOF", a cumulative MT flag, e.g, "case_A1/B1/A2" where the
+        index indicates the donor star.
 
     """
-    if mesa_flag is not None:
-        if "overflow from L2" in mesa_flag:
-            if "at ZAMS" in mesa_flag:
-                return "initial_RLOF"
-            else:
-                return "L2_RLOF"
-
-    if binary_history is None:
-        return "undetermined_flag_mass_transfer_with_no_binary_history"
+    if mesa_flag in TF1_POOL_ERROR:
+        return "None"
+        
+    if mesa_flag in TF1_POOL_INITIAL_RLO:
+        return "initial_RLOF"
 
     rel1 = binary_history["rl_relative_overflow_1"]
     rel2 = binary_history["rl_relative_overflow_2"]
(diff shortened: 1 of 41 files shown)


mkruckow:
Let me start with some comments to the plots:

1. The symbol and color for the TF2 looks a bit random assigned. I'd suggest to bring there some more order in there, e.g. symbol=donor sequence & color=cases. To be a bit more precise: triangle edge right (only star 1 is donor), triangle edge left (only star 2 is donor), square (both get donor at some point, star 1 is last donor), circle (both get donor at some point, star 2 is last donor); case A only (color1), case A/B (color2), case A/B/C (color3), case A/B/C/BB (color4), case B (color5), case B/C (color6), ...
2. Could we use a different color for initial RLO vs no RLO for TF2 like we do for TF12?
3. In the q=0.99 TF2 plot is a teal triangle, which does not show up in the legend. What is that? Please add it.
4. In the q=0.99 TF12 plot is still an unknown. What is that? Please add it.

posydon/grids/termination_flags.py
```diff
@@ -99,20 +99,16 @@ def get_mass_transfer_flag(binary_history, history1, history2,
...
+        "no_RLOF", a cumulative MT flag, e.g, "case_A1/B1/A2" where the
+        index indicates the donor star.
 
     """
-    if mesa_flag is not None:
-        if "overflow from L2" in mesa_flag:
-            if "at ZAMS" in mesa_flag:
-                return "initial_RLOF"
-            else:
-                return "L2_RLOF"
```
mkruckow, line 112: This option `L2_RLOF` is not possible in the new version anymore. Is that correct, that it got removed?

posydon/utils/common_functions.py
```diff
@@ -1545,11 +1545,14 @@ def cumulative_mass_transfer_string(cumulative_integers):
...
-                result += "case"
+                result += "case_"
                 added_case_word = True
             else:
                 result += "/"
-            result += MT_CASE_TO_STR[integer]
+            if integer in MT_CASE_TO_STR:
+                result += MT_CASE_TO_STR[integer] + '1' # from star 1
+            else:
+                result += MT_CASE_TO_STR[integer-10] + '2' # from star 2
```
mkruckow, line 1555: I'd suggest to add a comment to the definition for MT_CASE_TO_STR, that only single digit values are allowed in the keys.
I'm wondering, whether it would be better instead of using a shift by 10 turn the number to be negative for star 2. In this way we'd avoid to limit us to single digits, and would only limit to positive numbers.

posydon/visualization/combine_TF.py
```diff
@@ -13,119 +13,134 @@
...
-                        'ignored_no_BH',
-                        'ignored_scrubbed']
-
-TF1_POOL_ERROR = ['logQ_limit',
-                  'logQ_min_limit',
-                  'min timestep',
-                  'min_timestep_limit',
-                  'cluster timelimit',
-                  'reach cluster timelimit',
-                  'no termination code',
```
mkruckow, line 62: I think we should keep something like this. To allow us to flag stuff which is not covered by the termination flags.

mkruckow:
A general question to `combine_TF.py`, is it good to define there a lot of POOLs? Wouldn't it be better to create a function, which checks for a POOL?

posydon/grids/termination_flags.py
astroJeff: This option is obsolete, so I think it is okay to remove.

mkruckow:
I think we should make sure, that there are no unknowns in the current girds with this PR.

mkruckow:
When using this PR on the CO-HeMS grid with the thermohaline mixing reruns, I got an error in the interpolator training. More details I send via slack to @ssbvr .

mkruckow (approved):
Beside the small comments, fine.

ZepeiX (approved):

PR #143, opened 2023-09-21
Signal handling

mkruckow:
All started with hard drive issues on the cluster in Geneva. I got asked, whether we can use a different drive during the run and copy/move our stuff we need after the simulation finished at the end to the final storage place. The data at the storage during the run will get deleted (or at least offered to the OS to be deleted) after the run finished automatically.

It turned out, that we already have a functionality for this in the `posydon-run-grid`, there we can set a working directory with the argument `--temporary-directory`.
But the current implementation would leave all the data only in the working directory in case the simulation fails for any reason. Hence, we need to capture those cases and still move our data before it gets deleted. Such things need to be done by reacting on the kill signal send.

- [x] Put file movement into its own function. [`posydon-run-grid`]
- [x] Add signal handler [`posydon-run-grid`]
  - react on SIGABRT (aborting signal)
  - react on SIGUSR1 (user signal)
  - react on SIGSEGV (segmantation fault)
- [x] Introduce new argument `work_dir` in slurm section of the ini file and add its value to be the `--temporary-directory` on run. [`posydon-setup-grid`]
- [x] Request slrum to send a SIGURS1 300 seconds before the wall time is reached. (I hope that should be enough time for a sure file movement) [`posydon-setup-grid`]
- [ ] may provide new ini files in the MESA-INLISTS submodule
- [ ] needs to be tested


mkruckow:
Got replaced by [PR#157](https://github.com/POSYDON-code/POSYDON/pull/157)

PR #159, opened 2023-10-17
Matthias additions to the docs

mkruckow:
- [x] PDFs should be embedded in html instead of loaded as images. The sphinx html builder does not support PDFs to be easily embedded and don't support them as images types. Hence, the resulting html code was wrongly trying to load a PDF as an image. The solution sphinx provides here is to not specify an extension and let the builder chose a file. Hence, I converted the logo PDF to a png and added it. The sphinx html builder will now chose the png, while e.g. the latexpdf would use the pdf.
- [x] Correcting some typos in introduction.
- [x] Improvements on getting-started section.


ssbvr (approved):
LGTM!

Issue #161, opened 2023-10-18
Dataframe concatenation - type discrepancy


maxbriel:
# Issue
Concatenation fails when a synthetic grid with few binaries is ran using MPI.

# Triaging
After running, the concatenation fails due to a mismatch in the types of the HDF5 table.
The exact mismatch comes from the additional `scalar_names` being added, i.e. `f_fb`, `spin_orbit_tilt`, and `natal_kick_array`.
When these remain undefined, they are set to np.NaN. However, this casts to a float, when written to a hdf5 file with pandas.
When a value is set, it becomes a float64. As a result there's a mismatch between the types. 

# Solution
Cast these variables to a float64 before storage.

maxbriel:
Solved with PR #167

Issue #155, opened 2023-10-13
Consistency in where to store natal kick properties


ssbvr:
PR#129 introduced a fix to the spin-orbit tilt evolution of the first collapsed compact object given the second core collapse kick. The PR introduced new variables for the kicks that are in part degenerate and stored outside of the `natal_kick_array` of both stars. I suggest removing the `natal_kick_array` and storing each property with a name.

Please also note that after PR#129 is merged:
- the following tutorial showing how to re-evolve binaries given a kick might no longer work `docs/_source/tutorials-examples/population-synthesis/debug_pop.ipynb`
- the rate calculation method of the SyntheticPopulation class computes effective spins given the `spin_orbit_tilt` variable. Remember to update this variable in favor of the new nomenclature that distinguishes if the tilt before and after the first and second core collapse.

astroJeff:
`natal_kick_array` is not located in the `restore` function. A possible solution: make sure the `natal_kick_array` is a `star` property

astroJeff:
@maxbriel

astroJeff:
This issue resolved with PR #731. Closing

PR #162, opened 2023-10-18
Concatenate fix

maxbriel:
This fixes issue #161, where the files would not concatenate after a run due to the types of several `scalar_name` columns between different `oneline_df`'s being different.

However, a better solution needs to be found for these scalar_name variables such that they're always cast to the same type as they're used in the code.
See issue #155 for mention of storage consistency of some of these variables as well.


mkruckow:
This PR is a code change, hence I redirected it to go into development not in documentation branch.
@maxbriel, in the future you should always branch out from development when you do a change. And make a PR back to development. Only branch out from another branch in case it builds up on that branch, which wasn't the case here.

PR #167, opened 2023-10-20
Concatenate fix for casting current scalar_names

maxbriel:
This fixes issue https://github.com/POSYDON-code/POSYDON/issues/161, where the files would not concatenate after a run due to the types of several scalar_name columns between different oneline_df's being different.

However, a better solution needs to be found for these scalar_name variables such that they're always cast to the same type as they're used in the code.
See issue https://github.com/POSYDON-code/POSYDON/issues/155 for mention of storage consistency of some of these variables as well.

### DIFF
--- a/posydon/binary_evol/singlestar.py
+++ b/posydon/binary_evol/singlestar.py
@@ -314,16 +314,13 @@ class SingleStar:
         for name in scalar_names:
             if hasattr(self, name):
                 if name == 'natal_kick_array':
-                    natal_kick_array = np.array(getattr(self, name), dtype=np.float64)
+                    natal_kick_array = getattr(self, name)
                     for i in range(4):
                         col_name = prefix+name+'_{}'.format(int(i))
                         oneline_df[col_name] = [natal_kick_array[i]]
-                elif (name == 'spin_orbit_tilt') | (name == 'f_fb'):
-                    oneline_df[prefix+name] = [np.float64(getattr(self, name))]
                 else:
                     oneline_df[prefix+name] = [getattr(self, name)]
         return oneline_df
-
 
     def __repr__(self):
         """Return the object representation when print is called."""
--- a/posydon/binary_evol/singlestar.py
+++ b/posydon/binary_evol/singlestar.py
@@ -314,13 +314,16 @@ class SingleStar:
         for name in scalar_names:
             if hasattr(self, name):
                 if name == 'natal_kick_array':
-                    natal_kick_array = getattr(self, name)
+                    natal_kick_array = np.array(getattr(self, name), dtype=np.float64)
                     for i in range(4):
                         col_name = prefix+name+'_{}'.format(int(i))
                         oneline_df[col_name] = [natal_kick_array[i]]
+                elif (name == 'spin_orbit_tilt') | (name == 'f_fb'):
+                    oneline_df[prefix+name] = [np.float64(getattr(self, name))]
                 else:
                     oneline_df[prefix+name] = [getattr(self, name)]
         return oneline_df
+
 
     def __repr__(self):
         """Return the object representation when print is called."""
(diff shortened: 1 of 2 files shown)


posydon/binary_evol/singlestar.py
```diff
@@ -314,14 +314,17 @@ def to_oneline_df(self, history=True, prefix='', **kwargs):
         for name in scalar_names:
             if hasattr(self, name):
                 if name == 'natal_kick_array':
-                    natal_kick_array = getattr(self, name)
+                    natal_kick_array = np.array(getattr(self, name), dtype=np.float64)
                     for i in range(4):
```
mkruckow: As you are just doing changes here, I'd suggest to either change the range to go over the size of the array or to check, that the array has the correct size and otherwise raise an error.

posydon/binary_evol/singlestar.py
```diff
@@ -314,14 +314,17 @@ def to_oneline_df(self, history=True, prefix='', **kwargs):
...
                         col_name = prefix+name+'_{}'.format(int(i))
                         oneline_df[col_name] = [natal_kick_array[i]]
+                elif (name == 'spin_orbit_tilt') | (name == 'f_fb'):
+                    oneline_df[prefix+name] = [np.float64(getattr(self, name))]
```
mkruckow: Did you checked, that there are only 'spin_orbit_tilt' and 'f_fb', which need to get float64? I'd expect there should be more floats.

posydon/binary_evol/singlestar.py
```diff
@@ -314,14 +314,17 @@ def to_oneline_df(self, history=True, prefix='', **kwargs):
         for name in scalar_names:
             if hasattr(self, name):
                 if name == 'natal_kick_array':
-                    natal_kick_array = getattr(self, name)
+                    natal_kick_array = np.array(getattr(self, name), dtype=np.float64)
                     for i in range(4):
                         col_name = prefix+name+'_{}'.format(int(i))
                         oneline_df[col_name] = [natal_kick_array[i]]
+                elif (name == 'spin_orbit_tilt') | (name == 'f_fb'):
```
mkruckow: For consistent code style, please use `or` instead of `|`.

posydon/binary_evol/singlestar.py
```diff
@@ -314,14 +314,17 @@ def to_oneline_df(self, history=True, prefix='', **kwargs):
...
+                    oneline_df[prefix+name] = [np.float64(getattr(self, name))]
                 else:
                     oneline_df[prefix+name] = [getattr(self, name)]
         return oneline_df
 
+
```
mkruckow: When you make changes, remove the unnecessarily introduced new line here.

mkruckow:
A more general question to @ka-rocha: didn't you made a type check for stuff being written to h5 files somewhere else, see [PR#88](https://github.com/POSYDON-code/POSYDON/pull/88)? Why is this here needed and not done with the other stuff?

ka-rocha:
@mkruckow Yes I have tried enforcing data types at the end of the `to_df` with methods in `popsyn/io.py` where I have defined some default data types for binary and single star properties but not for the `scalar_names` variables! If this kind of solution is still ok, then I would suggest solving this `scalar_names` issue in a similar way so things are consistent.

maxbriel:
Thank you both for the feedback. Based on this, I've implemented a solution where `scalar_names` is cast to the right datatype at the same location the binary and single star properties are cast.

I have created a default `SCALAR_NAMES_DTYPES` dict in `popsyn/io.py` for this.

Currently, each element in `natal_kick_array` has to to be defined separately, ie.`natal_kick_array_0`, `natal_kick_array_1`, etc.
Since these properties might be separated as a fix for Issue #155, this is hopefully a sufficient solution for now.

mkruckow (approved):
Looks fine to me.

ka-rocha (approved):
This looks great!

PR #170, opened 2023-10-24
add 'stable_reverse_MT' and interpolation_class for CO

ZepeiX:
create a new MT subclass 'stable_reverse_MT' and attribute interpolation_class to CO that allows interpolating CO quantities based on 'WD', 'NS', 'BH', 'BH_reverse_MT'


mkruckow:
```
    if 'HMS-HMS' in grid_path and not '_RLO' in grid_path:
        interp_method = [method, method, method, method]
        interp_classes = ["no_MT", "stable_MT", "unstable_MT", "stable_reverse_MT"]
    elif 'HMS-HMS' in grid_path and '_RLO' in grid_path:
        # This grid currently does not exist. But in case we want to include it
        # in the future, we have added the functionality.
        interp_method = [method, method, method]
        interp_classes = ["stable_MT", "unstable_MT", "stable_reverse_MT"]
    elif '_RLO' in grid_path:
        interp_method = [method, method]
        interp_classes = ["stable_MT", "unstable_MT"]
    else:
        interp_method = [method, method, method]
        interp_classes = ["no_MT", "stable_MT", "unstable_MT"]
```
I'd still suggest to avoid the long way of getting all cases, instead do it in a logical and flexible way:
```
    interp_method = [method, method]
    interp_classes = ["stable_MT", "unstable_MT"]
    if '_RLO' not in grid_path: # add "no MT" for girds not forced to start at RLO
        interp_method += [method]
        interp_classes += ["no_MT"]
    if 'CO-' not in grid_path: # add "reverse MT" for grids without one star being a compact object
        interp_method += [method]
        interp_classes += ["stable_reverse_MT"]
```

mkruckow:
Could you please tell us, why we need the new field "interpolation_class" to be stored in the CC quantities?
The only use case of it is, when you train the interpolators, hence, you could simply use there the information from the CO_type and the TF2 to create the interpolation class only there. Probably best: write a function which returns you the interpolation class given CO_type and TF2.

posydon/grids/post_processing.py
```diff
@@ -293,7 +294,7 @@ def post_process_grid(grid, index=None, star_2_CO=True, MODELS=MODELS,
                                 if not isinstance(getattr(star_copy, quantity), str):
                                     flush = True
                                     warnings.warn(f'{MODEL_NAME} {mechanism} {quantity} is not a string!')
-                            else:
+                            elif quantity != 'interpolation_class':
```
mkruckow: Why don't we check the interpolation_class to be a string?

posydon/grids/termination_flags.py
```diff
@@ -245,7 +245,10 @@ def infer_interpolation_class(tf1, tf2):
     if tf2 in TF2_POOL_NO_RLO:
         return "no_MT"
     if tf1 in TF1_POOL_STABLE:
-        return "stable_MT"
+        if '1' in tf2 and '2' in tf2:
```
mkruckow: Add to check for 'case' in tf2 to be sure talking about stable MT.

tassos25:
This PR is submitted by @ZepeiX but it is really @prinse1545 's code. So I am not sure who is supposed to answer @mkruckow 's questions.

@mkruckow apart from the comments above, have you tested the PR and found that the output of the trained IF interpolation objects do not seem to have any issue?

This PR needs to be checked carefully and accepted ASAP! It is pending since early summer and it is blocking several other steps. @kkovlakas if you have some time, I would appreciate if you can look at it as well.

astroJeff:
Discussed briefly at dev meeting 14/12. We need to carefully review this when Zepei and Philipp are online so that we can determine exactly what needs to happen to finalize this.

astroJeff:
Update (Feb 1): We briefly discussed this with @ZepeiX online. He needs more time to review - it looks like the code works, but specific comments from @mkruckow above need to be addressed. @kkovlakas to review.

ZepeiX:
The major point from @mkruckow is that since what we want is to do class-wise interpolation for compact objects including a new 'stable_reverse_MT' class, we can just do it in the pipeline of training interpolator instead of saving the information in CC quantities.
I kind of agree that this could be better. In post processing, we don't save another 'interpolation class' separating BHs between normal and reverse MT. Instead, we get the MT case before training the interpolator.
To achieve this, I expect an overhaul regarding this PR and tests afterward.

astroJeff:
Update (Feb 8): Following @mkruckow's suggestion, @ZepeiX will adjust this code so the pipeline is what identifies systems as having gone through reverse mass transfer, rather than have a separate compact object class. This will take some time to update this PR.

astroJeff:
Update (Feb 15): Test is running right now. A positive test would indicate higher spins? @ZepeiX will get back to us with the results (by tomorrow?)

astroJeff:
Also, @ZepeiX please test this PR on the CO-HMS_RLO and CO-HeMS grids

astroJeff (approved):

elizabethteng (approved):

mkruckow (approved):
It is not in best shape and may cause issues in the future, but it works for now.

Issue #174, opened 2023-10-26
Duplicate entries in `SyntheticPopulation.df_oneline` + missing history in SyntheticPopulation.df`


maxbriel:
**Current Behaviour**

- Several binaries appear twice in the `SyntheticPopulation.df_oneline` DataFrame after parsing the population files.
- Parts of the history of some binaries are missing in `SyntheticPopulation.df`.

**Expected Behaviour**

- Each binary has a single entry in `SyntheticPopulation.df_oneline`.
- Each binary has its full history in `SyntheticPopulation.df.

**Triaging**
`SyntheticPopulation.df` only contains a single history of each binary. The issue originates from the chunking of the binaries in the history file. If a binary in the history file is split between two chunks, it will occur twice in `sel_met`, which is used to select the `oneline` elements to include. 

Similarly, the chunking causes history of a single binary to be split between different chunks and if a system is only tagged one of the chunks, it previous or continued history is not included. 

**Solution**
Make sure a binary is not split over two chunks by including the last binary of a chunk into the next chunk and performing a selected of `df_oneline` based on the selected `df` per metallicity.

PR #175, opened 2023-10-26
Synthetic Population: MPI removal + DataFrame fix

maxbriel:
Included in this PR:

- Removal of MPI for the Synthetic Population. The user submits a SLURM job array with the walltime and memory per job. Each `BinaryPopulation` is split between the sub-processes and a sub-process will move on to the next `BinaryPopulation` once finished with the previous metallicity. A separate job can be submitted that concatenates the files per metallicity, which depends on the job-array.
- Tutorial adapted to show the changes in how to submit a job array. 
- `SyntheticPopulation.binarypopulations` creation moved to separate class function and only initialised when population is evolved.
- Random Number generation using the JOB_ID if no entropy is given. Sub-processes will have the same entropy, but will use different sequence-spawns from the same entropy, which increases reproducibility if the user does not specify an entropy for their population. Furthermore, multi-metallicity runs use the metallicity index to create an entropy for each metallicity based on the JOB_ID. However, I have not tested if this leads to unique SeedSequences.
- Correct treatment of chunks for `SyntheticPopulation.parse()`. This solves both components of Issue #174. 
- Removal of `path_to_data` from initialisation of `SyntheticPopulation` class and shifted to parser as an additional input variable since it is only used here.
- Addition of `step_name` to the parser input.
 - fixes a few small typos in tutorials.


mkruckow:
>     * Removal of `path_to_data` from initialisation of `SyntheticPopulation` class and shifted to parser as an additional input variable since it is only used here.

For me it looks like, that path_to_data should be part of the metadata of the SyntheticPopulation to always know from which population it originated from. Hence, it could be set on initializing and should be saved to and loaded from the hdf5 file. Yes, we can have the additional option to specify it on parsing as well.

posydon/popsyn/binarypopulation.py
```diff
@@ -76,6 +76,8 @@
     'step_HMS_HMS', 'step_CO_HeMS', 'step_CO_HMS_RLO', 'step_CO_HeMS_RLO', 'step_detached','step_isolated','step_disrupted','step_initially_single', 'step_merged'
 ]
 
+POSYDON_METALLICITIES = [2e+00,1e+00,4.5e-01,2e-01,1e-01,1e-02,1e-03,1e-04]
```
mkruckow: Those are the metallicities for v2. In future versions, we might get more. Hence, I'm not sure, whether we should hard code this here. Wouldn't it be better the infer a list from the used data set?

posydon/popsyn/binarypopulation.py
```diff
@@ -105,10 +107,20 @@ def __init__(self, **kwargs):
...
+            self.rank = self.kwargs.pop('RANK', None)
+            self.size = self.kwargs.pop('size', None)
+            
+            # Make sure each of the MPI processes has the same entropy
+            # But unique per metallicity
+            if entropy is None:
+                met_shift = POSYDON_METALLICITIES.index(
+                                    self.kwargs.get('metallicity')) +1
+                
+                seq = np.random.SeedSequence(entropy=self.JOB_ID * met_shift)
```
mkruckow: Usually I'd suggest to use a sum instead of a product to avoid getting too easy above the largest possible number.

posydon/popsyn/binarypopulation.py
```diff
@@ -235,13 +246,15 @@ def _safe_evolve(self, **kwargs):
...
                 os.makedirs(temp_directory)
         else:
-            if self.rank == 0:
-                if not os.path.exists(temp_directory):
-                    os.makedirs(temp_directory)
+            # Create a directory for MPI runs
+            try: 
+                os.makedirs(temp_directory)
+            except FileExistsError:
+                pass
```
mkruckow: Why do we use a try clause here instead of checking, whether the directory already exists?

posydon/popsyn/io.py
```diff
@@ -487,13 +487,18 @@ def binarypop_kwargs_from_ini(path, verbose=False):
...
-                from mpi4py import MPI
-                pop_kwargs['comm'] = MPI.COMM_WORLD
+            # Check if we are running as a job array
+            JOB_ID = os.getenv('SLURM_ARRAY_JOB_ID')
+            if JOB_ID is not None:
+                pop_kwargs['JOB_ID'] = np.int64(os.environ['SLURM_ARRAY_JOB_ID'])
+                # account for job array not starting at 0
+                min_rank = np.int64(os.environ['SLURM_ARRAY_TASK_MIN'])
+                pop_kwargs['RANK'] = np.int64(os.environ['SLURM_ARRAY_TASK_ID'])-min_rank
+                pop_kwargs['size'] = np.int64(os.environ['SLURM_ARRAY_TASK_COUNT'])
```
mkruckow: Are we sure, that 'SLURM_ARRAY_TASK_MIN', 'SLURM_ARRAY_TASK_ID' and 'SLURM_ARRAY_TASK_COUNT' always exist if 'SLURM_ARRAY_JOB_ID' does?

posydon/popsyn/synthetic_population.py
```diff
@@ -62,38 +68,61 @@ def __init__(self, path_to_ini, path_to_data=None, verbose=False, MODEL={}):
...
     def evolve(self):
         """Evolve population(s) at given Z(s)."""
+        if self.binary_populations is None:
+            self.create_binary_populations()
         while self.binary_populations:
             pop =  self.binary_populations.pop()
             print( f'Z={pop.kwargs["metallicity"]:.2e} Z_sun' )
             pop.evolve()
-            met_prefix = f'{pop.kwargs["metallicity"]:.2e}_Zsun_'
-            pop.save( met_prefix + 'population.h5' )
```
mkruckow, line 95: Why can we skip to save the outcome of the evolution before deleting it?

posydon/popsyn/synthetic_population.py
```diff
@@ -28,18 +30,22 @@
...
         Parameters
         ----------
 
         path : str
             Path to the inifile to parse. You can supply a list in the
             metallicity parameter to evolve more than one population.
-
         """
+        self.synthetic_pop_params = None
+        self.metallicity = None
```
mkruckow: Shouldn't we better call this metallicities, because it is a list of them?
```
            if not isinstance( self.metallicity, list):
                self.metallicity = [self.metallicity]
```

posydon/popsyn/synthetic_population.py
```diff
@@ -62,38 +68,61 @@ def __init__(self, path_to_ini, path_to_data=None, verbose=False, MODEL={}):
...
+
+        for met, path_to_batch in zip(self.metallicity, path_to_batches):
+            met_prefix = self.create_met_prefix(met)
+            tmp_files = [os.path.join(path_to_batch, f)     \
+                         for f in os.listdir(path_to_batch) \
+                            if os.path.isfile(os.path.join(path_to_batch, f))]
+            
+            BinaryPopulation(**self.get_ini_kw()).combine_saved_files(met_prefix+ 'population.h5', tmp_files)
+            print(f'Population at Z={met:.2e} Z_sun successfully merged!')
+            os.rmdir(path_to_batch)
```
mkruckow: This might fail if the directory is not empty. The code only removed the files, which where successfully combined into the new file. The could be other files or maybe even some subdirectories left in there.

posydon/popsyn/synthetic_population.py
```diff
@@ -197,36 +226,56 @@ def parse(self, S1_state=None, S2_state=None, binary_state=None,
             Read the POSYDON binary population in chuncks to prevent OFM error.
 
         """
+        
+        # if the user provided a single string instead of a list of strings
+        if type(path_to_data) is str and ('.h5' in path_to_data):
+            path_to_data = [path_to_data]
+        
+        # catch the case where the user did not provide a path to data     
+        if not (isinstance(path_to_data, list) and '.h5' in path_to_data[0]):
```
mkruckow: Why do you only check for the first one? The second one could be a non h5 file. Thus I'd split the check on list and a check in a loop through the list.
Additionally, you should use `os.path.splitext(path_to_data[0])[-1]` to check the file extension.

posydon/popsyn/synthetic_population.py
```diff
@@ -197,36 +226,56 @@ def parse(self, S1_state=None, S2_state=None, binary_state=None,
...
-                logic = self.apply_logic(df, S1_state=S1_state,
-                                    S2_state=S2_state,
-                                    binary_state=binary_state,
-                                    binary_event=binary_event,
-                                    invert_S1S2=invert_S1S2)
-
+                
+                if last_binary_df is not None:
+                    df = pd.concat([last_binary_df, df])
+                    last_binary_df = None
```
mkruckow: Why do we need this? We set it here to None and in the next code line it gets a new assignment nevertheless.

posydon/popsyn/synthetic_population.py
```diff
@@ -239,22 +288,41 @@ def parse(self, S1_state=None, S2_state=None, binary_state=None,
...
+                                    step_name    = step_name,
+                                    invert_S1S2  = invert_S1S2)
+                
+                # The last binary is selected
+                if any(logic) == True:
+                    df_tmp = last_binary_df.loc[logic]
+                    df_sel_met = pd.concat([df_sel_met, df_tmp])
+
+            # get unique indicies
+            sel_met = df_sel_met.index.unique()
```
mkruckow: Could you please tell, why you use `unique` instead of `drop_duplicates` here, cf. `sel = df.loc[logic].index.drop_duplicates()` being used before?

posydon/popsyn/synthetic_population.py
```diff
@@ -239,22 +288,41 @@ def parse(self, S1_state=None, S2_state=None, binary_state=None,
...
+            
             if k > 0 and df_sel.shape[0] > 0:
                 shift_index = max(np.unique(df_sel.index)) + 1
```
mkruckow: Wouldn't it be better to always shift by the maximum index of all the data instead of only using the selection?
I have in mind to make two sections from the same population, than the same systems could get a different index at the moment.

posydon/popsyn/binarypopulation.py
```diff
@@ -105,10 +107,20 @@ def __init__(self, **kwargs):
         entropy = self.kwargs.get('entropy', None)
         seq = np.random.SeedSequence(entropy=entropy)
```
ka-rocha: I am realizing now that removal of MPI in this fashion will also remove functionality for users to parallelize populations locally without a compute cluster. Unless we want to implement multiprocessing as well, MPI was an easy way for the code to be parallelized locally & I remember this being an important part of code marketability, even if in practice we would use slurm 90% of the time.

Is it not possible to keep some of the old MPI communicator code and when running with job arrays and setting `comm=None`? Or is it the case that we are removing MPI permanently from the code base?

maxbriel:
> > ```
> > * Removal of `path_to_data` from initialisation of `SyntheticPopulation` class and shifted to parser as an additional input variable since it is only used here.
> > ```
> 
> For me it looks like, that path_to_data should be part of the metadata of the SyntheticPopulation to always know from which population it originated from. Hence, it could be set on initializing and should be saved to and loaded from the hdf5 file. Yes, we can have the additional option to specify it on parsing as well.
 
The class does not load with an origin population except for the ini file, since it also creates the populations. So I don't think it's possible to have it be declared at initialisation. The issue I can see arising is a user doing a population synthesis run and altering the ini file afterwards. We could store the ini filename inside the BinaryPopulations?

The `path_to_data` is only used for parsing the data from the raw BinaryPopulations. When a parsed population gets loaded, it gets loaded using `load_pop` or a synthetic population using `load_synthetic_pop`. In both those cases, `path_to_data` is not used. And the path to data never gets written to any of the parsed population files.

I agree that it's useful to know what population the parsed population originates from.
I can add the `path_to_data` as metadata to the parsed hdf5 file.
Would it be worth storing the parse query too?
However, other sub-populations do not currently store any metadata either.
Do we want these to also include these data there?

posydon/popsyn/binarypopulation.py
mkruckow: For running it on a single computer, what do we need from MPI, what we can't do with splitting the process and let things run in child processes with one parent keep track of all?
The issue on the cluster already appears when loading MPI without initializing an using it.

posydon/popsyn/binarypopulation.py
maxbriel: Because the directory check can occur in multiple processes before one creates the directory. As such, it passes the check, but fails at the directory creation.

posydon/popsyn/synthetic_population.py
maxbriel: Inside `BinaryPopulation.evolve()` the evolution is already saved for each individual process.

This code merged the individual subprocess files into one, which is now taken care of by `merge_parallel_runs()`

posydon/popsyn/binarypopulation.py
maxbriel: Would this provide enough uniqueness? ie. wouldn't we get similar seeds if the JOB_IDs of two runs are close together?

posydon/popsyn/synthetic_population.py
maxbriel: `unique` provides a numpy array, while `drop_duplicates` provides a pandas Series. It shouldn't matter much in practice, except maybe some memory usage.

mkruckow:
Yes, we should think about doing there some better bookkeeping. I'd say we should try to keep all information we may need. Hence, I suggest to make the decision, what we need based on the following thought: get a population from a file and ask yourself, what you can do with it, just with the information you get when loading it, because this is the situation, when you share the data or you'll look at it again years later.

posydon/popsyn/binarypopulation.py
maxbriel: This is true. This should totally be fine. I can add those components back in. Since this version doesn't use the `use_MPI` flag, it should be possible to have both exist side-by-side without requiring this part of the code to initialise MPI.

posydon/popsyn/binarypopulation.py
mkruckow: This depends on the random number generator, which is used and how the entropy is exactly defined there. For most random number generators neighboring numbers are not that likely to be part of the same sequence. For an ideal random number generator, any connection (sum, product, any more complex functionality) between the seed should have the same likelihood to get two of them being part of the same sequence. In reality, is depends on the generator, which kind of connections should be avoided.

posydon/popsyn/binarypopulation.py
mkruckow: Yes, it is useful to capture issues on creation. Why do we skip the check to avoid unnecessary exceptions?

posydon/popsyn/synthetic_population.py
mkruckow: Hm, why was it than saved here before?
Additionally, I'd suggest to rename the `evolve` function than to e.g. `evolve_and_save` to make clear, that this has to stay in this function.

posydon/popsyn/synthetic_population.py
mkruckow: If we don't care here, which to use, we should use the same for consistency.

posydon/popsyn/binarypopulation.py
maxbriel: I do not know if this really matters, since the randomness within a SyntheticPopulation is still fine.
But I was specifically referring to two different population synthesis runs.

For example, if JOB_1 has `JOB_ID =1` with 2 metallicities, then the entropy will be 1 and 2. 
If JOB_2 has 'JOB_ID=2', then the entropy will be 2 and 3. 
There will be overlap with the same entropy between those runs.

mkruckow: You shouldn't get an overlap for the same metallicity. And overlaps for different metallicities shouldn't matter as long as you don't join them, which you shouldn't as long as you have fixed and dedicated metallicity values.
If you care about overlaps, I'd suggest you to use the JOB_ID of each individual job and not the one of the job array. Otherwise you can use the start time.
But all in all, it shouldn't matter too much how to shift it exactly.

posydon/popsyn/binarypopulation.py
maxbriel: > For running it on a single computer, what do we need from MPI, what we can't do with splitting the process and let things run in child processes with one parent keep track of all?

Could you provide a bit more information on this?
To me it seems that if `mpi4py` is not imported, except if used locally, it shouldn't interfere with SLURM's MPI.

mkruckow: Yes, that's the point we can't import it. But usually in python you try to import stuff at the beginning to get script checks enabled for the imported parts. If we import stuff only under a condition during running the script, the syntax check for the imported script does happen at this moment and not at the beginning of running the script. It should be fine for an external module, but we need to be careful with such things.

posydon/popsyn/synthetic_population.py
maxbriel: There shouldn't be any files left if correctly finished, but I've created a check to make sure that the folder is empty before it tries to remove it. 
```suggestion
            if len(os.listdir(path_to_batch)) == 0:
```

posydon/popsyn/synthetic_population.py
maxbriel: This was code already present in the previous version. I just moved it to this function. The user can give a single string, which gets transformed into a list. 
I've added to following checks:
1. If not list, then a `ValueError`` is raised.
2. If list, then check extension of each file for `.h5`.

posydon/popsyn/synthetic_population.py
maxbriel: Since `pd.concat` at the moment is fine with a None input it works.

posydon/popsyn/binarypopulation.py
maxbriel: I have removed it and passed the `metallicities` as an input from the `SyntheticPopulation` to the `BinaryPopulation`. If that is not used, then the `metallicity` is used for backwards compatibility.

posydon/popsyn/io.py
maxbriel: Yes. These are always spawned at the same time.

posydon/popsyn/synthetic_population.py
maxbriel: This `save` waited for all sub-processes and then called `combine_saved_files` to move the data from individual runs into a single population file. That part is now done by `merge_parallel_runs()`. 

`evolve` does save binaries to disk, but only during the run if it exceeds the ram limit, dump_rate, or a parallel run.

posydon/popsyn/binarypopulation.py
```diff
@@ -100,25 +100,47 @@ def __init__(self, **kwargs):
...
-        if self.JOB_ID is not None:
+        
+        if self.comm is not None and self.JOB_ID is not None:
+            raise ValueError('MPI and Job array runs are not compatible.')
+        # To guarantee reproducibility, we need to set the seed sequence
+        # to be the same across all processes.
+        elif self.comm is not None and self.entropy is None:
+            raise ValueError('A local MPI run requires an entropy value to be set.')
+        # local MPI run
+        elif self.comm is not None:
```
mkruckow, line 119: Minor, just for readability of the code I suggest:
```
        # To guarantee reproducibility, we need to set the seed sequence
        # to be the same across all processes.
        elif self.comm is not None and self.entropy is None:
            raise ValueError('A local MPI run requires an entropy value to be set.')
        # local MPI run
        elif self.comm is not None:
            ...
```
should read:
```
        # local MPI run
        elif self.comm is not None:
            # To guarantee reproducibility, we need to set the seed sequence
            # to be the same across all processes.
            if self.entropy is None:
                raise ValueError('A local MPI run requires an entropy value to be set.')
            ...
```
because it is not another option of the outer decision tree. Usually, this should be a bit faster as well but may not in python.

posydon/popsyn/synthetic_population.py
```diff
@@ -110,18 +110,19 @@ def merge_parallel_runs(self, path_to_batches):
...
             met_prefix = self.create_met_prefix(met)
             tmp_files = [os.path.join(path_to_batch, f)     \
                          for f in os.listdir(path_to_batch) \
                             if os.path.isfile(os.path.join(path_to_batch, f))]
             
             BinaryPopulation(**self.get_ini_kw()).combine_saved_files(met_prefix+ 'population.h5', tmp_files)
             print(f'Population at Z={met:.2e} Z_sun successfully merged!')
-            os.rmdir(path_to_batch)
+            if len(os.listdir(path_to_batch)) == 0:
+                os.rmdir(path_to_batch)
```
mkruckow, line 126: Minor, I'd suggest to get here an output on verbose if there is nothing to remove, e.g.
```
elif self.verbose:
    print(f'{path_to_batch} is not empty')
```

posydon/popsyn/synthetic_population.py
```diff
@@ -256,9 +260,8 @@ def parse(self, path_to_data, S1_state=None, S2_state=None, binary_state=None,
             
             for i, df in enumerate(pd.read_hdf(file,  key='history', chunksize=chunksize)):
                 
-                if last_binary_df is not None:
-                    df = pd.concat([last_binary_df, df])
-                    last_binary_df = None
+                df = pd.concat([last_binary_df, df])
+                last_binary_df = None
```
mkruckow: Now it got even more strange after removing the if clause, why to have:
```
                last_binary_df = None

                last_binary_df = df.loc[[df.index[-1]]]
```
Just remove the first of those lines.

posydon/popsyn/binarypopulation.py
maxbriel: I've put back the check to import MPI. It only imports if `use_MPI` is set to True. Otherwise it is not imported. Moreover, additional checks are put in place to make sure that `use_MPI` and job arrays are not used at the same time.

maxbriel:
The suggested changes have been implemented. 
There were two main topic that we discussed additionally to this. 

1. I've re-implemented the MPI for local runs.
@ka-rocha can you please make sure it also works on your machine? Could you specifically check if I've implemented the entropy for these runs correctly. I couldn't find a way to give all sub-processes with MPI the same starting entropy, so I've required the user to give an entropy if they're using MPI.

2. Bookkeeping of the `ini` file and what populations the `parsed_populations`, `synthetic_population` and `observed_populations` come from.
I have not implemented this. I think this is outside the scope of this PR and I've raised an issue for it (#179).

mkruckow (dismissed):
Looks fine to me.

posydon/popsyn/synthetic_population.py
```diff
@@ -317,7 +320,7 @@ def parse(self, path_to_data, S1_state=None, S2_state=None, binary_state=None,
             sel_met = df_sel_met.index.drop_duplicates()
             
             if k > 0 and df_sel.shape[0] > 0:
-                shift_index = max(np.unique(df.index)) + 1 
+                shift_index += max(np.unique(df.index)) + 1
```
mkruckow: When introducing the `+` here, which makes sense, you should cover the case of `k > 0 and df_sel.shape[0] <= 0`, where we keep `shift_index` as is and not reset it to zero. I'd suggest this to become a non written else:
```
if k==0:
    shift_index = 0
elif df_sel.shape[0] > 0:
    shift_index += max(np.unique(df.index)) + 1
```
Usually, it won't matter, but we should may think, whether `k==0` or `k<=0` fits better our code style.

maxbriel: Would it actually be better to always shift the index; independent if any binaries are selected?

That way we can define `shift_index` outside the metallicity loop and always increase for every metallicity in the Synthetic population.

mkruckow: It depends on your needs:

- We need the shift index only to get unique IDs: no need to always shift
- We'd like to get reproducible and unique IDs: we should always shift

So, yes we can initialize the shift before the metallicity loop and increase it at the end of each metallicity.

maxbriel: I have chosen to always shift for each metallicity. This is the best for reproducibility and debugging BinaryPopulations.

mkruckow (dismissed):

tassos25:
@ka-rocha, can you please have a look at it? This PR seems ready to be accepted as well.

posydon/popsyn/binarypopulation.py
```diff
@@ -375,20 +407,8 @@ def save(self, save_path, **kwargs):
...
-                tmp_files = [os.path.join(
-                    self.kwargs["temp_directory"], f"evolution.combined.{i}")
-                             for i in range(self.size)]
-
-                self.combine_saved_files(absolute_filepath, tmp_files, mode=mode, **kwargs)
-
-            else:
-                return
+                
+            self.combine_save_files(absolute_filepath, temp_directory, file_name, **kwargs)
```
ka-rocha: @maxbriel I ran into an error here and I think this function call was supposed to be `combine_saved_files`, although there are one too many arguments here for that method.

At some point we had code running in the notebooks with explicit calls to this save method when we were running individual binary populations at a time instead of straight to the synthetic population like:
```python
pop = BinaryPopulation(...)
pop.evolve(tqdm=True)
pop.save('filename.h5')
```
But after looking through the `_safe_evolve` it looks like saving is automatically done regardless of whether you call the save method, and instead this method is a combination step now? If that is the cases maybe it should be renamed.

ka-rocha: Actually I think I confused myself, it looks like in the non-parallel case it does actually save, but with MPI it will combine? In that case, I think it should be `combine_saved_files(absolute_filepath, file_name, **kwargs)` but please chime in if this is correct.

maxbriel: Yeah that's definitely a wrong function reference and it should call `combined_saved_files`.
Although I think the function call should be  `self.combine_saved_files(absolute_filepath, tmp_files, mode=mode, **kwargs)`, since it will combine files from multiple-parallel runs.

with
       `tmp_files = [os.path.join(self.kwargs["temp_directory"], f"evolution.combined.{i}") for i in range(self.size)]

I haven't changed the structure of the `save`/`evolve` functions with respect to the how it save/stores data, but evolve is definitely saving the binaries while evolving them if the `ini` parameters are set so.

I've only removed the step in `SyntheticPopulation.evolve()` that called `BinaryPopulation.save`, because that it meant a sub-process had to wait for all processes to finish. Now it can move to the next metallicity and merging of files from reach sub-process happens after the runs have finished.

maxbriel: @ka-rocha  I've altered this part of the code to read the temporary file directory and merge the files in that folder together.

`evolve` runs and saves the population if the right parameters are set (see `_safe_evolve`).
In the case of a single run, the single `evolution.combined` file is taken and written to a file with a given filename what you want it to be. 
In the case of an MPI/job array run, there will be multiple `evolution.combined_{i}` files, save should then take these and combine them into a file with your given file name :)

ka-rocha: Alright this looks good!

ka-rocha (approved):

astroJeff:
Status update (Dec 7): @ka-rocha to test job array functionality on Quest.

ka-rocha (approved):
I have testing this on quest using Job Arrays and things are working well! I did run into more issues with combining the saved files as described in Issue #208  but without the occasional failed binary, things combine correctly.

mkruckow (approved):

PR #177, opened 2023-11-01
pipeline updates for reruns

mkruckow:
- add logic to identify runs
  - [x] **HeMB_MTLp_mesh**
  - [x] **more_mesh**
  - [x] **conv_bdy_weight**
- replace MESA-INLISTS commits
  - [x] **HeMB_MTLp_mesh** after [PR#27](https://github.com/POSYDON-code/POSYDON-MESA-INLISTS/pull/27) and [PR#28](https://github.com/POSYDON-code/POSYDON-MESA-INLISTS/pull/28) got merged
  - [x] **more_mesh** after a PR on [MESA-INLISTS](https://github.com/POSYDON-code/POSYDON-MESA-INLISTS) is done
  - [x] **conv_bdy_weight** use same as for **more_mesh**

### DIFF
--- a/bin/run-pipeline
+++ b/bin/run-pipeline
@@ -452,7 +452,7 @@ def copy_ini_file(grid_path, rerun_type, destination, cluster):
     elif rerun_type == 'thermohaline_mixing':
         replace_text = "development-22c1bb9e730343558c3e70984a99b3fc1f3c346e"
     elif rerun_type == 'HeMB_MTLp_mesh':
-        replace_text = "development-72d61123df1a669cebbf5811d6f1398221b1f081" #TODO:update
+        replace_text = "development-435d16c9158e4608530f21b8169ce6f31160e23e"
     elif rerun_type == 'more_mesh':
         replace_text = "development-72d61123df1a669cebbf5811d6f1398221b1f081" #TODO:update
     elif rerun_type == 'conv_bdy_weight':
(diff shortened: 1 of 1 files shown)


ezapartas (dismissed):

sgossage:
Thanks, Matthias.

I think it may be good to add `termination_flags = TF1_POOL_ERROR` to `rerun_type == conv_bdy_weight`, just in case.

In our rerun spreadsheet, you can see that for the HMS-HMS trial grid, `convective_bdy_weight = 0` solved some systems with e.g., `min_timestep` errors as well. So, in case some of those are remaining after the previous reruns, we could try to solve them with this.

mkruckow:
> I think it may be good to add `termination_flags = TF1_POOL_ERROR` to `rerun_type == conv_bdy_weight`, just in case.

I have not done this by intention, to only use it for systems, where it is needed. To my understanding, there is no good reason, why this should help for other systems then the ones with memory issues. For me it feels like, it might helped in a way like opacity_max does on systems it is not meant for. I'd suggest to may check it again after the two reruns before.

sgossage:
Sound good, I'm fine waiting to see how things look after the first two reruns, but my main concern is that we might be left with some `min_timestep` systems still.

My feeling is that with the `convective_bdy_weight` algorithm being experimental, leading to seg faults in some cases, there might be other cases where the algorithm could cause additional convergence issues. It seems like `convective_bdy_weight = 0` alone aided convergence for some of the systems on the HMS-HMS grid.

sgossage (dismissed):

astroJeff (approved):

kkovlakas (approved):

PR #180, opened 2023-11-05
Not printing metallicity value if verbose is False

kkovlakas:


mkruckow (approved):
The main change is small and OK.
I'd like to put up two questions:

1. Should we better merge it into [max_synpop](https://github.com/POSYDON-code/POSYDON/tree/max_synpop) instead of [development](https://github.com/POSYDON-code/POSYDON/tree/development)? So, it is the general question whether we should have different PRs working on the same file, hence need to resolve more potential merge conflicts?
2. Most changes in this PR remove spaces. Do we have an agreed style on this to not changes it back and fourth? (My opinion: we don't need spaces at the end of a line, but I'd prefer to keep spaces in empty lines to see intention blocks better.

maxbriel (approved):
The addition makes sense and is minor.

maxbriel:
@mkruckow For your question 2, I would suggest merging it into development.

I think it's best practice to merge development into the branch before merging the PR and resolve any merge conflicts there before accepting a PR anyway. Merge conflicts can happen but this keeps the commit tree the cleanest.
Otherwise you have a branch of development going into a separate branch it isn't based on.

For 1. does `pylama` automatically remove this additional whitespace, @kkovlakas? Is this something that could be a github action?

PR #181, opened 2023-11-05
Removing duplicated code and simplifying if statements in CEE step

kkovlakas:


posydon/binary_evol/CE/step_CEE.py
```diff
@@ -209,29 +208,17 @@ def __call__(self, binary):
...
 
         # Check to make sure binary can go through a CE
-        if donor_star.state in ['H-rich_Core_H_burning',
-                                'stripped_He_Core_He_burning']:
+        mergeable_donor = (donor_star.state in [
+            'H-rich_Core_H_burning', 'stripped_He_Core_He_burning'])
+        mergeable_HG_donor = (
+            self.common_envelope_option_for_HG_star == "pessimistic"
+            and donor_star.state in ['H-rich_Shell_H_burning'])
+        if mergeable_donor or mergeable_HG_donor:
```
mkruckow, line 216: Do we need the new variables `mergeable_donor` and `mergeable_HG_donor` here? Too me it looks like an unnecessary assignment. Btw. variable assignments are usually slow in python, hence one shouldn't do if you don't need it to be used several times.

mkruckow (approved):

astroJeff (approved):

Issue #178, opened 2023-11-02
api_reference files


mkruckow:
In the documentation, there are files in the directory `docs/_source/api_reference`, those have been automatically generated.

`sphinx-apidoc -o ./_source/api_reference ../posydon -M -t _source/sphinx-apidoc-templates -d 2`

Whenever we have a new file in the POSYDON code, we would need to rerun this command to recreate those files.
Ideally, we should add this to the `makefile` of the documentation to be run, when needed.
This would allow us to remove the `api_reference` directory from the repository. Currently, we have there files, which should never get manually changed. At least in my mind, such files, which should never get edited and being only automatically generated should not be excluded from the repository.

PR #186, opened 2023-11-07
Upgrades on documentation makefile

mkruckow:
This PR addresses [issue#178](https://github.com/POSYDON-code/POSYDON/issues/178).

- [x] The automatically generated files for the code reference are removed.
- [x] Those are automatically created on first build.
- [x] They are recreated if a code file changed.

Additionally, it includes the scripts from the `bin` directory.

- [x] Copy scripts and rename them, that sphinx will detect them.
- [x] Add rule to include them.


maxbriel (commented):
This is very useful and a nice way to implement the automatic creation of the API files.
A few things to resolve:
 
1. Is it possible to change the `api` folder name to `pipeline` or `scripts`? At the moment, this shows up as `api` under the API reference header in the documentation. Something more descriptive would be good.

2.  `make clean` does not empty the `api` folder. Would it be possible to save this the `_build` folder? Or is this not required to be removed when wanting to do a clean-up?

Either add it to the `clean` definition in the make file or move them to the `_build` folder, such that they get removed automatically. 
Actually, should we maybe do the following, such that we're not altering files in _source?
1. copy _source into _build/tmp.
2. copy the pipeline files over. 
3. populate the api files with sphinx-apidoc
4. build html into _build/html
This is just a suggestion/thought.

mkruckow:
>     1. Is it possible to change the `api` folder name to `pipeline` or `scripts`? At the moment, this shows up as `api` under the API reference header in the documentation. Something more descriptive would be good.

This was just, what was there before. I was thinking about renaming as well, but I didn't get to a convincing name yet:

- It is not only pipeline files, hence no to `pipeline`.
- On the one had those are scripts, but only because it is python. Usually, those are subprograms to run (that's way they are in the directory called `bin`), which on the other had does not make too much sense in python.
- Any good ideas?

>     2. `make clean` does not empty the `api` folder. Would it be possible to save this the `_build` folder? Or is this not required to be removed when wanting to do a clean-up?

Does it have to? I can include it in the `clean`.
But no, we can't put it in the `_build` directory, this doesn't exist yet, when the autodoc runs. The `_build` directory is created when building the sphinx documentation. And no, we shouldn't copy the `_source` in a `tmp` directory in `_build` either for the same reason given before.

mkruckow:
- [x] I made the cleaning routine being more general.
- [x] I renamed the directory to `executables` as we get up at lunch.
- [x] I made a `_tmp` directory for the temporary stuff.

maxbriel (commented):
Looks good! `executables` is a better name.

maxbriel (approved):

tassos25:
@kkovlakas can you please have a look at it?

kkovlakas (approved):

Issue #190, opened 2023-11-13
V1 progressbar2 v4.0.0 requirement


maxbriel:
POSYDON V1 has a dependency on `progressbar2 v4.0.0`.
However, this package loads in `python_utils>=3.0.0` of which the newest version requires python v3.8.0.

The latest version that work with python v3.7 is python_utils v3.5.2.

This should be added as a requirement for POSYDON v1 `python_utils=3.5.2`.

mkruckow:
Taken care of in https://github.com/POSYDON-code/POSYDON/pull/191

PR #191, opened 2023-11-13
V1 python_utils fix

maxbriel:
This PR sets the version of `python_utils` to v3.5.2, which is the last version which supports python v3.7. 
This way `progressbar2` which has a dependency on `python_utils>=3.0.0` does not import an unsupported version.

It fixes Issue #190.


tassos25 (approved):
Looks good

astroJeff (approved):

PR #193, opened 2023-11-14
Option for losing mass during cee that leads to merging

ezapartas:
We allow for the possiblity of mass loss during step_CEE, even in the case it leads to oMerging1/2. So we reduce the envelope mass before going to step_CEE, assuming that some of the envelope was indeed ejected during CEE, although in the end the binary did not avoid merging (due to inner RLOF of the core(s))

"mass_loss_during_CEE_merged": False # If False, then no mass loss from this step for a merged star
                                              # If True, then we remove mass according to the alpha-lambda prescription
                                              # assuming a final separation where the inner core RLOF starts.

# we calculate the ejected mass from part of the commone envelope, using
                # a_f = separation_postCEE so that one of the cores (or MS star) is filling its inner Roche lobe,
                # and assuming that lambda(Menvelope) ~ lamda(Mejected) although
                # Mejected < Menvelope (e.g. see Fig1 of Dewi+Tauris2000)

# in double_CE
                    # Assuming that the ratio of ejected mass of each (common) envelope
                    # is the same as the ratio of the initial envelope masses:
                    # the root of the mass lost for each is the solution of a quadratic algebraic equation

We also assume the radius of the surface of the pre-merging stars is np.nan

### DIFF
--- a/posydon/binary_evol/CE/step_CEE.py
+++ b/posydon/binary_evol/CE/step_CEE.py
@@ -828,7 +828,7 @@ class StepCEE(object):
                 # GMMej/(lambda * R)  = a_CE * [ - GM1M2/(2a_i) +  G(M1-Mej)*M2/(2a_f) ]
                 # with a_f = separation_postCEE
                 # and solving it for Mejected
-                Mejected_donor = alpha_CE * (m1_i * const.Msun * m2_i * const.Msun)/2. * (1./separation_i - 1./separation_postCEE) \
+                Mejected_donor = alpha_CE * (m1_i * const.Msun * m2_i * const.Msun)/2. * (1./separation_postCEE - 1./separation_i) \
                           / ( (m1_i * const.Msun/(lambda1_CE*radius1 * const.Rsun)) + ((alpha_CE*m2_i * const.Msun)/(2.*separation_postCEE))  )
 
                 Mejected_donor = Mejected / const.Msun # to go from cgs to solar masses again
--- a/posydon/binary_evol/CE/step_CEE.py
+++ b/posydon/binary_evol/CE/step_CEE.py
@@ -827,7 +827,7 @@ class StepCEE(object):
                 # Ebind(Mejected) = a_CE*D_orb =>
                 # GMMej/(lambda * R)  = a_CE * [ - GM1M2/(2a_i) +  G(M1-Mej)*M2/(2a_f) ]
                 # with a_f = separation_postCEE
-                # and solvind it for Mejected
+                # and solving it for Mejected
                 Mejected_donor = alpha_CE * (m1_i * const.Msun * m2_i * const.Msun)/2. * (1./separation_i - 1./separation_postCEE) \
                           / ( (m1_i * const.Msun/(lambda1_CE*radius1 * const.Rsun)) + ((alpha_CE*m2_i * const.Msun)/(2.*separation_postCEE))  )
 
--- a/posydon/binary_evol/CE/step_CEE.py
+++ b/posydon/binary_evol/CE/step_CEE.py
@@ -42,7 +42,7 @@ from posydon.binary_evol.binarystar import BINARYPROPERTIES
 from posydon.binary_evol.singlestar import STARPROPERTIES
 from posydon.utils.common_functions import PATH_TO_POSYDON
 from posydon.utils.common_functions import check_state_of_star
-from posydon.utils.common_functions import calculate_lambda_from_profile
+from posydon.utils.common_functions import calculate_lambda_from_profile, calculate_Mejected_for_integrated_binding_energy
 
 
 warnings.simplefilter('always', UserWarning)
@@ -783,7 +783,7 @@ class StepCEE(object):
                                             'co_core_mass',
                                             'co_core_radius'
 
-                        ]) 
+                        ])
                 elif star_type == "not_giant_companion":
                     continue
                 else:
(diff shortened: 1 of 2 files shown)


posydon/binary_evol/CE/step_CEE.py
```diff
@@ -484,6 +492,10 @@ def CEE_simple_alpha_prescription(
         core_definition_He_fraction: float
             The value of the He abundance to define the envelope-core boundary
             in CO_cores.
+        mass_loss_during_CEE_merged: Boolean
+             If False, then no mass loss from this step for a merged CEE
+             If True, then we remove mass according to the alpha-lambda prescription
+              assuming a final separation where the inner core(s) RLOF starts, and the same lambda(s)
```
mkruckow, line 498: Using the same `lambda` is a bad idea, because lambda is a function of M_ejected. that's why the definition of the core-envelope boundary is so crucial.
Can't we use the profiles?

posydon/binary_evol/CE/step_CEE.py
```diff
@@ -814,4 +826,92 @@ def CEE_simple_alpha_prescription(
...
+                    # A*Mejected_donor^2 + B*Mejected_donor + C = 0
+
+                    A = alpha_CE / (WF*2.*separation_postCEE)
+                    B = ( ( - alpha_CE*m1_i * const.Msun/ (WF*2.*separation_postCEE)) - (alpha_CE*m2_i * const.Msun/ (2.*separation_postCEE)) \
+                        - (m1_i * const.Msun/ (lambda1_CE*radius1 * const.Rsun)) - (m2_i * const.Msun/ (WF*lambda2_CE*radius2 * const.Rsun)))
+                    C = alpha_CE * (m1_i * const.Msun * m2_i * const.Msun)/2. * (1./separation_postCEE - 1./separation_i)
+                    roots = np.roots([A,B,C])
+                    if min( np.real(roots[0]), np.real(roots[1]) ) < 0.:
+                        root = max( np.real(roots[0]), np.real(roots[1]) ) # if one root is negative, the other is considered the physical solution
+                    elif max( np.real(roots[0]), np.real(roots[1]) ) > m1_i  - mc1_i: # if one root is more positive than the Menvelope that we allow to ejected
```
mkruckow: I think here we have different units on both sides. They should differ by a factor `const.Msun`. Either adding it as factor on the right hand side or dividing the left hand side by it.

posydon/binary_evol/CE/step_CEE.py
```diff
@@ -814,4 +826,92 @@ def CEE_simple_alpha_prescription(
...
+                                                                                     # then the other is considered the physical solution
+                        root = min( np.real(roots[0]), np.real(roots[1]) )
+
```
mkruckow, line 874: We need an `else` clause here, otherwise `root` could be undefined.
I'd suggest to use the minimum in there. Because this is reached first and afterwards there shouldn't be a reason to eject more.

posydon/binary_evol/CE/step_CEE.py
```diff
@@ -814,4 +826,92 @@ def CEE_simple_alpha_prescription(
...
+                    Mejected_donor = root  / const.Msun # to go from cgs to solar masses again
+                    Mejected_comp = Mejected_donor / WF
+
+                    if verbose:
+                        print("The roots are: ", roots / const.Msun, "in Msun")
+                        print("so Mejecta_donor = ", Mejected_donor, "in Msun compared to its initial envelope =",  m1_i  - mc1_i)
+                        print("so Mejecta_comp = ", Mejected_comp, "in Msun compared to its initial envelope =",  m2_i  - mc2_i)
+
+                    if not ( (Mejected_donor <= m1_i  - mc1_i) and (Mejected_comp <= m2_i  - mc2_i) ):
```
mkruckow: Again, the units on both sides of the comparison look to be different.
As you use the original envelope a lot, it might be better to even store this in a variable.

posydon/binary_evol/CE/step_CEE.py
```diff
@@ -814,4 +826,92 @@ def CEE_simple_alpha_prescription(
...
+                        raise Exception("Mejected_donor in double CEE with mass_loss_during_CEE_merged is found more than Menvelope")
```
mkruckow: For this and other raised exceptions: I'd suggest to put in there values. That one gets the value always displayed when the exception is raised. At the moment you'd need to rerun it with verbose=True.

posydon/binary_evol/CE/step_CEE.py
```diff
@@ -814,4 +826,92 @@ def CEE_simple_alpha_prescription(
...
+                    if not (Mejected_donor >= 0.):
+                        raise Exception("The root of the equation in double CEE with mass_loss_during_CEE_merged is found negative")
+
+
+                    if (Mejected_donor > m1_i  - mc1_i) or (Mejected_comp > m2_i  - mc2_i):
+                        print("MANOS", Mejected_donor > m1_i  - mc1_i, Mejected_comp > m2_i  - mc2_i)
```
mkruckow: I guess, this was a print you planed to remove before getting it merged.
Or maybe the whole if here, because it checks the same as the one in line 890, while will raise an exception before the code can get here.

posydon/binary_evol/CE/step_CEE.py
```diff
@@ -814,4 +826,92 @@ def CEE_simple_alpha_prescription(
...
+                        Mejected_donor = (m1_i  - mc1_i) -0.01 # at least this value of envelope is left.
+                        Mejected_comp = (m2_i  - mc2_i) -0.01
+                        warnings.warn("Mejected of at least one star in double CEE is found to be more that the initial envelope. Reduced both to their initial_envelope - 0.01 Msun")
+
+                donor.mass = m1_i - Mejected_donor
+                donor.log_R = np.nan
```
mkruckow: I was thinking, whether we could/should set here a radius estimate:
Usually, I'd expect it to be the Roche-lobe radius, which was assumed to do all this calculation.

astroJeff:
Status Update (Dec 7): @ezapartas to update: if a profile exists (nearest neighbor), calculate mass lost based on alpha using profile. If a profile does not exist, we will eventually want to use @elizabethteng's profile interpolation to calculate a profile, then calculate mass lost with alpha. This will be enables in a future PR/code enhancement - can be accepted without.

posydon/binary_evol/CE/step_CEE.py
ezapartas: I get your point, but it is not really needed because these stars will merge and the log_R will change even further. For now we are agnostic with radii of merged products, we mostly focus on mass and surface/center abundances to  make a crude estimation about the merged product

posydon/binary_evol/CE/step_CEE.py
ezapartas: Now the binding energy is calculated through the profile, not really caring about a lambda for the partial mass ejection

posydon/utils/common_functions.py
```diff
@@ -2578,6 +2578,68 @@ def calculate_binding_energy(donor_mass, donor_radius, donor_dm,
...
+    # Sum of gravitational energy from surface towards inside
+    Grav_energy = 0.0
+    # Sum of internal energy from surface towards.
+    U_energy = 0.0
+    # sum from surface to the core. Your threshold is in element [ind_threshold]
+    # in a normal MESA (and POSYDON) profile
+    i = 0
+    Ebind_so_far = 0.0 # the integration from surface going inwards of the binding energy (negative in principle)
+
+    while abs(Ebind_so_far) < Ebind_threshold:
```
mkruckow: I'd suggest to add here a check on `i` to make sure it will never exceed the number of points in the profiles.

posydon/utils/common_functions.py
```diff
@@ -2578,6 +2578,68 @@ def calculate_binding_energy(donor_mass, donor_radius, donor_dm,
...
+        Grav_energy_of_cell = (-const.standard_cgrav * donor_mass[i]
+                               * const.Msun * donor_dm[i]*const.Msun
+                               / (donor_radius[i]*const.Rsun))
+        # integral of gravitational energy as we go deeper into the star
+        Grav_energy = Grav_energy + Grav_energy_of_cell
+        U_energy = U_energy + specific_internal_energy[i]*donor_dm[i]*const.Msun
+        Ebind_so_far = Grav_energy + factor_internal_energy * U_energy
+        ind_threshold = i
```
mkruckow: Is there a need to set `ind_threshold` in each iteration and not just once after the loop?

posydon/utils/common_functions.py
```diff
@@ -2578,6 +2578,68 @@ def calculate_binding_energy(donor_mass, donor_radius, donor_dm,
...
+        i=i+1
+
+    if donor_mass[ind_threshold]< mc1_i or  donor_radius[i]<rc1_i:
```
mkruckow: Shouldn't the radius be checked on `donor_radius[ind_threshold]` like the mass?

posydon/utils/common_functions.py
ezapartas: Done

posydon/utils/common_functions.py
ezapartas: Done

posydon/utils/common_functions.py
```diff
@@ -2620,18 +2620,18 @@ def calculate_Mejected_for_integrated_binding_energy(profile, Ebind_threshold,
...
         Grav_energy_of_cell = (-const.standard_cgrav * donor_mass[i]
                                * const.Msun * donor_dm[i]*const.Msun
                                / (donor_radius[i]*const.Rsun))
         # integral of gravitational energy as we go deeper into the star
         Grav_energy = Grav_energy + Grav_energy_of_cell
         U_energy = U_energy + specific_internal_energy[i]*donor_dm[i]*const.Msun
         Ebind_so_far = Grav_energy + factor_internal_energy * U_energy
-        ind_threshold = i
         i=i+1
+    ind_threshold = i
```
mkruckow: Because you moved it now behind the increment of `i` it should read `ind_threshold = i-1`

dimsour94 (approved):

mkruckow (approved):

PR #201, opened 2023-11-27
Updates in installation instructions in the V1 documentation

kkovlakas:
- [x] Adding introduction in the installation instructions, explaining the two ways to install/download data, and why we care.
- [x] Restructuring the sections
- [x] Repeating the download-script instruction in the conda installation (and the `bashrc` trick)
- [x] Adding the `git checkout main` instruction before pip installing from github (V1)
- [x] Language check
- [x] Final move to the website (using the `github.io`

### DIFF
--- a/docs/install/install.rst
+++ b/docs/install/install.rst
@@ -5,22 +5,19 @@ Installation
 ############
 
 The easiest way to start with POSYDON is to install it from Anaconda
-(Automatic Installation), and call an executable script that downloads the grid
-data automatically (Automatic Data Download).
+(Recommended way), and call an executable script that downloads the grid
+data automatically.
 
 Alternatively, if you are interested in modifying the code, you can install
-POSYDON manually through `github` (Manual Installation). You can still use the
-automatic script to download the data, however, if you are about to modify the
-grid data, you can get them using the `git-lfs` module (Manual Data Download).
+POSYDON manually through `github` (Manual Installation). Below we describe both
+ways.
 
-Below we describe both ways.
+=======================================================
+Installing POSYDON v1.0.0 from Anaconda (Recommend way)
+=======================================================
 
-==========================================================
-Recommended way (Automatic installation and Data Download)
-==========================================================
-
-Installing POSYDON v1.0.0 from Anaconda (Automatic Installation)
-----------------------------------------------------------------
+Installation
+------------
 
 We recommend to install the Python distribution Anaconda and to install POSYDON
 in a virtual environment. Specifically, we recommend using the installation we
@@ -52,8 +49,8 @@ Now, you can source the environment with
     conda activate posydon
 
 
-Getting the POSYDON data (Automatic Data Download)
---------------------------------------------------
+Getting the POSYDON data
+------------------------
 Export the path to where you want to clone the data, e.g. `/home/`, and
 download the data from ZENODO with the following commands
 
@@ -69,7 +66,6 @@ download the data from ZENODO with the following commands
 ==========================================================
 Installing POSYDON v.1.0 from GitHub (Manual Installation)
 ==========================================================
-
 
 Creating a conda environment
 ----------------------------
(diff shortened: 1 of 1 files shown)


docs/install/install.rst
```diff
@@ -4,22 +4,36 @@
...
-conda environment can be created (we have named our environment posydon-example, 
-but you can choose any name), the conda-forge channel added, and the required 
+The easiest way to start with POSYDON is to install it from Anaconda
+(Automatic Installation), and call an executable script that downloads the grid
+data automatically (Automatic Data Download).
+
+Alternatively, if you are interested in modifying the code, you can install
+POSYDON manually through `github` (Manual Installation). You can still use the
+automatic script to download the data, however, if you are about to modify the
+grid data, you can get them using the `git-lfs` module (Manual Data Download).
```
mkruckow: Is the git-lfs still in place? I remember Simone told me once, this isn't available anymore. On the other hand, do we need this for v1? Shouldn't we there always use the data on Zenodo?

mkruckow:
I think, in the automatic installation now there is the part about exporting the global path missing.
`export PATH_TO_POSYDON=/home/POSYDON/`
or is this automatically done?
If it is done automatically, is it done only for the session or added to the `.bash_rc` or `.bash_profile`? I haven't seen anything in the repo doing this.
Finally, I'm wondering, whether we should have a default for `export PATH_TO_POSYDON_DATA=/home/` or `export PATH_TO_POSYDON_DATA=$PATH_TO_POSYDON/data/`, in the case of automatically doing one export to do all.

astroJeff:
I just reviewed the language. I made some changes (not just to your text, but what was there before. Please read through to make sure you agree with my updates.

tassos25 (approved):
Looks good

maxbriel:
These changes look good. I would recommend adding the following information to the Mac section, because many packages for python3.7 are not available on osx-arm

After creating an environment the user has to change the architecture channels. This will automatically enable Rosetta to translate the osx64 architecture into osx-arm. Although many things should just work, I expect it to reduced performance and I have not tested this extensively to see if anything breaks. 

```
conda create --name posydon_conda

conda activate posydon_conda

 ## use x86_64 architecture channel(s)
conda config --env --set subdir osx-64

conda install python=3.7

conda install posydon

```

kkovlakas:
> I think, in the automatic installation now there is the part about exporting the global path missing. `export PATH_TO_POSYDON=/home/POSYDON/` or is this automatically done? If it is done automatically, is it done only for the session or added to the `.bash_rc` or `.bash_profile`? I haven't seen anything in the repo doing this. Finally, I'm wondering, whether we should have a default for `export PATH_TO_POSYDON_DATA=/home/` or `export PATH_TO_POSYDON_DATA=$PATH_TO_POSYDON/data/`, in the case of automatically doing one export to do all.

Thank you for considering this @mkruckow . Actually, `PATH_TO_POSYDON` is only used in unit-testing and running grids, so we define it only as developers. Users installing via `conda` or `pip install` won't need it. Although there is a variable defined in `common_function.py` which read the environment variable, it's empty if undefined, and POSYDON works normally.

kkovlakas:
> These changes look good. I would recommend adding the following information to the Mac section, because many packages for python3.7 are not available on osx-arm
> 
> After creating an environment the user has to change the architecture channels. This will automatically enable Rosetta to translate the osx64 architecture into osx-arm. Although many things should just work, I expect it to reduced performance and I have not tested this extensively to see if anything breaks.
> 
> ```
> conda create --name posydon_conda
> 
> conda activate posydon_conda
> 
>  ## use x86_64 architecture channel(s)
> conda config --env --set subdir osx-64
> 
> conda install python=3.7
> 
> conda install posydon
> ```

Thank you @maxbriel , I add this extra set of instruction -however, I am not sure about conda channels, can I include them?

maxbriel:
I've added the additional conda channels to the instructions, although I don't think they're explicitly necessary to define.

maxbriel (approved):

PR #204, opened 2023-11-28
added a new rerun type: hemb_mltp_mesh

sgossage:
This adds a new rerun type `hemb_mltp_mesh` which will introduce several changes to the MESA inlists found to help convergence.

1. **Issue: HeMB is not implemented, but magnetic braking is suspected of causing convergence issues on the CO-HeMS grid. So, during reruns, magnetic braking would be turned off for that grid. This will need to be disabled in the `run_binary_extras.f` file.**
2. MLT+ (mltp) is implemented, and changes the parameter `gradT_excess_f2` to 0.1 (from a value of 0.001). A larger value of this parameter makes the superadiabaticity reduction of MLT++ weaker (with 1 meaning no reduction).
3. "mesh" is implemented and refers to changes in MESA's mesh algoorithm, adopted from MIST, which were found to help convergence. These are:

```
      max_logT_for_k_below_const_q = 100
      max_q_for_k_below_const_q = 0.995
      min_q_for_k_below_const_q = 0.995
      max_logT_for_k_const_mass = 100
      max_q_for_k_const_mass = 0.99
      min_q_for_k_const_mass = 0.99

```

### DIFF
--- a/bin/run-pipeline
+++ b/bin/run-pipeline
@@ -451,6 +451,8 @@ def copy_ini_file(grid_path, rerun_type, destination, cluster):
         replace_text = "development-22c1bb9e730343558c3e70984a99b3fc1f3c346e"
     elif rerun_type == 'thermohaline_mixing':
         replace_text = "development-22c1bb9e730343558c3e70984a99b3fc1f3c346e"
+    elif rerun_type == 'hemb_mltp_mesh':
+        replace_text = "development-22c1bb9e730343558c3e70984a99b3fc1f3c346e"
     elif rerun_type == 'other': # e.g. 'envelope_mass_limit', 'fe_core_infall_limit'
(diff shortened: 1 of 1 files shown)


tassos25:
@sgossage a switch to turn of MB has been implemented in another PR by @ezapartas . Please make sure that the two are not conflicting.

mkruckow:
This is not the way we should do. It would fail, because the directory names would get too long.
Additionally, there are parts missing.

The aim of this PR is covered in [PR#177](https://github.com/POSYDON-code/POSYDON/pull/177) and the MESA-INLISTS [PR#27](https://github.com/POSYDON-code/POSYDON-MESA-INLISTS/pull/27) and [PR#28](https://github.com/POSYDON-code/POSYDON-MESA-INLISTS/pull/28)

Issue #205, opened 2023-11-29
Binaries entering "undefined" state, not completing evolution


celiotine:
There are many binaries that are not completing their evolution because they enter an "undefined" state and then immediately go to the "END" event. This only happens when S1 is a compact object, S2 is in a non-burning state ('stripped_He_non_burning'' or 'H-rich_non_burning'), and the binary is going through the CO_HMS_RLO or CO_HeMS steps:

<img width="1010" alt="Screen Shot 2023-11-29 at 1 14 24 PM" src="https://github.com/POSYDON-code/POSYDON/assets/13874913/2a62c306-3b31-4885-beff-ee58684115bf">

These binaries do not fail, but produce a warning, which is making it difficult to debug:

<img width="787" alt="Screen Shot 2023-11-29 at 1 40 51 PM" src="https://github.com/POSYDON-code/POSYDON/assets/13874913/044e8d2a-f003-44ce-a426-39132bb0c043">

I noticed that the "undefined" state is not in the flow. It is only defined in 'utils/common_functions.py' in the function `get_binary_state_and_event_and_mt_case()`. I am wondering if this is the issue. However, if you look at this function, it also seems these binaries should not reach this state in the first place, as they all are in RLO2 when this happens.

To reproduce this warning for a single binary, you can evolve the following binary in a notebook using the current development branch and get the simulation properties from the given file on Quest: 

```
sim_kwargs = simprop_kwargs_from_ini("/projects/b1119/cliotine/dev_branch_pops/params_v2.ini")
sim_prop = SimulationProperties(**sim_kwargs)
sim_prop.load_steps(verbose=True)

star_1 = SingleStar(**{'mass': 7.294196, 'state': 'H-rich_Core_H_burning', \
                                 'natal_kick_array': [57.65029, 5.359689, 1.86049, 6.200635]})
star_2 = SingleStar(**{'mass' : 7.028671, 'state' : 'H-rich_Core_H_burning'})
binary = BinaryStar(star_1, star_2,
                    **{'time': 0.0, 'state': 'detached', 'event': 'ZAMS', 'orbital_period': 77.926729, 'eccentricity': 0.0},
                    properties = sim_prop)

binary.evolve()

```

celiotine:
[...]

PR #206, opened 2023-11-30
Allow non-burning stars in MT

astroJeff:
We have added code that allows non burning stars to enter mass transfer without failing immediately. @celiotine will test this. Addresses Issue #205.

### DIFF
--- a/posydon/utils/common_functions.py
+++ b/posydon/utils/common_functions.py
@@ -1445,7 +1445,9 @@ def infer_mass_transfer_case(rl_relative_overflow,
                   rl_relative_overflow, lg_mtransfer_rate)
         return MT_CASE_NO_RLO
 
-    if "H-rich" in donor_state:
+    if "non_burning" in donor_state:
+        return MT_CASE_NONBURNING
+    elif "H-rich" in donor_state:
         if "Core_H_burning" in donor_state:
             return MT_CASE_A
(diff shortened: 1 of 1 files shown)


celiotine (changes requested):
I tested these changes and the issue was not fixed. Looking more closely at the code, the updated logic in `infer_mass_transfer_case` is not quite correct. It checks for substrings that also exist in the `non_burning` states first, so it enters these statements, finds no MT case, and returns undetermined. Moving the check for `non_burning` first fixes this issue, and there are no longer any `undefined` binaries in the population.

celiotine:
I have looked more thoroughly at the "fixed" tracks and most (>90%) of the stars undergoing `stripped_He_non_burning` look this, which seems odd to me:
Screen Shot 2023-12-05 at 11 13 30 AM
i.e. they go to `initial_RLOF` and then end.

celiotine:
A nice example of a corrected track after fixing w/ my comment above:
Screen Shot 2023-12-07 at 11 54 16 AM

celiotine (approved):

astroJeff:
Update (Jan 18): We need one more test of this. @ezapartas and @dimsour94 to test

kkovlakas (approved):

ezapartas (approved):

dimsour94 (approved):

PR #210, opened 2023-12-06
Numerical precision fix failed

mkruckow:
- [x] Adding a rule if the new and old age_to_remove would be the same.

In such a case, the numerical precision fix would fail, because there won't be a change to fix anything.
To explain, when this happens: it occurs in the rare case, the normal division would be an integer already. Thus the floor division will give the same result, which cancels with the multiplication. The solution is simply enforcing in such a case to go down by 1 before the multiplication. Mathematically it simply changes `new_age_to_remove<=age_to_remove` to `new_age_to_remove<age_to_remove`.

### DIFF
--- a/posydon/grids/scrubbing.py
+++ b/posydon/grids/scrubbing.py
@@ -139,7 +139,7 @@ def keep_after_RLO(bh, h1, h2):
         min_dt = min(np.diff(new_bh["age"]))
         new_age_to_remove = (age_to_remove // min_dt) * min_dt
         if new_age_to_remove==age_to_remove:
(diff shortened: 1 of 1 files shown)


ezapartas (dismissed):

kkovlakas (dismissed):

kkovlakas (approved):

mkruckow:
OK, I reminded myself. Subtracting `min_dt` is even better, because it will be less effected by numerical truncation than the division.
I hope, this statement holds for python, which doesn't apply all hardware limits in the same way as other programming languages.

astroJeff (approved):

Issue #215, opened 2023-12-08
Reading a binary from a hdf5 file to evolve fails


maxbriel:
When setting up or running a binary population to read from an hdf5 file for the initial state of the binaries.

In `PopulationManager().from_hdf()` the process fails in `BinaryStar.from_df(hist.loc[i], extra_columns=self.kwargs.get('extra_columns', []))`.

`hist.loc[i]` returns a pandas.Series, as such fails in `BinaryStar.from_df()` when its index/index name is requested.

A quick solution is to call `hist.loc[[i]]` which returns a pandas.DataFrame, which will have a index/index name.

PR #216, opened 2023-12-08
Change loc call to list to always return a pandas.DataFrame

maxbriel:
This is a problem when the user is (re-)evolving a BinaryPopulation by reading the binary from an HDF5 file.

The change to an input list ensures that a `pandas.DataFrame` is returned instead of a `pandas.Series`. `BinaryStar.from_df` expects a `pandas.DataFrame` as an input.

Solves issue #215

### DIFF
--- a/posydon/popsyn/binarypopulation.py
+++ b/posydon/popsyn/binarypopulation.py
@@ -614,7 +614,7 @@ class PopulationManager:
         binary_holder = []
         for i in np.unique(hist.index):
             binary = BinaryStar.from_df(
-                hist.loc[i],
+                hist.loc[[i]],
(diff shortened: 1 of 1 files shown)


ka-rocha (approved):
Nice catch!

mkruckow:
Is this the only case in the code, where we should do this change?
I'd expect a lot of other places in the code may needing this change, too. Could you please check for them?

maxbriel:
As far as I can see with a search for this structure, this is the only place where this is an issue.
A bit further down in the same function the correct implementation is done for example. I think this was just an oversight. 

```if bool(oneline.loc[[i]]['FAILED'].values[-1]):```

astroJeff (approved):

Issue #138, opened 2023-09-11
How to remove data points in RLO grid?


mkruckow:
I have the impression, we get an issue with a case not considered in the RLO girds:

To save space grids which are aimed to start at RLO do not contain any system, which did not go into RLO. So far OK.
The join routine only overwrites data which is present in the grid with the rerun. So far OK.

But now, we get a rerun, which turns a system going into RLO into a noRLOF system. Here we get the problem, that this system is not stored in the RLO grid of the rerun, hence the join routine doesn't know, that this systems was rerun and need to be overwritten. In fact, it needs to be deleted from the PSYgrid object, therefore first we'd need to keep the information, which systems where run in the RLO gird even they have no RLO.

mkruckow:
Some more details can be found on slack developer's channel: https://posydon.slack.com/archives/CDCS3MS31/p1694519159696839

PR #217, opened 2023-12-08
Allows non-Roche lobe overflowing systems in the RLO grids

kkovlakas:
- [x] Allow non-Roche lobe overflowing systems in the RLO grids so that reruns can update them.
- [x] Change what is reported in case of failed system that are non-RLO: termination flags should mention "failure" instead non-RLO, so that we keep track of these systems
- [x] A new class `IgnoreReason` that allows ignore reason (when processing runs in new grids) to keep a certain order. Also this helps in reducing the code in `psygrid.py`
- [x] Test that the detached step and flow can handle the new termination flag `ignored_no_RLO`
- [x] Test that it allows to join rerun grids with updating the non-RLOing systems


mkruckow:
One thing we don't get for those systems would be "X", "Y", and "Z" in the initial values.
I'm not sure, why those values aren't determined for ignored systems and whether this could cause trouble not having them.

astroJeff:
Status Update (Dec 14): Still needs to be tested. Are there other places in the code where this needs to be fixed as well? @kkovlakas and @mkruckow to resolve XYZ abundances as raised above.

mkruckow:
I have tested to use it to join grids: it works on the data level, but it causes our plotting routines to fail. Hence, I'm skeptical that it will work well in the detached step, but maybe this code part is more robust. Beside that, we first need to check, whether the interpolator will have issues with a new interpolation class.

The reason for the failing is are histories being None, especially the history in the first run in the grid, which @ssbvr often used as template for the grid, without checking, that it is not None. This for example causes a call for a dtype to fail. It is even worse, when this happens inside try statements, which prevent python from reporting where it failed.

Thus, I'm not sure, whether this PR addresses all aspects of issue #138, yet. I opened a new issue #236, which is a generalization of the problem I run into when testing this PR.

astroJeff:
Update (Jan 25): @kkovlakas to finalize and test with a population run (see @mkruckow's comments above).

posydon/grids/psygrid.py
```diff
@@ -2165,14 +2175,24 @@ def say(something):
...
-    say("    {} substituions detected.".format(n_substitutions))
+
+            if (grid[run_index].final_values["termination_flag_1"]
+                    == "ignored_no_RLO"):
+                if params_from_path in initial_params:
+                    del initial_params[params_from_path]
+                n_ignored_noRLO += 1
+
+    say("    {} substitutions detected.".format(n_substitutions))
+    say("    {} runs ignored beacuse of no RLOF.".format(n_ignored_noRLO))
```
mkruckow: Just a minor thing: should we have this output in case of a grid not starting at RLO? I'd say it would be better to only output this for grids starting at RLO. Hence, I'd put it behind:
`if (newconfig["start_at_RLO"]):`

mkruckow:
We need to decide how to deal with cases, where the run crashed and no RLO was detected before the crash. Currently, the no RLO would overwrite the information about the crash. I'd suggest to keep the information about the crash.

posydon/grids/psygrid.py
kkovlakas: Done!

astroJeff:
Update (Feb 15): @kkovlakas explained the status of this PR. Termination flags for non-RLO systems need to be changed. @kkovlakas will take care of this.

astroJeff:
Update (Feb 22): No update on this PR. Waiting for @kkovlakas to finish updates

mkruckow:
Update (Feb 29): @kkovlakas hopes to get back to it next week.

kkovlakas:
Update (Mar 3): progress was made and the PR is ready for testing.

kkovlakas:
@mkruckow I am not proud of this patch. Ideally, a new class could be created, e.g., `IgnoreReason` which ensures that the correct ignore reason is used (see priority) during the assignment. For example:

```Python
ignore = IgnoreReason()    # automatically None
ignore.reason = "A"    # updating from None to anything from the permitted values 
ignore.reason = "B"    # B is updated only if it has higher priority than "A"
```

Let me know if you prefer this solution (also moving it to a different file so that the `psygrid.py` is less busy.

kkovlakas:
> @mkruckow I am not proud of this patch. Ideally, a new class could be created, e.g., `IgnoreReason` which ensures that the correct ignore reason is used (see priority) during the assignment. For example:
> 
> ```python
> ignore = IgnoreReason()    # automatically None
> ignore.reason = "A"    # updating from None to anything from the permitted values 
> ignore.reason = "B"    # B is updated only if it has higher priority than "A"
> ```
> 
> Let me know if you prefer this solution (also moving it to a different file so that the `psygrid.py` is less busy.

Actually I implemented it because it also simplifies the code in other places.

kkovlakas:
When PSyGrid processes runs, an instance of `IgnoreReason` is created:

```Python
ignore = IgnoreReason()
```
which results in `ignore.reason` to be `None` (initial state).

Assignment is only allowed for property `reason`, and only from a list of predefined strings:
```Python
ignore.something = 13                       # this has no effect
ignore.reason = 'ignored_no_BH'             # this is the new value of ignore.reason
ignore.reason = 'because'                   # this results in a ValueError because `because` is not allowed
ignore.reason = 'corrupted_binary_history'  # this updates the value
ignore.reason = 'ignored_no_RLO'            # this is not updating the reason because it has lower priority!
ignore.reason = None						# always allowed to rest the ignore reason
```

More importantly, `ignore_data = True` is not necessary anymore! It can be checked in this way (through the reason):
```Python
if ignore:
    print("The data will be ignored because of", ignore.reason)
```
which, under the hood, checks if the `ignore.reason` is not `None`. Note that his means a reason for ignoring data **must always be reported**. This was my intention in `PSyGrid` but having two separate variables (a flag and a string) allowed for mistakes.

astroJeff:
Update (Feb 7): Max pointed out that this may change the grid boundaries. Konstantinos to look into this - could be resolved during merge of grids. This PR seems to be ready, but needs to be tested - Matthias to look into.

astroJeff:
Can this PR also be used to address Issue #254 in which some runs fail with multiple MESA output error codes? @kkovlakas to discuss with @mkruckow

posydon/utils/ignorereason.py
```diff
@@ -0,0 +1,48 @@
...
+
+IGNORE_REASONS_PRIORITY = [
+    'corrupted_binary_history',
+    'corrupted_history1',
+    'corrupted_history2',
+    'ignored_no_BH',
+    'ignore_no_H1',
+    'ignored_scrubbed',
+    'ignore_no_FP',
+    'ignored_no_RLO']
```
mkruckow: I suggest to change the priority order. E.g. having no history should have a higher priority to having a corrupted history.
Additionally, I'd like to think about aligning then reasons more, e.g. using the short version BH vs. the long version binary_history or whether we need the key word `ignore` in this class anymore.
Finally, I'd suggest to add already some more pre defined reasons for future use, e.g. `ignore_no_H2` and/or being more specific by splitting `ignored_scrubbed` into `ignored_scrubbed_BH`, `ignored_scrubbed_H1`, `ignored_scrubbed_H2`.
So, I'd suggest:
```
IGNORE_REASONS_PRIORITY = [
    # missing data files
    'no_history1',              # always needed
    'no_binary_history',        # for all binaries needed, while not for single stars
    'no_history2',              # needed, if there are two stars in the binary
    # corrupted data files
    'corrupted_history1',       # always needed
    'corrupted_binary_history', # for all binaries needed, while not for single stars
    'corrupted_history2',       # needed, if there are two stars in the binary
    # scrubbed data
    'scrubbed_history1',        # always needed
    'scrubbed_binary_history',  # for all binaries needed, while not for single stars
    'scrubbed_history2',        # needed, if there are two stars in the binary
    # issues with final profile data
    'no_final_profile',         # missing file
    'corrupted_final_profile',  # corrupted file
    'scrubbed_final_profile',   # scrubbed data
    # not needed data
    'no_RLO']
```

posydon/utils/ignorereason.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+class IgnoreReason():
+    """Class for handling ignore reasons when creating new grids."""
+
+    def __init__(self):
+        self.reason = None
+        self.order = None
```
mkruckow, line 25: I'd suggest to use only integers for `order`. Thus, I'd use e.g. `self.order=np.inf` if `self.reason==None`. This would ensure, that e.g. `1<None` will never be tried, which causes an error and we don't need to check for the reason to be `None` extra, because any index will be smaller than infinity.

posydon/grids/psygrid.py
```diff
@@ -891,12 +886,11 @@ def decide_columns(key_in_config, defaults):
                     if self.config["accept_missing_profile"]:
                         warnings.warn("Including MESA run despite the missing "
                                       "profile in {}\n".format(run.path))
-                        ignore_data = False
+                        ignore.reason = None
```
mkruckow: Should we may better remove this line, because it is not doing anything, while may confusing. It will only reset ignore to `None`, if it was `None` before.

mkruckow:
> Can this PR also be used to address Issue #254 in which some runs fail with multiple MESA output error codes? @kkovlakas to discuss with @mkruckow

I'd say, this is a very different issue. This PR deals with stuff internally in the PSyGrid. The issue #254 is about dealing with an issue in the MESA output file `out.txt`.

mkruckow:
Update (March 21): @kkovlakas should checkout the latest comments. Otherwise it is close to be finished.

astroJeff:
Update (Apr 11): PR is close to completion - a couple minor comments from @mkruckow need to be taken care of. @kkovlakas to finish up.

posydon/utils/ignorereason.py
kkovlakas: @mkruckow I will change the order of the errors as you mentioned. I am reluctant to put flags that do not exist yet in `psygrid` because there might be reasons for that. For example, the `scrubbed_history1` is not necessary because for a scrubbing history issue, we don't really need to have this resolution - it's a complicated problem that needs manually checking all history files, and it may be the case that don't know where the issue is (e.g., if you have a binary history and a history file with no overlap in the steps... you don't know which one if off).

I will change some names for consistency though - thanks for inspiring this.

posydon/utils/ignorereason.py
kkovlakas: I'd prefer not to change this:
1. First of all, infinity could be set to ensure something goes to the edge of the list, so I wouldn't encourage using it instead of `None`
2. You say to use only integers, but `np.inf` is not an integer.
3. The `1 < None` for me is not bad - because it will cause an error and the developer will really start to understand the logic of this object.
4. For the `non-developer`, this is not a problem because they cannot actually access the `.order` property outside the object. So we're safe.

posydon/grids/psygrid.py
kkovlakas: It actually does - you can always reset the ignore reason to `None`, you just can't set it to a lower-order ignore reason. So this does have an effect.

posydon/utils/ignorereason.py
mkruckow: You introduced some new inconsistencies: e.g. 'ignored_no_H1' vs. 'corrupted_history1'.

posydon/grids/psygrid.py
mkruckow: That's true. Here I'm more questioning, why this line is there at all because it is behind a `if not ignore`
Hence for me the effective code reads:
```
if ignore.reason == None:
    ignore.reason = None
```

posydon/utils/ignorereason.py
kkovlakas: If you mean about `H1` vs `history1` I fixed that now, thanks.

mkruckow: Mainly this, yes.
But I'd leave out the "ignored" as well or put it everywhere, because all reasons are a reason to have something ignored. (not a requirement for approval)

posydon/grids/psygrid.py
kkovlakas: I see what you mean. I guess I put it there for emphasis. But it's OK now, I removed it.

mkruckow (approved):
Looks fine to me.

astroJeff (approved):

Issue #194, opened 2023-11-16
PopSynth: Binary stuck between steps `detached` and `CO_HMS`


maxbriel:
**Problem**
A binary with a massive BH outside the `CO_HMS` grid mass will get redirect to the `detached` step. However, if the binary then experiences RLOF, it will be moved to the `CO_HMS` grid at RLOF, where it falls outside the grid and gets redirected back to `detached`. Since this binary is undergoing RLOF, it gets put back to `CO_HMS`. This causes a cyclical process.
This process repeats itself several times with a different number of repetitions for different binaries.

Interestingly, the period reduces slightly every time it goes from `step_CO_HMS_RLO` -> `step_detached`. Tides?

**Solution**
_Temporary:_ Stop the systems from being send to the CO_HMS grid if they already came from there?
_Actual solution:_ extend the secondary grid to include such massive BHs.

**Example**
ZAMS binary at $Z=0.00142$.
S1_mass | S2_mass | Period | eccentricity
-- | -- | -- | -- 
170.638207 | 37.917852 | 113.352736 | 0

Evolution:
step_names |state | event | S1_mass | S2_mass | S1_state | S2_state | orbital_period | eccentricity 
-- | -- | -- | -- | -- | -- | -- | -- | -- 
initial_cond | detached | ZAMS | 170.638207 | 37.917852 | H-rich_Core_H_burning | H-rich_Core_H_burning | 113.352736 | 0.000000
step_HMS_HMS | detached | CC1 | 150.423664 | 35.663753 | H-rich_Central_He_depleted | H-rich_Core_H_burning | 131.463720 | 0.000000 
step_SN | detached | NaN | 148.677267 | 35.663753 | BH | H-rich_Core_H_burning | 132.374753 | 0.018719 
step_detached | RLO2 | oRLO2 | 148.677267 | 34.844310 | BH | H-rich_Core_He_burning | 128.487365 | 0.000000
step_CO_HMS_RLO | detached | redirect | 148.677267 | 34.844310 | BH | H-rich_Core_He_burning | 128.487365 | 0.000000 
step_detached | RLO2 | oRLO2 | 148.677267 | 34.893625 | BH | H-rich_Core_He_burning | 128.456467 | 0.000000

etc etc

tassos25:
Can you also please add the ID of the system and where the relevant population datafile can be found?

maxbriel:
[...]

PR #222, opened 2023-12-14
Add error for returning to step_detached after leaving it.

maxbriel:
This should stop systems from going into loops between the `detached` and `CO_HMS_RLO` grids, when the compact object is outside the range of the grid.

This fixes issue #194


tassos25 (approved):

astroJeff (approved):

mkruckow:
Please add this for the other CO gird, too.

PR #223, opened 2023-12-14
minor bug fixes to profile interpolation

elizabethteng:
fixed a few issues in the density class. already tested


astroJeff:
Update (Jan 18): We need Elizabeth on a developers' call to explain these bug fixes before we can accept.

kkovlakas (approved):

mkruckow (approved):

Issue #47, opened 2023-03-22
path MESA column list -- submission script


ssbvr:
I am reporting here something that might be an issue for people when running MESA grids. The `posydon-setup-grid` script clones the inlists submodule in the home directory of the user. Then, it does the layering of the inlists in the working directory which are copied over in the simulation directory once the job starts. However, the files `binary_history_columns.list`, `history_columns.list` and `profile_columns.list` are copied to the job directory from the cloned inlist submodule and not from the working directory. If someone deletes such a directory or runs a different version of the files, it will mess up the previously submitted simulations. Seth run again such a problem last week.

astroJeff:
From a brief glance at the code, @ssbvr, this looks fixed. Can you confirm?

ssbvr:
I think this issue was never fixed.

PR #228, opened 2023-12-19
Copy column list for MESA

mkruckow:
This PR addresses issue #47.

- [x] Copy column list into work directory and use them for the runs.
- [x] Add waits on git checkout.
- [x] testing

### DIFF
--- a/bin/posydon-setup-grid
+++ b/bin/posydon-setup-grid
@@ -132,8 +132,7 @@ def find_inlist_from_scenario(source, gitcommit, system_type):
                                  stdout = subprocess.PIPE,
                                  stderr = subprocess.PIPE
                              )
-    proc.wait()
-
+    
 
     print("checking out commit/tag: {0}".format(githash))
 
--- a/bin/posydon-setup-grid
+++ b/bin/posydon-setup-grid
@@ -984,6 +984,20 @@ if __name__ == '__main__':
                                                           grid_parameters=grid_df.columns,
                                                           working_directory=args.run_directory)
 
+    # handle column lists
+    # first, creating a directory
+    column_lists_folder = os.path.join(args.run_directory, 'column_lists')
+    if os.path.exists(column_lists_folder):  shutil.rmtree(column_lists_folder)
+    os.makedirs(column_lists_folder)    
+    # second, getting new location
+    star_history_columns = os.path.join(column_lists_folder, 'history_columns.list')
+    binary_history_columns = os.path.join(column_lists_folder, 'binary_history_columns.list')
+    profile_columns = os.path.join(column_lists_folder, 'profile_columns.list')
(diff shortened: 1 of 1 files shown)


astroJeff:
We briefly discussed this PR. @mkruckow has tested this, but we want @sgossage to test it before accepting.

astroJeff:
Update (Jan 18): We discussed this PR briefly, but we want Seth to test before accepting

sgossage:
I've tested this PR and it does solve issue #47. I think it is ready to merge.

I remember that there may be a related niche issue to consider though when running grids using various branches of `POSYDON-MESA-INLISTS`. Currently when `posydon-setup-grid` tries to clone the inlist submodule to the home directory, it first checks if `.posydon_mesa_inlists` already exists there and will not clone the inlist submodule if it does exist. So, if you run one grid using an inlist submodule from `branch_x` and then try running another with `branch_y`, you have to remember to delete `.posydon_mesa_inlists` in your home directory in order to clone the inlist submodule of `branch_y` (or make local edits to inlist files).

I'm not sure if it would be worth disabling the check to only download clone the inlists submodule if it doesn't exist, but wanted to bring this up in case people think it may be worthwhile to change.

mkruckow:
Yes, there is this issue, that sometimes the old instead of the new branch is used.
But I think, this is not related to not be cloned. The code only clones the repo if non is there. This is correct in terms of efficiency. In any case the new branch is checked out and pulled  Hence, it should get updated. But I think, often the code proceeds further, before the git actions are completed on disk. Thus, it uses old data from disk for the copy.
Why I think it is the case that the git actions are simply delayed instead of not performed, like @sgossage suggested: if you do the `posydon-setup-grid` twice, it will usually have the new version, because till than the first git actions are completed on disk. For a prove, after the update to the new branch didn't work, check the current branch in `.posydon_mesa_inlists` by hand. If @sgossage is correct, it will have the old branch, if I'm correct it will show you the desired branch.

mkruckow:
@kkovlakas , would you think that it would be sufficient to add a `proc.wait()` in line 144 or would we need this several times for each call of the subprocess module?

astroJeff:
Update (Jan 25): There is an issue listed above about adding a wait statement to make sure the subprocess has completed before moving on. @mkruckow to test.

mkruckow:
OK, I have inserted three waits. On made the code fail, has I removed this again. Now it runs through.
As I can't enforce a problematic situation, where the file system is slow on writing, I can't test it further.

If anybody has an idea who to ensure a more detailed test, please do so!
Otherwise, @sgossage , @astroJeff , @kkovlakas please do a final review of this PR.

kkovlakas (approved):

astroJeff (approved):

Issue #224, opened 2023-12-18
BinaryPopulation: Out of Memory


maxbriel:
When running a large population on a cluster, the BinaryPopulation goes out of memory even when the mem_per_cpu and maximum CPU in the population file are set the same.

My expectation is that some binaries take up more memory than expected, probably due to the "looping" issue described in Issue #194. As such, slurm kills the job before POSYDON can dump the binary into a file. However, I have not investigated this in detail.

@ka-rocha have you ran into this issue? Or suggestions on the possible origin for this?

maxbriel:
I've performed a few test to check what the memory consumption is with different options. 

First, I performed 4 different runs:
1. No binary
2. 10 binaries with `dump_rate=10` while `ram_per_cpu=None`
3. 10 binaries with a `dump_rate=1` while `ram_per_cpu=None`
4. 10 binaries with a `dump_rate=1` while `ram_per_cpu=4`

![test_runs](https://github.com/POSYDON-code/POSYDON/assets/14039563/e4ee9d44-b1b9-4c8e-bf50-d4871838954c)

Both **2.** and **4.** keep all binaries in memory as expected from the code. [Is this wanted behaviour?]

However, **3.** shows unexpected behaviour, it's memory usage keeps increasing, while binaries are written to file.
The below figure shows the calls to `save` for this run between the blue brackets. It's clear that even though the file gets saved, the memory usage slowly keeps increasing.

![Figure_1](https://github.com/POSYDON-code/POSYDON/assets/14039563/65abb745-4bd3-45e8-a3ad-0d973d16ba8e)

maxbriel:
The initial increase comes from loading in the interpolators (approx 1.2 GB). The next increases in the plots come from `binary.evolve`, which is expected. In the first few calls to `step_detached` additional memory is allocated for different grid `DataScaler()` in `detached_step.scale()`

maxbriel:
This Out-of-Memory error is difficult to track down. Here's a summary:
1. It is not the binaries themselves. They only take up a small amount of space compared to grids etc
2. It is not the initial loading in of the grids and interpolators. This is about 1.2 GB. 
3. Three possible origins:
    1. The garbage collection does not function correctly on the cluster/yggdrasil?
    2. A dataset allocation in a package calling `h5py`, once in a while?
    3. IF interpolator?

[...]

PR #229, opened 2023-12-21
HDF5 read fix

maxbriel:
When (re-)evolving a BinaryPopulation from an HDF5 file, the file was kept open while items were being read.
Because the binaries are stored in a table format, pandas leeks memory (See Pandas [Issue](https://github.com/pandas-dev/pandas/issues/22082)). In our Issue #224, the problem is described in more detail.

The below plot shows 3 runs in a row each with all 8 metallicities and 100 binaries per metallicity. 
The first run is with random generated binaries. 
The second with the constant open pd.HDFStore in PopulationManager. 
The third with this fix.

There's still a small increase present in the random and fix run, but it's unclear if this is actually a continuous increase or just an increase due to metallicity switches. For now, this fix should solve evolving binaries from an HDF5 file.

The generated run and third run differ a bit in memory usage. This is likely due to the type of binaries being ran. If more grids have to be loaded in the `detached_step` then more memory will be used.

Screenshot 2023-12-21 at 16 33 51

### DIFF
--- a/posydon/popsyn/binarypopulation.py
+++ b/posydon/popsyn/binarypopulation.py
@@ -464,7 +464,7 @@ class BinaryPopulation:
     def close(self):
         """Close loaded h5 files from SimulationProperties."""
         self.population_properties.close()
-        self.manager.close()
+
 
     def __getstate__(self):
         """Prepare the BinaryPopulation to be 'pickled'."""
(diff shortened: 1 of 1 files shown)


mkruckow (approved):

maxbriel:
Just noting that another memory issue pops up now. The run does continue for longer than previously; 5 metallicities compared to 3 before this fix. The run fails after approximately 13 hours.
There must be another location where memory is also leaked.

This PR can be accepted if approved to fix the hdf5 input file read leak.

astroJeff (approved):

Issue #208, opened 2023-12-01
Need to extend item sizes for output


celiotine:
When running with the current development branch, I am getting errors that the `ONELINE_MIN_ITEMSIZE` for `event_f` and `mass_transfer_case_f` are not big enough for the output strings in these columns, which is preventing many binaries from being written to the output dataframe. These itemsizes are set on lines 61 & 65 of `binarypopulation.py`. Increasing the size of `event_f` to 25 and `mass_transfer_case_f` to 20 fixes this issue.

celiotine:
Also should extend size of `mass_transfer_case` in `HISTORY_MIN_ITEMSIZ` to 20

ka-rocha:
<img width="1519" alt="Screen Shot 2023-12-11 at 11 52 13 AM" src="https://github.com/POSYDON-code/POSYDON/assets/48293898/2eb45e96-ef25-4845-bc30-fb192a8f469f">

I found a culprit binary, where failing during `step_detached` leaves the binary with an unexpectedly long final event. I suspect the secondary is close to out of the single star grid (?) but did not look further at the error. The orbital period of this systems was ~5600 days. As suggested before, it may be worth having all strings in our output come from a discrete set of names defined in one place for consistency.

celiotine:
Yes @ka-rocha I think the recently added `redirect` events are the strings causing this issue. For example, `redirect_from_CO_HeMS` is 21 characters which is why the previous max length was exceeded. I agree we should discuss renaming things in general and have some procedure to make sure these changes are propagated to the output.

maxbriel:
Since both fail, their last "event" is from the redirect, which doesn't get cleaned in `step_detached`, as mentioned in #218.
A similar thing occurs in `step_SN`, where the state remains `CC1` or `CC2` and the binary fails.

When running a BinaryPopulation, these are caught by a `try/except`, where `binary.event` is rewritten to FAILED.
However, in the `binary.event_history` the last event remains the same as before, ie. `redirect...`,`CC1`, etc.

Even if we fix the writing to file, should we end these binaries like this? Or should we set the state/event of the star/binary to ERR/FAILED?

maxbriel:
The main issue for the shown binary here seems to be that the secondary has a mass outside the grid (with q<0.05), which should be solved in PR #231.

Although it might be worth adding some more informative errors and setting the binary to an FAILED state.

[...]

PR #239, opened 2024-01-12
Add loop checks for other CO grids

maxbriel:
This adds checks for other CO grids and sets the binary state to `"FAILED"` instead of `"ERR"`, which seems to be used in `BinaryPopulation` as well when an error is thrown. Issue  #194

### DIFF
--- a/posydon/binary_evol/DT/step_detached.py
+++ b/posydon/binary_evol/DT/step_detached.py
@@ -1072,7 +1072,7 @@ class detached_step:
                     # ZAMS stars in wide (non-mass exchaging binaries) that are
                     # directed to detached step at birth
                     m0, t0 = star1.mass, 0
-                elif co: 
+                elif co:
                     m0, t0 = copy_prev_m0, copy_prev_t0
                 else:
                     t_before_matching = time.time()
@@ -1094,7 +1094,7 @@ class detached_step:
             else:
                 self.grid = self.grid_strippedHe
             
-            # cehck if m0 is in the grid
+            # check if m0 is in the grid
             if m0 < self.grid.grid_mass.min() or m0 > self.grid.grid_mass.max():
                 binary.state = "ERR"
                 binary.event = "FAILED"
--- a/posydon/binary_evol/DT/step_detached.py
+++ b/posydon/binary_evol/DT/step_detached.py
@@ -32,7 +32,8 @@ from posydon.utils.common_functions import (
     roche_lobe_radius,
     check_state_of_star,
     PchipInterpolator2,
-    convert_metallicity_to_string
+    convert_metallicity_to_string,
+    set_binary_to_failed,
 )
 from posydon.binary_evol.flow_chart import (STAR_STATES_CC, STAR_STATES_CO)
 import posydon.utils.constants as const
@@ -1096,8 +1097,7 @@ class detached_step:
             
             # check if m0 is in the grid
             if m0 < self.grid.grid_mass.min() or m0 > self.grid.grid_mass.max():
-                binary.state = "ERR"
-                binary.event = "FAILED"
+                set_binary_to_failed(binary)
                 raise ValueError(f"The mass {m0} is out of the single star grid range and cannot be matched to a track.")
 
             get_track = self.grid.get
--- a/posydon/binary_evol/DT/step_detached.py
+++ b/posydon/binary_evol/DT/step_detached.py
@@ -1072,7 +1072,7 @@ class detached_step:
                     # ZAMS stars in wide (non-mass exchaging binaries) that are
                     # directed to detached step at birth
                     m0, t0 = star1.mass, 0
-                elif co:
+                elif co: 
                     m0, t0 = copy_prev_m0, copy_prev_t0
                 else:
                     t_before_matching = time.time()
@@ -1081,7 +1081,7 @@ class detached_step:
                     if self.verbose or self.verbose == 1:
                         print("Matching duration: "
                               f"{t_after_matching-t_before_matching:.6g}")
-
+            
             if pd.isna(m0) or pd.isna(t0):
                     #    binary.event = "END"
                     #    binary.state += " (GridMatchingFailed)"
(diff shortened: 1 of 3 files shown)


maxbriel:
I'm not certain in the way I'm changing the `binary.state` and `binary.event` when an error is thrown.
I think changing `binary.state` to `FAILED` and `binary.event` to `ERR`, makes it clear to the user where the binary failed when they look at its history and if these cases show up should stop `event_f` from being larger than 10 characters in the cases discussed in Issue #208.

This also adds 3 extra errors for additional clarity:
1. An error if the mass of the detached_step is outside the grid.  
2. An error if M1 is outside the HMS-HMS grid, but inside the period range.
3. An error if M2 is outside the HMS-HMS grid, but inside the period range.
 
For 2 and 3, I would want to know:
Is there a possibility where M1 or M2 is outside the HMS-HMS grid and the period is inside the grid?
Such a system could potentially go `step_HMS_HMS` -> `step_detached` and just evolve?
This is not considered in the original flow and the added errors stop this from happening in a clearer way, but maybe we want to account for such systems?

mkruckow:
On more thought: Currently, the checks assume a rectangular grid. This isn't the case at low m1 and low q. Should we already think about including a more detailed approach here to detect cases outside the grid?

mkruckow:
I think, we should always set the new state and event. (I'd suggest to clean the code by always having the same order of state and event code lines, which currently is sometimes one way some times the other way, which make is easy to accidentally swap the correct values those two should be set to.

maxbriel:
I'm not sure I understand your point on the rectangular grids.
Do you mean that in the HMS-HMS grid, m2 is forced to be larger than 0.5 in the grid creation and otherwise isn't created?
So depending on m1, q_min and m2_min is different?

This would already be a thing in the standard grid check. I would suggest to get `m2` from the interpolator and check if m1 and/or m2 is outside the those values.

As for the state+event change, I'm suggesting that I put a function in `utils/common_functions`, which sets the binary properties to "FAILED" and "ERR". That should provide the greatest flexibility for usage in multiple spaces (and files).

maxbriel:
During the dev meeting we had some discussion on handling FAILED binaries.

2 things for consideration: 
1. How we handle failed binaries.
2. The detail of error catching.

# 1. Handling failed binaries

Currently, a failed binary does not have a clear indicator in the dataframe/history that it failed. This binary failed, but it's unclear looking at the dataframe if it failed or just hasn't evolved further. Moreover, this leads to the issue with `event_f` being too long for storage in the output file, since these events/states are not meant to be a final state.

Screenshot 2024-01-22 at 15 15 47

I suggest we use a uniform function, such as `set_binary_to_failed` to catch errors that we know about. This way, we know if we have binaries that are behaving outside what we expect. For example, we know that this example binary will fail because `S2_mass` is outside the grid for binary and single stars.

**Do we want to implement this everywhere?**

# 2. Detail of error catching.
This is more specific to the changes in this PR. There are some edge cases that @celiotine and @ka-rocha came across in Issue #208. While this issue was due to the initial sampling having a problem, we should catch in more detail if a binary is outside the grid.

A. star_mass outside the single star grid (addition in `step_detached`)
B. star_mass outside the binary grid (additions in `step_mesa` for each grid type)

The example binary above originally crashes with the following warning. 
Screenshot 2024-01-22 at 15 31 31

This is not informative. With this PR (due to additions A), they will crash like this:
Screenshot 2024-01-22 at 15 33 59

This provides more information for tracking what went wrong in the evolution.

For B. we now get these type of errors:
Screenshot 2024-01-22 at 15 37 12


<details><summary>B example binary</summary>
<p>

```

STAR1 = SingleStar(**{'mass': 13.889634,
                      'state': 'H-rich_Core_H_burning'})
STAR2 = SingleStar(**{'mass':0.490231,
                      'state': 'H-rich_Core_H_burning'})

                    **{'time': 0.0, 'state': 'detached', 'event': 'ZAMS', 'orbital_period':4513.150157, 'eccentricity': 0.0},
                    properties = sim_pop)

```


</p>
</details>

maxbriel:
@mkruckow would you be able to clarify on the rectangular grids assumption?

I believe that using `m2_min` and `m2_max` not `q` would not solve a secondary mass being outside the grid, since the first if-statement to determine if something is in the grid would already capture this binary.
To clarify, q_min =0.05, but for low mass primaries, m2*q >= 0.5 is enforced. Thus, the grid is not rectangular and some values won't be present at low m1?

It also seems to be the case that `m2_min/m1_min` from the interpolator gives `q_min=0.1`, but it's hard coded as 0.05.

mkruckow:
Usually, we have:
m1_min<=m1<=m1_max & q_min<=q<=q_max, where m1_min, m1_max, q_min=0.05 and q_max=1.0 are fixed values (those four values define a rectangular area).
But we use a third limiting criterion: 0.5=m2_min<=m2. Thus we effectively have one three constants: m1_min, m1_max and q_max. q_min is effectively a function of m1: f_q_min(m1)=max(q_min,m2_min/m1). This functional dependence turns the area to become a pentagon.

mkruckow:
As you added the request of p_min<=p, I guess, we should add a case for p<p_min. When the grids are set up well, all below p_min should be initial RLO. Should we mark them as failed or better should we send them to merger step, @ezapartas can the merger routine deal with two stars without well defined cores?

astroJeff:
Update (Jan 25): PR close to completion. @mkruckow and @maxbriel to finish up, resolving previous discussion about how to identify when binaries are outside the grids (grids are not exactly square)

maxbriel:
1. I've added a minimum period redirect, which follows the same direction as having a larger period. This ends up in `step_detached`, where it will be tagged as "initial_RLOF" and should stop. I don't think these systems go to the `step_merged`.
If this should be implemented, then that would be a new PR.

2. I've implemented a mass ratio selection as you suggested, based on $M_1$ and $q_\mathrm{min}$. 

@mkruckow @ezapartas Would you be able to review the changes?

mkruckow:
1. OK, if the detached step can deal with those cases of too low period and don't redirect them to MESA RLO girds. Just for semantics: I'm not sure, that I like to get them classified as "initial_RLOF", when they come from a CO+star gird.
2. You have made the change at one place, please do it at all places, where `m2` or `q` is checked.

maxbriel:
For point 1, there's a slack conversation about this already, where it was noted that it felt a bit confusing.

For 2, I've changed the original code to also check calculate $q_\mathrm{min}$ for the HMS-HMS grids. The CO+star grids don't have this problem, because they're generated from CO and star masses that are always the same, right?

mkruckow (approved):

astroJeff (approved):

PR #241, opened 2024-01-17
Refinements in the `compress-mesa` tool

kkovlakas:
The script was created by @prinse1545 . I made some changes for easier use. E.g.,

`compress-mesa MY_DIR`

instead of

`compress-mesa -f compress_dir -md MY_DIR`.

In addition, a dependency was removed.

### DIFF
--- a/bin/compress-mesa
+++ b/bin/compress-mesa
@@ -110,12 +110,35 @@ def compress_dir(args):
                     total_size += os.path.getsize(filepath)
         return total_size
 
+    def compress_subdirs(track_dirs):
+        for track_obj in track_dirs:
+            for root, _dir, files in os.walk(track_obj):
+                # traversing over directory tree of copied mesa tracks
+                # and compressing .data, .mod, .txt files
+                for file in files:
+                    ext = file.split(".")[-1]
+                    if ext not in ["data", "mod", "txt"]:
+                        continue
+                    os.system(f"gzip -1 {os.path.join(root, file)}")
+
+
     if not os.path.isdir(args.mesa_dir):
         sys.exit("The MESA directory does not exist.")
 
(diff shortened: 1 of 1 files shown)


astroJeff:
I made a couple changes. The motivation is that if I'm generating new grids, I may not want to rerun the compress-mesa script on everything I've already run. Maybe I just want to apply it to the newest grids. This functionality allows for the script to be run on one directory lower in the tree. I've tested that functionality and it works, but I haven't checked the original functionality, when run from one directory higher up.

astroJeff:
I would like to add the functionality to remove and `core.12345` files that result from core dumps after a segmentation fault.

astroJeff:
I just updated the script to also remove core dump files when it goes through our MESA grids compressing data. This is recognized by the string "core.", which I don't think should conflict with anything else (i.e., accidentally delete other files we want to keep).

I ran this on `/projects/b1119/POSYDON_GRIDS/HMS-HMS/v2/2e-01_Zsun/rerun_thermohaline_mixing_grid_random_combined` on Quest, and the directory size was reduced from 46 GB down to ~4 GB:
```
Compressed MESA tracks
Original size 46.1G | Compressed size 3.91G

```

I am still waiting to hear back from NUIT about file permissions, but once that goes through we can have a dry run of this script. If all looks good, then we'll run it.

astroJeff (approved):

astroJeff:
Update (Feb 1): Need to test that post-processing scripts can correctly read in and interpret compressed files. Check with v1 HMS-HMS grids.

astroJeff:
Update (Feb 8): Still need to check with post-processing script for v1 grid

elizabethteng (approved):

PR #250, opened 2024-02-02
correct np.max function usage

maxbriel:
In yesterday's PR, I accidentally used the function notation for `max()` with the `np.max()` function.
I don't know why this wasn't caught in the test I performed.

I've added the correct function call in this PR.


mkruckow (approved):
Oh, correct.
It wasn't registered because the old code is correct code, it simply does something different. The change is instead of calling the function with two arguments we now call it with one. the second argument would usually be interpreted as axis and e.g. allows an int there, which your mathematical expression was interpreted as.

ZepeiX (approved):

Issue #235, opened 2024-01-09
Error with track interpolation


dimsour94:
Had multiple runs with POSYDON within 1 month (using the latest development branch, the latsest ini file and track interpolation on) and in each run I got the same error, which is the following: 

```
Error during breakdown of BinaryStar(initial_RLOF, END, p=0.01, S1=(WD,M=1.21), S2=(stripped_He_non_burning,M=0.16)):
setting an array element with a sequence. The requested array has an inhomogeneous shape after 1 dimensions. The detected shape was (28,) + inhomogeneous part.
ValueError: setting an array element with a sequence. The requested array has an inhomogeneous shape after 1 dimensions. The detected shape was (28,) + inhomogeneous part.

```
**The error always refers to binary_state = initial_RLOF**
Similar errors in the past were related to the fact that some conditions in the ini file were not specified well (not importing a step correctly or if a step is missing in ini file, wrong options etc). @celiotine @ezapartas @kasdaglie 
Thus, starting from default ini.file that works fine, and adding in each run a new change in my ini file, I figured out that this error  pops up only when the **track_interpolation = True and only for the step_CO_HeMS_RLO.** I'll try to solve the issue by myself but ideas are welcome.

ezapartas:
@dimsour94 realized that it may not be only happening for step_CO_HeMS_RLO, but in general in step_mesa with **initial_RLO** with **track_interpolation = True**

```
Error during breakdown of BinaryStar(initial_RLOF, END, p=0.55, S1=(NS,M=1.20),\
 S2=(H-rich_Core_H_burning,M=4.35)):
setting an array element with a sequence. The requested array has an inhomogene\
ous shape after 1 dimensions. The detected shape was (28,) + inhomogeneous part\
```

astroJeff:
Error has been identified in try/except statement: binary_population.py, line 561. 

@dimsour94, @ezapartas, @kkovlakas can you work on this? @celiotine @kasdaglie

dimsour94:
Updates on 1/2/2024:
The issue remains unsolved.
Eirini did some digging and printed out some additional information:   
![image](https://github.com/POSYDON-code/POSYDON/assets/60255936/b9d42d9c-d0f3-4b82-b9fc-09e37bd5f88b)

Me and Manos made sure that NN did not select a failed run by checking a few NN systems in the raw MESA data that are printed with verbose in the pop_syn.out file and they looked fine.

[...]

PR #251, opened 2024-02-06
Inhomogeneous binary array

ezapartas:


ezapartas:
trying to generally solve both inhomogeneity of singlestar array and solve the specific issue was occurring for track_interpolation and intial_RLO #235

maxbriel:
This works for me on the test binary I found for Issue #235. However, I've not tested this in a population run.

astroJeff:
Update (Feb 8): PR appears to address issue #235 dealing with inhomogeneous arrays. It has been tested for individual problematic binaries but needs to be tested on a population. @dimsour94 to run a population and update.

astroJeff:
A question was raised about the use of `["NAN"]` as a placeholder. We should try changing this to `[np.nan]`. @dimsour94 to test this, to see if other issues arise in making this change.

maxbriel (approved):

dimsour94:
Feb 13/02/2024:

["NAN"] changed to [np.nan]+tested in a population run of 10000 binaries with track interpolation and works fine. I think it's ready to be accepted.

dimsour94 (approved):

PR #257, opened 2024-02-21
String len in [step_names] columns

dimsour94:
Running a population with binary fraction less than 1, a value error appears which says cannot store a string with len[20] in step names columns as this column has a limit of [20]. Having a look at the step names that we are using, I noticed that indeed the length of "initially_single_star" has a length of 21, so I increased the length of all [step names] columns by 1.


astroJeff (dismissed):

maxbriel (dismissed):
Good catch!

maxbriel (approved):

astroJeff (approved):

PR #266, opened 2024-03-05
conda_v1 fix

maxbriel:
I changed the requirements to be minimum and changes the`numpy` maximum requirement, which caused the solver to fail.
Locally on my machine, this passed `conda mambabuild`. 

I've changed `setup.py` to follow the same maximum allowed versioning.


maxbriel:
This solves the issue with the publishing to conda. Hopefully, if accepted this should propagate to `conda` and solve the installation issues.

tassos25 (approved):

maxbriel:
Technically this is already publish-to-anaconda due to how the github action is setup.

I've check that on my computer the conda install works.

astroJeff (approved):

PR #271, opened 2024-03-08
Unifying the code for finding grid boundaries and checking validity in case of ignored data.

kkovlakas:
- [x] Simplify the boundary detection code in MESA grids
- [x] Determining if specific runs should be excluded from the calculation of the boundaries.


astroJeff:
Update (Apr 11): Code does not affect logic - just a more elegant solution to finding grid boundaries with fewer lines of code. @kkovlakas to test this with a grid to make sure boundaries are unaffected.

maxbriel (commented):
This is a nice simplification of the code. 
@kkovlakas I was just wondering if a function definition inside a function is a good idea, especially since this is quite a general function. It might be better to define it as another private function of the class. That will also be easier for testing.

Otherwise it looks fine!

astroJeff:
@kkovlakas, have you had a chance to test this PR yet?

kkovlakas:
> This is a nice simplification of the code. @kkovlakas I was just wondering if a function definition inside a function is a good idea, especially since this is quite a general function. It might be better to define it as another private function of the class. That will also be easier for testing.
> 
> Otherwise it looks fine!

I think this function is not going to be used in other methods of the class. So I would keep it hidden to be really "private".

kkovlakas:
> @kkovlakas, have you had a chance to test this PR yet?

Not yet. I was having some issues with Quest, but I am planning to continue again this week.

astroJeff:
Update (Aug 22): This is a very minor update that we should be able to accept with not too much conversation in the future. One check: we want to make sure that in the CO-HMS_RLO grid, we aren't accidentally setting the boundary to models that never overfilled their Roche lobes. Those simulated points are still stored in the grids for bookkeeping, even if they never interacted.

maxbriel:
@kkovlakas asked me to check the grid boundaries between the dev and this PR:

# HMS-HMS

PR271
```m1_min, m1_max, m2_min, m2_max, p_min, p_max
5.5508124513 286.3926178896 0.5550812451 283.5286917107 0.1 5179.4746792312
```
dev
```
5.5508124513 286.3926178896 0.5550812451 283.5286917107 0.1 5179.4746792312
```

# CO_HMS_RLO
PR
```
m1_min, m1_max, m2_min, m2_max, p_min, p_max
0.4528572095827115 242.2905821692 1.0 307.4623904763753 0.09776443350050948 4259.673746605375
```
dev
```
0.4528572095827115 320.9213780446 1.0 307.4623904763753 0.09776443350050948 4259.673746605375
```
# CO_HeMS_RLO
PR
```
m1_min, m1_max, m2_min, m2_max, p_min, p_max
0.4799498483185959 191.91562108852466 1.0 307.4566117375904 0.0012552624375754397 1165.776701178937
```
dev
```
0.4799498483185959 191.9157747907 1.0 307.4566117375904 0.0012552624375754397 1165.776701178937
```

# CO_HeMS
PR
```
m1_min, m1_max, m2_min, m2_max, p_min, p_max
0.5 191.9157747907 1.0 307.4561367244 0.02 1147.2305020897
```

dev
```
0.5 191.9157747907 1.0 307.4561367244 0.02 1147.2305020897
```

mkruckow:
The difference in the m1_max values in the RLO grids between dev and this PR are indeed from runs, where the final values of the mass are `nan`. All of them are systems labeled as `forced_initial_RLO`. I think, those are runs, where MESA does not produce any output and are reconstructed from the grid point, thus having no final values being set. Surely, those runs should not contribute to the interpolation area of a step. (There are more systems with initial RLO which could be excluded from the determination of the interpolation area, but this is an enhancement which might be beyond the scope of this PR)

posydon/binary_evol/MESA/step_mesa.py
```diff
@@ -231,6 +231,24 @@ def __init__(
         self.stop_var_name = stop_var_name
         self.stop_value = stop_value
         self.stop_interpolate = stop_interpolate
+        self._find_boundaries()
+
+    def _find_boundaries(self):
+        """Infer the grid boundaries (min/max of masses and orbital period)."""
+        def initial_values_min_max(parameter_name):
+            # get the request column from the initial values
+            arr = self._psyTrackInterp.grid.initial_values[parameter_name]
```
mkruckow: I'd suggest to give it a better name then `arr`, especially when having `final` in mind.

posydon/binary_evol/MESA/step_mesa.py
```diff
@@ -231,6 +231,24 @@ def __init__(
...
+            # get the final values - NaN for ignored/failed runs
+            final = self._psyTrackInterp.grid.final_values[parameter_name]
+
+            # only use runs for which both initial and final value is defined
+            arr = arr[np.isfinite(arr + final)]
+
+            return np.min(arr), np.max(arr)
+
+        self.m1_min, self.m1_max = initial_values_min_max('star_1_mass')
+        self.m2_min, self.m2_max = initial_values_min_max('star_2_mass')
```
mkruckow, line 250: Beside `m2` should we always calculate a `q_min` and `q_max`? At the moment this is still hard coded as 0.05 and 1.0 for the HMS-HMS grid.

posydon/binary_evol/MESA/step_mesa.py
```diff
@@ -1243,11 +1250,11 @@ def __call__(self, binary):
...
         if (state_1 == 'H-rich_Core_H_burning' and
-            state_2 == 'H-rich_Core_H_burning' and
-            event == 'ZAMS' and
-            self.m1_min <= m1 <= self.m1_max and
-            np.max([self.q_min, 0.5/m1]) <= mass_ratio <= self.q_max and
-            self.p_min <= p <= self.p_max):
+                state_2 == 'H-rich_Core_H_burning' and
+                event == 'ZAMS' and
+                self.m1_min <= m1 <= self.m1_max and
+                np.max([self.q_min, 0.5/m1]) <= mass_ratio <= self.q_max and
```
mkruckow, line 1256: Shouldn't we better use `np.max([self.q_min, self.m2_min/m1])`? The same applies at other places, too.

posydon/binary_evol/MESA/step_mesa.py
astroJeff: I think your suggestion is a good one. However, we will have to build in some checks, since some grids are based on m2_min/m2_max and some grids are based on q_min/q_max.

posydon/binary_evol/MESA/step_mesa.py
astroJeff: Agreed. Since the scope of this variable is limited to this function, I think we can get away calling it `initial`

posydon/binary_evol/MESA/step_mesa.py
```diff
@@ -312,7 +330,6 @@ def __call__(self, binary):
             # binary.event = 'END'
             return
 
-        binary_start_time = binary.time
```
astroJeff, line 315: This is a strange change. Is `binary_start_time` a legacy variable that is no longer needed?

posydon/binary_evol/MESA/step_mesa.py
```diff
@@ -1215,17 +1229,10 @@ def __init__(self, metallicity=1., grid_name=None, *args, **kwargs):
...
-        # If nothing to do, no init necessary
 
-        # load grid boundaries
-        self.m1_min = min(self._psyTrackInterp.grid.initial_values['star_1_mass'])
-        self.m1_max = max(self._psyTrackInterp.grid.initial_values['star_1_mass'])
-        self.m2_min = min(self._psyTrackInterp.grid.initial_values['star_2_mass'])
-        self.m2_max = max(self._psyTrackInterp.grid.initial_values['star_2_mass'])
+        # set mass ratio
         self.q_min = 0.05 # can be computed m2_min/m1_min
         self.q_max = 1. # note that for MESA stability we actually run q_max = 0.99
```
astroJeff, line 1235: These should not be hard-coded, right? They ought to be taken from the grid initial values themselves.

posydon/binary_evol/MESA/step_mesa.py
```diff
@@ -1243,11 +1250,11 @@ def __call__(self, binary):
...
-            state_2 == 'H-rich_Core_H_burning' and
-            event == 'ZAMS' and
-            self.m1_min <= m1 <= self.m1_max and
-            np.max([self.q_min, 0.5/m1]) <= mass_ratio <= self.q_max and
-            self.p_min <= p <= self.p_max):
+                state_2 == 'H-rich_Core_H_burning' and
+                event == 'ZAMS' and
+                self.m1_min <= m1 <= self.m1_max and
+                np.max([self.q_min, 0.5/m1]) <= mass_ratio <= self.q_max and
+                self.p_min <= p <= self.p_max):
```
astroJeff, line 1257: Rather than construct this series of if statements, can we make a function that checks if all the conditions are correct for a binary to fall within a MESA grid?

posydon/binary_evol/MESA/step_mesa.py
mkruckow: Yes, that's the only occurrence of this variable in the code.

posydon/binary_evol/MESA/step_mesa.py
mkruckow: They differ for different grids. E.g. check `m2` vs check `q`.

posydon/binary_evol/MESA/step_mesa.py
kkovlakas: OK, I made changes to the names (also introducing another varibale `initial_values_to_use` to avoid unsafe references).

posydon/binary_evol/MESA/step_mesa.py
kkovlakas: I believe that if this is to be done, it should be a different PR, and employ the change everywhere in the code in a consistent way (even in the submission script).

posydon/binary_evol/MESA/step_mesa.py
kkovlakas: Feel free to open an issue on this. The change here is in the indentation, so it's better to not tie it on this PR.

posydon/binary_evol/MESA/step_mesa.py
kkovlakas: As in the above comment - I believe this could be an `enhancement` issue if you think there is a way to do it more automatically. Here, I modified the indentation only.

posydon/binary_evol/MESA/step_mesa.py
kkovlakas: @astroJeff I think @mkruckow 's comment below is relevant to that.

mkruckow (approved):
For further improvements I opened issue #565 .

sgossage (approved):

PR #237, opened 2024-01-10
Fixes in postprocessing

mkruckow:
This PR should fix most of open errors and warnings appearing in the post-processing.

- [x] 1) Check for `None`s before requesting dtypes from an object in the visualization [`plot1D.py`, `plot2D.py`], see issue #236 
- [x] 2) In case of start at RLO turn warning of no remaining history into a verbose output [`psygrid.py`]. When we keep track of non MT systems there this isn't worth a warning anymore.
- [x] 3) Remove warning about no RLO when cutting the histories. [`scrubbing.py`]
- [x] 4) Add a check, whether a star can collapse before trying to collapse it in binary girds with no or stable MT. [`post_processing.py`] This is needed, when we have low mass stars like in the CO-HMS grid, which stop at max_age.
- [x] 5) Adopt change of return in `do_core_collapse_BH`. [`profile_collapse.py`]
- [x] 6) Replace MT cases, if an earlier one would happen after a later one. [`termination_flags.py`, `common_functions.py`]
- [x] 7) Renaming of pipeline scripts to start with posydon. [`bin/posydon-run-pipeline`, `bin/posydon-setup-pipeline`, `pipeline_steps.rst`, `pipeline_ini.rst`, `just_step_1.ipynb`, `run_full_piepeline.ipynb`, `step_rerun.ipynb`] -> **Those changes require to reinstall posydon via** `pip install -e .`**. Additionally, old pipeline scripts need to be adopted or recreated in case to rerun them.**
- [x] 8) New ability to add unknown keys to ensure plotting [`plot2D.py`, `plot_defaults.py`], see issue #209
- [x] 9) Add cleanup jobs [`posydon-setup-pipeline`, `pipeline_ini.rst`, `pipeline_quest.ini`, `pipeline_yggdrasil.ini`]
- [x] 10) Don't use replace to detect a format specifier of a number in the title to be replaced by `all`, because it won't work in case of a value with a unit behind. Thus using a split on the equal sign instead. [`posydon-run-pipeline`]
- [x] 11) Put all checks in run-pipeline to come from flags in the csv files created in setup-pipeline [`posydon-run-pipeline`, `posydon-setup-pipeline`, `pipeline_steps.rst`, `pipeline_additions.rst`]
  - step_1
  - step_3
  - step_4
  - step_R
  - plots
[...]

PR #273, opened 2024-03-08
Stable reverse mass transfer cases - updated interpolation objects

astroJeff:
Re-opening this PR from before. Previously accepted as PR #170, but it was not ready for accepting due to incomplete V2 grids. Interpolation objects need to be re-trained before we can accept this PR.


mkruckow:
It is the other way around, the current dataset was already trained with Zepei's changes. Those are now back in development with PR #237 . Hence, I'd close this PR as out of date.

PR #289, opened 2024-04-15
Generalize and improve on compress MESA

mkruckow:
- [x] Reduce the number of filesystem walks from 4 to 2, hence it is a bit faster now
- [x] Generalize the code to be able to be run on any directory level above the MESA runs
- [x] Ensure that only files in the MESA runs are changed without assuming a certain file structure (e.g. not only exclude files in `star1`, `star2` and `binary` instead only include files in a MESA run file)
- [x] Give more statistical output: number of runs, number of changed files and number of their base directories (the last number is the one shown on the progress bar; the first two are for checks for the user, whether it acts on what is expected)
- [x] Final check, that there are no files to act on after the file actions are completed
- [x] Update documentation
- [x] Added by @astroJeff : catch fails on removals, e.g. because of missing permissions
- [x] Testing, here the new output format from the test:
> remove 0 core dump files in 0 directories of 803 MESA runs
0it [00:00, ?it/s]
compress 7194 files in 2026 directories of 803 MESA runs
100%|███████████████████████████████████| 2026/2026 [20:14<00:00,  1.67it/s]
>
> Compressed MESA tracks
Original size 41.3G | Compressed size 10.8G

### DIFF
--- a/bin/compress-mesa
+++ b/bin/compress-mesa
@@ -102,72 +102,75 @@ def compress_dir(args):
     """
     def get_size(start_path="."):
         total_size = 0
+        remove_files = []
+        compress_files = []
+        n_runs = 0
+        n_remove_files = 0
+        n_compress_files = 0
         for dirpath, _, filenames in os.walk(start_path):
+            if "_grid_index_" in dirpath:   # checking if directory is mesa run
+                new_remove_files = []
+                new_compress_files = []
+                if "_grid_index_" in os.path.basename(dirpath):
+                    n_runs += 1
+            else:
+                new_remove_files = None
+                new_compress_files = None
             for filename in filenames:
                 filepath = os.path.join(dirpath, filename)
                 # skip if it is symbolic link
                 if not os.path.islink(filepath):
                     total_size += os.path.getsize(filepath)
-        return total_size
-
-    def compress_subdirs(track_dirs):
-        for track_obj in track_dirs:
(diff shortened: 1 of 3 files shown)


sgossage:
A minor QOL thing to add might be to list the full paths of the newly created compressed files, to make it clear to new users where things are stored.

mkruckow:
> A minor QOL thing to add might be to list the full paths of the newly created compressed files, to make it clear to new users where things are stored.

No file gets moved or joint with another file. They are compressed individually and in place, e.g. `out.txt` becomes `out.txt.gz` at the same location. Hence, I can add a statement like "in place" and/or "by adding extension .gz". But I think, this might be more something to add to the documentation (which I have done now) instead of the screen output.

sgossage (dismissed):

astroJeff:
Update (Apr 25): Waiting on @astroJeff to test.

astroJeff:
Update (May 9): @astroJeff to run test on Quest

sgossage (dismissed):

mkruckow:
Adding a try around the removing is a good idea as an additional improvement to the old code.

astroJeff:
Running script on Quest in `/projects/b1119/POSYDON_GRIDS_v2` right now. Everything looks good and seems to be working well

astroJeff (approved):

sgossage (approved):

PR #281, opened 2024-03-19
More fixes in post-processing pipeline

mkruckow:
Parts done/to do by @mkruckow:
- [x] Allow to search for a second termination code in case of `min_timestep_limit`, to over rule `min_timestep_limit` (see issue #254 )
- [x] There is now a general framework to add values from the histories at He depletion to the `final_values`.
  - `Error during MODEL05 Patton&Sukhbold20-engineN20 core collapse prescrition!  The history did not contain core masses at He depletion!` (seen in step_3 of CO-HeMS_RLO; see question on [slack #running_grids](https://posydon.slack.com/archives/C01LXP17BH7/p1708530484623489); c.f. issue #260 )
- [x] Take care of "unknown" termination flags
  - Some "unknown" resulted from the new interpolation class `stable_reverse_MT`, which were not recognized as `stable_MT` (see issue #284 )
  - Some "unknown" have been caused by having termination flag 2 stating initial MT, while termination flag 1 doesn't
  - Some "unknown" termination flags have been caused by using different thresholds, hence all limits and thresholds got spin off into `utils/limits_thresholds.py`
- [x] Use `np.nan` for lambda, where it is not well defined and set Ebind to 0 if lambda is not well defined. (see issue #279 ) Additionally, correct doc-string for `CEE_parameters_from_core_abundance_thresholds` and energy check.
- [x] Add more options in the pipeline ini files:
  - Add global switches to turn off all plots/checks.
  - Add cluster name to rerun section to be no longer inferred from the path.
- [x] Combining pipeline plots into image files with several pages (see issue #278 )
  - Allow the user to specify a plot file extension (make pdf new default)
  - Make pdfs multi-page plots if the extension is `multipage-pdf`.
- [x] Adding profile interpolation to the pipeline (see issue #213 )
  - It need some more refinement to get the profile interpolation working for all girds, but it is now available in the pipeline
- [x] Updating doc-strings and Errors in the pipeline
- Update tutorial in the documentation (may be skipped here because it is unclear what base data set to use for an independent user.)

Parts on the interpolators, hence feedback from @philipp-rajah needed:
- [x] see PR #295 :
  - `UserWarning: y_pred contains classes not in y_true` (seen in step_4 of CO-HeMS grid)
    - Not understood yet and looks to happen inside sklearn, hence may not be solved.
  - `ValueError: All-NaN slice encountered` (seen in step_4 of CO-HeMS_RLO)
    - Has a dirty workaround since PR #237. Is there a better solution without a `try`?
[...]

PR #295, opened 2024-04-26
Requested fixes on the interpolation (to complete PR #281)

mkruckow:
- `UserWarning: y_pred contains classes not in y_true` (seen in step_4 of CO-HeMS grid)
  - not understood yet and looks to happen inside sklearn, hence may not be solvable
- [x] `ValueError: All-NaN slice encountered` (seen in step_4 of CO-HeMS_RLO) -> has a dirty workaround since PR #237
  - Is there a better solution without a `try`?
- [x] `UserWarning: 1NN interpolation used for N binaries out of hull.` popping up too often (seen in step_4_plotting of HMS-HMS and CO-HMS gird)
  - the warning should get extended by telling how far the most extreme one is outside the hull (if this output happens in normalized values, it needs to state the scaling parameters)
- `RuntimeWarning: overflow encountered in divide` (seen in step_4_plotting of CO-HMS gird)
  - unclear how to solve, hence may not be solved (playing with variable types already tried by @mkruckow )
- [x] `UserWarning: skip normalize: c=initial_MT, inds=[...]` (seen in step_4_plotting of CO-HMS gird)
  - this warning is removed hence the skipping will happen in a silent way


posydon/interpolation/IF_interpolation.py
```diff
@@ -963,15 +963,15 @@ def scale_one(in_scaling, klass = None):
                             np.abs(ypred_i[:, which_abs] - Y_t[:, which_abs]), 90,
                             axis=0)
 
-            try:
-                where_min = np.nanargmin(np.nanmean(err, axis=2), axis=0)
-            except:
-                # replace nans by a large number before looking for minimum
-                where_min = np.nanargmin(np.nan_to_num(np.nanmean(err, axis=2), nan=1e99), axis=0)
+
+            no_nan_slices = ~np.isnan(np.nanmean(err, axis = 2)).all(axis = 0)
```
mkruckow: Once we said, we should better use `pandas.isna` instead of `numpy.isnan`, could you please check, whether we should make this change here?
Edit: Additionally, `pandas` has a function `notna`, which directly returns the inverse of `isna`. On the other hand `numpy` has the function `isfinite`, would that be appropriate here in case of staying with `numpy`?

posydon/interpolation/IF_interpolation.py
```diff
@@ -963,15 +963,15 @@ def scale_one(in_scaling, klass = None):
...
+
+            where_min = np.full_like(np.zeros(shape = (self.n_out, )), fill_value = np.nan)
```
mkruckow: This looks a bit more complicated than needed. Shouldn't it be equivalent to the simpler and more efficient version:
`where_min = np.full(shape = self.n_out, fill_value = np.nan)`?

posydon/interpolation/IF_interpolation.py
```diff
@@ -1140,8 +1140,10 @@ def predict(self, Xt):
         if np.any(wnan):
             Ypred[wnan, :] = self.interpolator[1].predict(Xt[wnan, :])
         if np.any(wnan):
+            dists, _ = self.interpolator[1].kneighbors(Xt[wnan, :])
```
mkruckow, line 1145: Wouldn't it be saver to specify the number of neighbors here, just to ensure that we are independent from the global settings? Otherwise we may check for too many neighbors, which are further away and will dominate the maximum.
Additionally, I'm not sure whether this works as we'd need it: we are interested in the distance on each axis individually because they are probably scaled independently.
Finally, this only gives the distance to the nearest neighbor. Is it guaranteed, that this is part of the hull, e.g. if this is an island of 1 system?

PR #296, opened 2024-04-26
Add user modules

mkruckow:
For population synthesis we have the option to specify modules and functions which should get loaded dynamically. Unfortunately, this functionality is restricted to installed python packages. Hence, I introduce a new directory `user_modules` in `posydon` (which is installed as a name space when doing the POSYDON installation). This I aim to be the desired place for POSYDON users to put their own modules, which are then automatically part of the posydon name space within their installation. In this way they can be easily and dynamically imported wherever needed while they don't mix up with the modules provides by us, the POSYDON developers.

Additionally, I added an example I developed for a current use case of mine and shows, how one can modify the flow chart of POSYDON. It is used by simply modifying the import in the `population_params.ini` to `import = ['posydon.user_modules.my_flow_chart_example', 'modified_flow_chart']`


astroJeff:
When adding the flat mass ratio, without a minimum q, option, I suggest rather than have a completely separate option for the user to choose from (`flat_mass_ratio_all`), I propose adding a `q_min` limit that the user can specify.

celiotine (dismissed):

ka-rocha (dismissed):

ka-rocha (approved):

celiotine (approved):

PR #299, opened 2024-05-07
Add functions to read initial values from file

mkruckow:
Add the functionality to be able to read the initial binary values from a given file.

- [x] add a field to the `population_params_default.ini` to take a file name, if empty data will be randomly generated as before
- [x] add function to read the file
  - try to infer the name a value may have in the file otherwise take a random sample of the same length as the file provides
  - extend the population in the file by doubling it, if needed
  - read kick values additionally to initial values
- [x] sort POSYDON errors and add description of python standard ones
- [x] tested with `ZAMS_values.csv`, see below

`ZAMS_values.csv` content:
> M_1,M_2,P_orb,e,w1,phi1,theta1,s1_natal_kick_array_3
> 10.0,10.0,10.0,0.0,0.0,0.0,0.0,0.0
> 10.0,10.0,100.0,0.0,1.0,0.0,0.0,0.0
> 10.0,10.0,1000.0,0.0,10.0,0.0,0.0,0.0
> 20.0,10.0,10.0,0.0,10.0,0.785398163,0.0,0.0
> 20.0,10.0,100.0,0.0,10.0,1.570796327,0.0,0.0
> 20.0,10.0,1000.0,0.0,10.0,2.35619449,0.0,0.0
> 40.0,10.0,10.0,0.0,10.0,3.141592654,0.0,0.0
> 40.0,10.0,100.0,0.0,10.0,0.785398163,0.785398163,0.0
> 40.0,10.0,1000.0,0.0,10.0,1.570796327,0.785398163,0.0
> 20.0,20.0,10.0,0.0,10.0,2.35619449,0.785398163,0.0
> 20.0,20.0,100.0,0.0,10.0,3.141592654,0.785398163,0.0
> 20.0,20.0,1000.0,0.0,10.0,0.785398163,0.785398163,0.785398163
> 40.0,20.0,10.0,0.0,10.0,1.570796327,0.785398163,0.785398163
> 40.0,20.0,100.0,0.0,10.0,2.35619449,0.785398163,0.785398163
> 40.0,20.0,1000.0,0.0,10.0,3.141592654,0.785398163,0.785398163
> 40.0,40.0,10.0,0.0,10.0,4.712388981,0.785398163,0.785398163
> 40.0,40.0,100.0,0.0,10.0,4.712388981,0.785398163,1.570796327
> 40.0,40.0,1000.0,0.0,10.0,4.712388981,0.785398163,4.712388981

### DIFF
--- a/posydon/popsyn/sample_from_file.py
+++ b/posydon/popsyn/sample_from_file.py
@@ -2,7 +2,7 @@
 
 
 __authors__ = [
-    "Matthias Kruckow <matthias.kruckow@unige.ch>",
+    "Matthias Kruckow <Matthias.Kruckow@unige.ch>",
 ]
 
 
@@ -10,7 +10,6 @@ import os
 import numpy as np
 import pandas as pd
 import warnings
-from posydon.utils.posydonerror import FileError
 from posydon.popsyn.independent_sample import (generate_orbital_periods,
                                                generate_orbital_separations,
                                                generate_eccentricities,
@@ -92,7 +91,7 @@ def get_samples_from_file(orbital_scheme='', **kwargs):
     if os.path.isfile(filename):
         df = pd.read_csv(filename)
     else:
-        raise FileError(f'{filename} not found!')
+        raise FileNotFoundError(f'{filename} not found!')
 
     # Get number of data frame entries
     set_n = len(df)
@@ -202,7 +201,7 @@ def get_kick_samples_from_file(**kwargs):
     if os.path.isfile(filename):
         df = pd.read_csv(filename)
     else:
-        raise FileError(f'{filename} not found!')
+        raise FileNotFoundError(f'{filename} not found!')
 
     # Get number of data frame entries
     set_n = len(df)
--- a/posydon/popsyn/sample_from_file.py
+++ b/posydon/popsyn/sample_from_file.py
@@ -23,6 +23,18 @@ PERIOD_NAMES = ['orbital_period', 'period', 'p_orb', 'porb', 'p']
 SEPARATION_NAMES = ['orbital_separation', 'separation', 'semi-major_axis',\
                     'semi_major_axis', 'a']
 ECCENTRICITY_NAMES = ['eccentricity', 'ecc', 'e']
+PRIMARY_KICK_VELOCITY_NAMES = ['s1_natal_kick_array_0', 'w_1', 'w1']
+SECONDARY_KICK_VELOCITY_NAMES = ['s2_natal_kick_array_0', 'w_2', 'w2']
+PRIMARY_KICK_AZIMUTHAL_ANGLE_NAMES = ['s1_natal_kick_array_1', 'phi_1', 'phi1']
+SECONDARY_KICK_AZIMUTHAL_ANGLE_NAMES = ['s2_natal_kick_array_1', 'phi_2',\
+                                        'phi2']
+PRIMARY_KICK_POLAR_ANGLE_NAMES = ['s1_natal_kick_array_2', 'theta_1', 'theta1']
+SECONDARY_KICK_POLAR_ANGLE_NAMES = ['s2_natal_kick_array_2', 'theta_2',\
+                                    'theta2']
+PRIMARY_KICK_MEAN_ANOMALY_NAMES = ['s1_natal_kick_array_3', 'mean_anomaly_1',\
+                                   'mean_anomaly1']
+SECONDARY_KICK_MEAN_ANOMALY_NAMES = ['s2_natal_kick_array_3',\
+                                     'mean_anomaly_2', 'mean_anomaly2']
 
 def infer_key(available_keys=[], allowed_keys=[]):
     """Infer key from list of allowed keys.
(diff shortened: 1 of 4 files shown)


posydon/popsyn/sample_from_file.py
```diff
@@ -0,0 +1,168 @@
...
+    if 'read_samples_from_file' in kwargs:
+        filename = kwargs['read_samples_from_file']
+    else:
+        raise KeyError("In get_samples_from_file no 'read_samples_from_file'"
+                       " in kwargs.")
+    # Check and read file into dataframe
+    if os.path.isfile(filename):
+        df = pd.read_csv(filename)
+    else:
+        raise FileError(f'{filename} not found!')
```
celiotine: Do we want this to be a POSYDON error? (If the file does not exist should it be automatically thrown and not suppressed with other POSYDON errors?)

mkruckow: Hm, I had the impression we'd like to capture all kinds of errors as much as possible by ourselves.
I may have a different view here than others, but for me an error is a critical issue, which requires the code the stop/not continue there, otherwise it would be a warning. Hence, errors shouldn't get suppressed unless one knows that it won't hurt to let the code crash, but usually it would be better to instead not execute the code part if it isn't needed to run properly through.

celiotine: Right, so I think with the POSYDON Errors, we are currently using them to capture errors in the code that are known but not fixable, e.g. a binary could not be matched, or there are limits with numerical methods. Any other errors that occur are not categorized as POSYDON Errors, just regular python errors, and they are always thrown. So, the default currently with the population runs is that we suppress POSYDON Errors and everything else is thrown automatically.

celiotine: So, for the binary populations, the POSYDON Errors indicated a binary failed, but there is nothing we can do about it for now. If we want this behavior to be different for other parts of the code, that is also possible.

astroJeff:
To clean up the input string options, can we use either `.lower()` or `.casefold()`, see SO link: https://stackoverflow.com/questions/319426/how-do-i-do-a-case-insensitive-string-comparison

mkruckow:
> To clean up the input string options, can we use either `.lower()` or `.casefold()`, see SO link: https://stackoverflow.com/questions/319426/how-do-i-do-a-case-insensitive-string-comparison

I have made the change to use those functions to shorten the lists. Should we get an own function, which removes dashes, underscores and spaces to shorten it even more?

posydon/popsyn/sample_from_file.py
mkruckow: I have removed the FileError and replaced it by the in-build FileNotFoundError. Additionally, I added a list of in-build Error classes to the bottom of our POSYDONErrors, to give a complete overview of available errors there to allow us to always pick the most appropriate error.

posydon/utils/posydonerror.py
```diff
@@ -1,5 +1,13 @@
 """The POSYDON exception class and subclasses for more specific errors."""
 
+
+__authors__ = [
+    "Camille Liotine <cliotine@u.northwestern.edu>",
```
celiotine, line 5: Add Eirini as a primary author, as she did most of this code development with me.

maxbriel:
A bit late to this PR, but this functionality is already there in the `binarypopulation.py`, where you can read and evolve binaries from an hdf5 file. That function is **not** optimised for reading the binaries from the file, but works correctly in evolving the binaries using custom initial ZAMS value.

I'm not sure what the added benefit of two frameworks for the same thing is. Is it possible to combine the two in some way or is there additional functionality here that is missing in the other?

mkruckow:
> A bit late to this PR, but this functionality is already there in the `binarypopulation.py`, where you can read and evolve binaries from an hdf5 file. That function is **not** optimised for reading the binaries from the file, but works correctly in evolving the binaries using custom initial ZAMS value.
> 
> I'm not sure what the added benefit of two frameworks for the same thing is. Is it possible to combine the two in some way or is there additional functionality here that is missing in the other?

The `from_hdf` function is aimed to read in from an existing population file. Usually, you'd even use it to read in the evolution not just ZAMS values. Any hdf5 file read by this function needs to be formatted in the POSYDON way, hence having all columns named as POSYDON would name them. Additionally, I'm not sure whether it will work well, if the load is specified in the `population_params.ini` (surely the default version doesn't include value, which you'd need to change). Using the `from_hdf` at least requires a two step process from a user.

So, additional functionality here is:

- read from a simpler file
- flexibly detect column names
- easy to use in `population_params.ini`
- it has a fallback to the POSYDON functions for non given values
- it allows to take less or more binaries than specified (this allows to get a subpopulation or simulate the same binary but with different kicks several times with and without setting them externally)

Additional functionality of `from_hdf`:

- read any amount of the evolution (more than initial values)

kasdaglie (approved):

mkruckow:
@maxbriel, maybe we can add the option to switch to the `from_hdf` in case the given file has an `h5` file extension in an new PR?

astroJeff (approved):

PR #300, opened 2024-05-07
Add POSYDON Warnings

mkruckow:
- [x] New `POSYDONWarning` class
  - [x] add more sub classes
- [x] New `Pwarn` function to easily create POSYDON warnings. It works like the `warnings.warn`, but ensures to have a POSYDONwarning class or a derivative of it as category, which can be given as class or as string containing the class name.
- [x] filter functions `SetPOSYDONWarnings` (a general function, which should only be used in case the latter ones are not applicable), `NoPOSYDONWarnings`, `AllPOSYDONWarnings`
- [x] get a summary option
- [x] get own content manager to catch warnings
- [x] no memory leak in POSYDON warning, see issue #364


posydon/utils/posydonwarning.py
```diff
@@ -17,6 +19,11 @@ class ReplaceValueWarning(POSYDONWarning):
     def __init__(self):
         warning.filterwarnings("default", category=ReplaceValueWarning)
 
+class InterpolationWarning(POSYDONWarning):
+    """Warning that a value got replaced."""
```
mkruckow, line 97: @mpgalleg , please update the docstring.

posydon/binary_evol/binarystar.py
```diff
@@ -224,10 +224,7 @@ def run_step(self):
             next_step_name = self.properties.flow.get(total_state)
 
             if next_step_name is None:
-                warnings.warn("Undefined next step given stars/binary states "
-                              "{}.".format(total_state))
-                self.event = 'END'
```
mkruckow, line 231: I'd suggest to leave the setting of the event here, just to make sure we have a well defines state in case the error gets suppressed.

posydon/utils/posydonwarning.py
```diff
@@ -2,9 +2,9 @@
 
 
 __authors__ = [
-    "Matthias Kruckow <Matthias.Kruckow@unige.ch>",
     "Monica Gallegos-Garcia <monicagallegosgarcia@u.northwestern.edu>",
     "Camille Liotine <cliotine@u.northwestern.edu>"
```
mkruckow: here is a comma missing.

celiotine:
In my most recent commit #[70e05fc](https://github.com/POSYDON-code/POSYDON/pull/300/commits/70e05fc2d3b672c97c151e32d07a7823d0d88500), I've adjusted the warning filter settings in binarypopulation. To ensure that the warnings filters are applied consistently at runtime, I think we should set the filter settings in the sections of code where we want to control the filtering to standard output/error. If we set these filters in the warnings class, the application of the filters is inconsistent, and I found it was not working in general. 

In my changes, I added a BinaryPopulation kwarg "warnings_verbose." Similar to the error checking, if "warnings_verbose=True", all POSYDON warnings will be printed to standard output at runtime with the "default" filter setting (using the "always" filter sometimes prints many copies of the exact same warning for one binary depending on its evolution). If "warnings_verbose=False", all POSYDON warnings are ignored and not written to standard output/error. This is the default setting.

Now, previously, all warnings that occurred when evolving a binary population were suppressed from being written to output/error in this section of the code. Instead, they were caught and added to a list, which was then added as an attribute of the binary (binary.warning_message). It could still be useful to keep this feature, but I will need to think about a way to do this and also be able to print warnings to standard output. If you record the warnings when you catch them, it suppresses them being written to output/error, which is why no warnings were showing-up previously when running a binary population.

The remaining changes I made in this commit catch the occurrence of a few python divide-by-zero RunTime Warnings. I added a check for this behavior and a corresponding POSYDON Warning that I think is more informative, and now these warnings will be included in the general POSYDON Warning filtering. I think code could be re-written in general so that these divide-by-zero warnings do not happen (it seems avoidable to me), but that is outside the scope of this PR.
Monica and I are planning to go through and change any python warnings to POSYDON warnings, if appropriate, similar to what we did for the error checking. This will ensure that when developers add new code, they can see if new python warnings arise from their changes.

posydon/binary_evol/DT/step_detached.py
```diff
@@ -1217,7 +1217,7 @@ def ev_rlo1(t, y):
             if (interp1d_sec["mass"](t - t_offset_sec) == 0) or (interp1d_pri["mass"](t - t_offset_sec) == 0):
                 warnings.warn("Trying to compute difference between stellar radii for non-existent companion",
                             EvolutionWarning)
-                RL = np.nan
+                RL = np.ones_like(interp1d_sec["mass"](t - t_offset_sec))*np.nan
```
mkruckow: What is the attempt here? Conserving the shape of the interpolator output? If so, why the one of the secondary and not the one of the primary?
I think, we shouldn't infer a shape here, the doc string says, we are returning a float not a numpy array.
So instead we should better check that the interpolator outputs are floats otherwise we should raise an error.
Additionally, I don't know, why we check for 0 here, we should disallow any mass <=0.
(This applies to the other three similar cases, too.)

P.S.: This is a bad use of np.ones_like(array)! Better use np.fill_like(array, fill_value). This does the same as you do with the multiplication, but in one step instead of two. Btw. np.fill_like(array, 1.0) is the same as np.ones_like(array) in the background code; hence I don't see any good use case for np.ones_like. My guess is that np.ones_like might existed first with a different implementation, but these days I'd call it outdated.

celiotine: I was finding that simply setting RL = np.nan was causing problems because while the function says it returns a float, in reality it is returning an odd data structure containing one float value. I assumed that the primary and secondary interpolator output would have the same shape but maybe this is not a good assumption? I was finding that the cases where this was failing were happening when the masses were equal to zero, but I guess in theory they could also be less than zero.
I did not know np.fill_like() existed. I can change it to this instead of np.ones_like(). I think calling this a "bad use" of np.ones_like() is a bit unfair given that it does the same thing as what you suggested, with essentially no difference in functionality!
I will look at this again today anyway, something still isn't right with how things are being fixed, which could be related to any number of the things you brought-up.

celiotine: Screen Shot 2024-07-10 at 10 01 01 AM
@mkruckow I printed-out the data types of the RL radius being calculated in the original code (not my changes) and you can see that sometimes it has a type of numpy float64 and other times it is an ndarray (presumably because the type of the interpolator output is changing as I mentioned above). If you think this is a bigger issue, I can just leave this warning alone for now and revert back to the original code.

mkruckow: Hm, if we want to support numpy arrays, we should update the doc string, ensure that all the arrays have the same shape, and switch to use `np.full_like`.
We need to update the check for masses, too. Because this might not work, in case there is a 0 inside the numpy array. Additionally, we should exclude negative masses in this go, too.
P.S.: Usually for better code performance I'd call the interpolators only once at the beginning and save the values and do all the checks and later calculations with them.

maxbriel: @celiotine Do you know why it returns a numpy array?
Is it just a single value inside an array? Or are there multiple values?

mkruckow:
With the last commit, I added a new powerful function `Pwarn` short for POSYDONwarn, which makes the call of `warnings.warn`, but beforehand it checks the category to be a POSYDON warning. In case the category is given as a string but not a POSYDON warning it defaults to our base POSYDONWarning class.

Additionally, it helps to shorten coding:
```
import warnings
from posydon.utils.posydonwarning import (InappropriateValueWarning,
                                          UnsupportedModelWarning)
warnings.warn("Test is bad.", InappropriateValueWarning)
warnings.warn("No test model.", UnsupportedModelWarning)
```
**can now be written as**

```
from posydon.utils.posydonwarning import Pwarn
Pwarn("Test is bad.", "InappropriateValueWarning")
Pwarn("No test model.", "UnsupportedModelWarning")
```
There is no need anymore to import `warnings` and `posydonwarning`, there is no need anymore to import all POSYDONwarning classes separately. And the function call is shorter.
**One important thing: using this requires quotation marks around the category to turn it into a string.**
FYI: The old version still works and you can although use this:
```
from posydon.utils.posydonwarning import (Pwarn,
                                          InappropriateValueWarning,
                                          UnsupportedModelWarning)
Pwarn("Test is bad.", InappropriateValueWarning)
Pwarn("No test model.", UnsupportedModelWarning)
```

posydon/binary_evol/DT/step_detached.py
celiotine: I do not know why the interpolator is returning a numpy array for some values. From what I have checked, it contains one float value but has an odd shape?

celiotine: BTW I don't really have an interest in rewriting this function, I was just trying to turn a divide-by-zero warning into a POSYDON warning. If there are bigger issues here, I think that should be in a different PR

maxbriel: No worries! I agree with not rewritting the function as part of this PR :) 
Maybe make is a separate issue instead?

As for this code, does the `roche_lobe_radius` function crash if a non-existent companion calculation is done or is it just giving the divide-by-zero warning? If the latter, you could just "suppress" the divide-by-zero warning for this part of the code, since you're throwing the POSYDON error, and still call the `roche_lobe_radius` function. Then you do not have to deal with proper handling of the Roche Lobe function input/output ;)  

```
with np.errstate(divide='ignore'):
    RL = roche_lobe_radius(....)
```

or if it's 0/0:

```
with np.errstate(divide='ignore', invalid='ignore'):
    RL = roche_lobe_radius(....)
```

celiotine: Yes that is what I was thinking! The code doesn't crash, it just throws the python warning. I might be able to suppress the warning locally as an easy fix, and create an issue for the interpolator output/this function in general. Thanks!!

posydon/utils/posydonwarning.py
```diff
@@ -0,0 +1,187 @@
...
+            A string, which will be converted into a POSYDONWarning class.
+        **kwargs : dict (optional)
+            Dictionary containing extra options passed to
+            warnings.filterwarnings.
+    """
+    category = _get_POSYDONWarning_class(category)
+    if isinstance(category(), POSYDONWarning):
+        warnings.filterwarnings(action=action, category=category, **kwargs)
+
+def NoPOSYDONWarnings(category=POSYDONWarning):
```
celiotine, line 570: I am curious if these two functions (NoPOSYDONWarnings, AllPOSYDONWarnings) are needed. To me it is more explicit to the user if we just use the SetPOSYDONWarnings function, then they can see the see the filter type.

posydon/binary_evol/SN/step_SN.py
```diff
@@ -589,42 +585,13 @@ def collapse_star(self, star):
...
-                                star.SN_type = 'CCSN'
-                            elif star.state == 'massless_remnant':
-                                star.SN_type = 'PISN'
-                            else:
-                                raise ValueError('Star state not recognized.')
-                    
+                                      'does not match the predicted CO, or the interpolated '
+                                      'values for the SN remnant are NaN. '
+                                      'If use_profiles or use_core_masses is set to True, '
+                                      'continue with the collapse.', "ApproximationWarning")
```
mkruckow, line 593: I'm not sure, that we can safely remove this code part.
It was aimed to correct issues in the case `use_profiles` and `use_core_masses` are `False`, which should be the standard for pop-syn. You need to keep in mind, that putting them to `True`, will always overwrite/replace the values gotten from the interpolator.
So I see two options:

1. Check whether those values are `True`
   1. In case one of them is, throwing a warning is OK.
   2. In case they are both `False`, we need to throw an error or see 2..
2. We need code to deal with the case of inappropriate values coming from the interpolator.

posydon/popsyn/binarypopulation.py
```diff
@@ -318,12 +318,12 @@ def _safe_evolve(self, **kwargs):
...
-            with warnings.catch_warnings(record=False):
-        
+            with warnings.catch_warnings(record=True) as w:
+                       
                 if self.kwargs.get("warnings_verbose", False):   
-                    warnings.simplefilter("default", POSYDONWarning)                                
+                    SetPOSYDONWarnings("default", "POSYDONWarning")                         
                 else:
-                    warnings.simplefilter("ignore", POSYDONWarning)
+                    SetPOSYDONWarnings("ignore", "POSYDONWarning")
```
mkruckow: This is not the aimed use of this function. We should use `NoPOSYDONWarnings` here. Or did you missed to put a comment on the definition of `NoPOSYDONWarnings`, that we shouldn't need this function?

posydon/popsyn/binarypopulation.py
```diff
@@ -343,12 +343,15 @@ def _safe_evolve(self, **kwargs):
...
-                    print(warning_formatted, file=sys.stderr)
-                    binary.warnings.append(warning_formatted)
+                ## catching warnings with record=True automatically suppresses them being written standard error,
+                ## but we need record=True to save all warnings as a binary attribute (binary.warnings)
+                ## so, we manually reformat them and print to stderr
+                binary.warnings = []
+                if len(w) > 0:
+                    for x in w: 
+                        warning_formatted = warnings.formatwarning(x.message, x.category, x.filename, x.lineno, line=None)
+                        print(warning_formatted, file=sys.stderr)
```
mkruckow: I'm not sure, why you don't use the `warn` function here. With simply printing the formatted warning you build a workaround the filtering of warnings, which I think isn't your aim here.

posydon/binary_evol/DT/step_merged.py
```diff
@@ -17,7 +17,7 @@
 from posydon.utils.posydonerror import FlowError
 
 import warnings
```
mkruckow: In all files were we use `Pwarn`, we don't need to import `warnings` anymore unless we still have a dedicated use of it. Hence, we can remove it unless there is `warnings.*` anywhere in the file. This saves use from loading same stuff several times.

posydon/popsyn/binarypopulation.py
celiotine: I left a comment that I think the SetPOSYDONWarnings function is preferable because you can explicitly see the filter setting.

posydon/popsyn/binarypopulation.py
celiotine: The filtering and printing are two different things. We want to catch and record all of the warnings, and the warnings that are recorded depends on the filters. We need to record the warnings if we want to keep the functionality of saving a binary's warnings as an attribute. However, recording the warnings suppresses them being printed to output, so they must be manually printed.

celiotine: So if you filter-out the POSYDON warnings, they are not included in "w". If you want to keep the POSYDON warnings, they are saved in "w".

posydon/binary_evol/SN/step_SN.py
celiotine: The code I removed is not in the current development branch?

posydon/binary_evol/DT/step_merged.py
celiotine: I must have forgotten to remove the import. I very much understand the functionality of how imports and Pwarn work.

posydon/binary_evol/SN/step_SN.py
mkruckow: Oh, OK.
It reminded me of the code max added to capture inconsistencies from the interpolator.

posydon/popsyn/binarypopulation.py
mkruckow: OK. I already had in mind the replacement I was building for the catch. So that we can get other functionality on top of it. Additionally, I was not that happy about the catching function, because it is classified as not being thread save.

posydon/utils/posydonwarning.py
mkruckow: True, I made the functions for people who don't want to check the allowed filter actions on the warnings class. Would you know a better way than having the extra functions to ensure that lazy developers will get it without postponing it to later (and finally never do)?
Surely, I'd prefer if all code writers would be skilled enough to always know or at least check stuff.

posydon/popsyn/binarypopulation.py
```diff
@@ -317,7 +317,8 @@ def _safe_evolve(self, **kwargs):
                 binary = self.manager.generate(index=index, **self.kwargs)
             binary.properties = self.population_properties
 
-            with warnings.catch_warnings(record=True) as w:
+            with Catch_POSYDON_Warnings(record=True) as cpw:
```
celiotine: The filtering here is not working as expected anymore. When warnings_verbose is set to True, the POSYDON warnings do not show up consistently. 

For example, if I run a population with warnings_verbose=True in a Jupyter notebook, POSYDON warnings only show-up for the first evolved binary, which does not seem normal (before there was almost one POSYDON warning per binary). If I set warnings_verbose=True and run a small population, then set warnings_verbose=False, run a small population, then set it to True again, the POSYDON warnings do not show-up in the population at all the second time warnings_verbose=True. 

I tried adding-in the filters here explicitly like was being done before, but that does not fix the problem. Simply removing the reset_cache() block also does not fix the problem. So there is something deeper happening in the new Catch_POSDYON_Warnings function that is causing the unexpected behavior.

posydon/utils/posydonwarning.py
```diff
@@ -77,33 +105,336 @@ def __init__(self, message=''):
...
+        return cache
+    
+    def reset_cache(self):
+        """Resets the catched warnings."""
+        self.catched_warnings = []
+
+_CATCHED_POSYDON_WARNINGS = _Catched_POSYDON_Warnings()
+
+
+class Catch_POSYDON_Warnings:
```
celiotine, line 450: I am confused, does this catch all warnings, including python warnings, or just POSYDON warnings? I think we want to keep functionality to catch both, at least for the use in binarypopulation.

posydon/popsyn/binarypopulation.py
celiotine: If the new Catch_POSYDON_Warnings() function still suppresses the warnings being written to standard error if record=True, then we need to explicitly print the warnings at the end of the code block, like I was doing before, for them to show-up consistently.

posydon/utils/posydonwarning.py
```diff
@@ -77,33 +105,336 @@ def __init__(self, message=''):
...
+            Dictionary containing all options passed to warnings.warn.
+    """
+    filtered_warning = _apply_POSYDON_filter(warning)
+    if filtered_warning is None:
+        return
+    else:
+        warnings.warn(**filtered_warning)
+
+
+class _Catched_POSYDON_Warnings:
```
celiotine: A general comment, it is unclear to me how much of these new functions are *essentially* the same as the python source code for catch_warnings, and what changes introduce new functionality. I think it would be _very_ helpful, if there are only a few minor changes in functionality, to highlight exactly where those are in the code itself so that people understand the differences here. I know that you wrote a couple sentences in the docs at the top, but at least for me, that is not enough to really understand what was changed from the python source code, without having to go look at the python source code and compare.

posydon/popsyn/binarypopulation.py
mkruckow: It does, what the code tells: the filter is default (show the first warning of a type and suppress the later ones, it was a bug before, that it wasn't doing that always).
You can add
```
if self.kwargs.get("warnings_verbose", False):
    from posydon.utils.posydonwarning import AllPOSYDONWarnings
    AllPOSYDONWarnings()
```
(You can move the import up and merge it with the other if you like)

In the case of running multiple populations: it will work as you intend when you really run a new population (new python instance), if you run it again in a notebook, the records of the warning of the first run are still there, hence we talk about recurring warnings. Would you like to get a function, to reset our registry of warnings or getting the ability to have a context registry?

Removing the `reset_cache()` doesn't change anything on the filtering. So surly it can't effect any warnings, which are filtered out.

The new `Catch_POSYDON_Warnings()` does not print the warnings at the moment they occur, instead they will be printed to standard error at the end of the content (end of the with block). The statement `reset_cache()` suppresses that print out in case there are no warnings wanted.

posydon/utils/posydonwarning.py
mkruckow: We can't catch python warnings, because they never call any of our functions. But as we said that we want to have all our Warnings to be subclasses of POSYDONwarnings, those are all caught. It's our responsibility to not call other functions with arguments causing warnings in there.

posydon/utils/posydonwarning.py
mkruckow: I added more comments, please let me know, if there is need for more (tell please where you'd need more).

posydon/popsyn/binarypopulation.py
celiotine: Ok, I guess I am confused on a couple points then. I tried adding-in the code block you suggested for the filtering, and that did not fix the filtering. Ideally if warnings_verbose=True, we want the first instance of warnings to show-up for every binary separately, which is what was happening before, but now those warnings seem to be suppressed. 

Possibly related to your second point, I think before, in the python version of catch_warnings, the warning registry was being cleared every time a population was run, even in a notebook without refreshing the kernel. So I am confused what changed where this is not happening anymore. 

For printing the warnings, nothing was being printed at the end of the content block when I was running populations even when I removed the reset_cache() block, which is why I am confused about the functionality.

posydon/utils/posydonwarning.py
celiotine: Ok, I am thinking that in the main catch_warnings block in binarypopulation, we want to catch both, mostly because the way things are currently set-up, if _any_ warning for a binary occurs, we want to record this in the oneline dataframe. So we need to know if python warnings were caught, too. Would it make sense then to use the regular python catch_warnings there?

celiotine: In general if you have an idea of how to get the functionality in binarypopulation be consistent with what happening before using the new POSYDON warnings stuff, that would be helpful.

posydon/utils/posydonwarning.py
celiotine: Ok I think the additional comments are helpful!

posydon/popsyn/binarypopulation.py
mkruckow: OK, there are two things (buggy), which have took effect here with the python warnings purely:

1. The warnings may include some binary specific detail, e.g. the index, and so the python warnings call them based on text different.
2. There was a filter change made: python warnings clear their registry completely whenever a filter change is done, independent on the fact whether any warning in the registry would have been effected by the filter change.

The suggested code block should go at the beginning directly behind the with statement and would surprise me that it should have no effect.
To debug things you can try
```
import warnings
print(warnings.filters)
```
to see whether the filter was successfully added.
To check the warnings registry you can use
```
from posydon.utils.posydonwarning import print_stats
print_stats()
```
To see what is recorded in the POSYDON warnings registry.

mkruckow: I have added an option to use a context registry instead of the global one. Hence, it will issue the first warning in each context if used, without a need to change filters.

posydon/utils/posydonwarning.py
mkruckow: Having the exact same behavior as before is not possible, without put bugs in our code.

Now, there is an option for backward compatibility to put the python catching on top, which brings all the drawbacks of the python catching.
Be aware, that future code validation, e.g. in unit testing, will require the code to work without that option, because we should handle all warnings on our own.

bin/posydon-setup-grid
```diff
@@ -78,8 +79,8 @@ def find_inlist_from_scenario(source, gitcommit, system_type):
                                      )
             (clone, err) = proc.communicate()
         else:
-            print("git repository is already there, using that")
-
+            Pwarn("git repository is already there, using that",
+                  "OverwriteWarning")
```
astroJeff, line 83: Nothing is overwritten, right? The code just uses the repository that already exists. I don't see another warning that is appropriate here. Maybe just leave the print statement?

bin/posydon-setup-grid
```diff
@@ -107,7 +108,8 @@ def find_inlist_from_scenario(source, gitcommit, system_type):
                                      )
             (clone, err) = proc.communicate()
         else:
-            print("git repository is already there, using that")
+            Pwarn("git repository is already there, using that",
+                  "OverwriteWarning")
```
astroJeff, line 112: I have the same comment as above - the repository isn't really overwritten

bin/posydon-setup-grid
mkruckow: Later there will be a `checkout` and `pull` performed. Hence, if there is already one, which might be used in a different running script, it might break that, hence this warning.

celiotine (dismissed):

celiotine (approved):

astroJeff (approved):

Issue #291, opened 2024-04-18
BinaryStar.restore() does not reset the `step_names_history`


maxbriel:
When rerunning a binary I reset it, but the `step_names_history` is not reset. 

I would have expected for this to remove this history, when calling `BinaryStar.restore()` or `BinaryStar.reset()`.

celiotine:
I think this is happening because the step_names are added with a hook class and are not part of the BINARYPROPERTIES. This should hopefully be simple to fix.

celiotine:
[branch to fix this issue is here](https://github.com/POSYDON-code/POSYDON/tree/fix_binary_restore)

PR #306, opened 2024-05-22
Add extra hooks columns to binary.restore()

celiotine:
This PR adds extra columns associated with hooks in the SimulationProperties (e.g. step names, step times) to the BinaryStar and SingleStar restore() functions. It looks for any hooks classes in the simulation properties and checks if they add extra binary or star columns to the evolution. These column names now must defined in the hooks constructors as shown. 

NOTE: Currently, the hooks columns can only be restored if a binary is initialized and evolved with the desired simulation properties/hooks. The binary cannot be restored with these columns, for example, if the user reads a binary from an hdf file. If we want this functionality, the dev is more involved and I would suggest a separate PR.

### DIFF
--- a/posydon/binary_evol/singlestar.py
+++ b/posydon/binary_evol/singlestar.py
@@ -177,7 +177,7 @@ class SingleStar:
         for item in STARPROPERTIES:
             getattr(self, item + '_history').append(getattr(self, item))
 
-    def restore(self, i=0, hooks=[]):
+    def restore(self, i=0, hooks=None):
         """Restore the SingleStar() object to its i-th state, keeping the star history before the i-th state.
 
         Parameters
@@ -190,6 +190,9 @@ class SingleStar:
             object containing this SingleStar(), if applicable. This parameter is 
             automatically set when restoring a BinaryStar() object. 
         """
+        if hooks is None:
+            hooks = []
+            
         # Move current star properties to the ith step, using its history
         for p in STARPROPERTIES:
             setattr(self, p, getattr(self, '{}_history'.format(p))[i])
--- a/posydon/binary_evol/singlestar.py
+++ b/posydon/binary_evol/singlestar.py
@@ -177,24 +177,35 @@ class SingleStar:
         for item in STARPROPERTIES:
             getattr(self, item + '_history').append(getattr(self, item))
 
-    def restore(self, i=0, delete_history=True):
-        """Restore the object to the i-th state.
+    def restore(self, i=0, hooks=[]):
+        """Restore the SingleStar() object to its i-th state, keeping the star history before the i-th state.
 
         Parameters
         ----------
         i : int
             Index of the star object history to reset the star to. By default
             i == 0, i.e. the star will be restored to its initial state.
+        hooks : list
+            List of extra hooks associated with the SimulationProperties() of the BinaryStar()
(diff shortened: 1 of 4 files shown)


posydon/popsyn/binarypopulation.py
```diff
@@ -644,11 +644,15 @@ def from_hdf(self, indices, where=None, restore=False):
...
             Query performed on disk to select history and oneline DataFrames.
         restore : bool
-            Restore binaries back to initial conditions.
+            If true, restore binaries back to initial conditions.
 
         """
         if where is None:
-            query_str = 'index==indices'
+            if len(indices) > 0:
+                query_str = f'index=={indices}'
```
ka-rocha: The original code had `query_str = 'index==indices'` but the edits have replaced this with an f string. Is there a reason this behavior was changed? In principle these query strings are executed as python code during the search in the h5 and can reference local variables.

posydon/binary_evol/singlestar.py
```diff
@@ -177,24 +177,35 @@ def append_state(self):
         for item in STARPROPERTIES:
             getattr(self, item + '_history').append(getattr(self, item))
 
-    def restore(self, i=0, delete_history=True):
-        """Restore the object to the i-th state.
+    def restore(self, i=0, hooks=[]):
```
ka-rocha: FYI setting mutables (e.g. an empty list) in kwargs may lead to unexpected behavior if any updates occur within the function. Kwargs are bound during _define time_ and not _run time_. Catching it like this guarantees to avoid this issue:
```python
def restore(self, i=0, hooks=None):
    ...
    if hooks is None:
        hooks = []
```

posydon/popsyn/binarypopulation.py
celiotine: The f-string makes this behavior more readable.

ka-rocha: I think the original is actually more clear. If `indices` is an integer this code will fail. Using f-strings requires the conversion of `indices` to the string representation of the object before the logical evaluation which may raise errors down the road (e.g. interpreted as a float / int / string is unclear). In the original, the logical evaluation is executed without this step.

celiotine: Indices by definition must be a list of integers as explained in the docs for this function, so I don't forsee this being an issue.

ka-rocha: I think this function should support integers (as it did with the original code). You also do not explicitly check that this will be a list of integers so any error raised will be opaque. You also have not provided any sufficient reasoning for why the string conversion may not cause issues downstream. These edge cases will need to be handled and checked that they do not impact IO performance.

Furthermore, the original code is what is described in the pandas documentation and this change would remove functionality for advanced queries within tables.

celiotine: You did not see my most recent commit, I changed things back.

celiotine: Going forward I would prefer discussion be handled in a tone that is less aggressive, thanks!

ka-rocha: Apologies!

ka-rocha (approved):
😎

astroJeff (approved):

maxbriel:
This solved Issue #291 🔱

PR #309, opened 2024-05-23
added the M4 mu4 quantities as stored in step_SN, to be kept in the s…

ezapartas:
…inglestar object as scalars


ezapartas:
The idea is to use M4 and m4 for estimates of Ekin of each SN event. WIth @dimsour94 we already have some results with this method, but we relied up to now only to nearest neighbour interpolation, where M4 and m4 we recalculated. We want them to be added in the post-processin of the interpolated quantities

mkruckow:
We should discuss this in the next Dev meeting, whether we need those values, which are specific to a SN prescription or whether the new values at He depletion are sufficient to simply recalculate M4 and mu4 in case they are needed.

ezapartas:
Keeping this discussion of whether to add these values in the interpolated quantities too.

ezapartas:
avg_c_in_c_core_at_He_depletion and co_core_mass_at_He_depletion should in general also be stored to be used in post-processing to get info about the collapse. In principle we have functions to recalculate them after a POSYDON run, but the functions are working on a singlestar POSYDON object not the data for the h5d5 file, so it is much much more straightforward to keep these values during the calculation of step_SN (and the interpolator).

mkruckow:
The values at He depletion are ready to be used in the star object, which takes them from the ones from the hdf5 file of the grids, see https://github.com/POSYDON-code/POSYDON/pull/281/files#diff-e4b192b1c4e79f131a84d9c341737303dc7932b4a3c7136ea75c33592320bf86

ezapartas:
As a side thing, the PR also allows for *_He_depletion parameters to be an scalar singlestar output of a POSYDON population

posydon/grids/post_processing.py
```diff
@@ -70,13 +71,13 @@ def print_CC_quantities(EXTRA_COLUMNS, star, MODEL_NAME=None):
             print(format_val.format(MODEL_NAME,
                     star.state, star.SN_type, star.f_fb,
                     star.mass, spin, star.m_disk_accreted,
-                    star.m_disk_radiated))
+                    star.m_disk_radiated, star.M4, star.mu4))
```
mkruckow: We'd need to add a check, whether those values are None. Cf. the check for the spin, otherwise we'll get a lot of exceptions for this try.

It would be even better to restructure the print out function here completely -- but I think that's beyond the scope of this PR.

posydon/popsyn/io.py
```diff
@@ -146,15 +146,19 @@
     'f_fb': 'float64',
     'spin_orbit_tilt_first_SN': 'float64',
     'spin_orbit_tilt_second_SN': 'float64',
+    'M4': 'float64',
+    'mu4': 'float64',
```
mkruckow, line 150: I'd suggest to move `M4` and `mu4` to `EXTRA_STAR_COLUMNS_DTYPES`.

posydon/popsyn/io.py
```diff
@@ -146,15 +146,19 @@
...
+    'avg_c_in_c_core_at_He_depletion' : 'float64',
+    'co_core_mass_at_He_depletion' : 'float64',
```
mkruckow, line 152: I'd suggest to move `avg_c_in_c_core_at_He_depletion` and `co_core_mass_at_He_depletion` to `STARPROPERTIES_DTYPES`.

posydon/popsyn/io.py
ezapartas: They are CC parameters, so I put them where I saw the SN_type and other SN parameters too

ezapartas: I am not sure I understand the difference between  EXTRA_STAR_COLUMNS_DTYPES and SCALAR_NAMES_DTYPES. Are both at least available for output in a POSYDON run, by adding then in the scalar_names in the .ini file of the run?

posydon/grids/post_processing.py
ezapartas: ok yes. Why dont we do the same for m_disk_accreted, m_disk_radiated?

mkruckow: They are 0 not None, but we could do for them, too.

posydon/popsyn/io.py
mkruckow: I'm not sure about all the details of the split-up. I thought to put M4 and mu4 together with the disk masses, because M4 and mu4 only make sense for a single SN prescription. Similarly the disk masses only make sense for BHs.

posydon/popsyn/io.py
mkruckow: You put them to `EXTRA_STAR_COLUMNS_DTYPES`, why didn't you put them to `STARPROPERTIES_DTYPES`?

mkruckow (dismissed):
Looks OK, now.
We need to check carefully interactions with PR #340 .

posydon/popsyn/io.py
maxbriel: Given that these are just a single value and not a history. This is the right location.

posydon/popsyn/io.py
maxbriel: The intent here was to split "standard" output and special output between  `SCALAR_NAMES_DTYPES` and `EXTRA_STAR_COLUMNS_DTYPES`. I'm actually surprised that the `m_disk_*` are located there, since they don't have a history.

The `SCALAR_NAMES_DTYPES` is supposed to be standard output with only 1 value per STAR.

maxbriel (dismissed):
Please note that this requires retraining of the interpolators!

ezapartas:
Included in PR #340

PR #316, opened 2024-05-29
Add correction for eccentric pre SN orbits for SNflag2

mkruckow:
It turned out, that the derivation of SNflag2 assumed a circular pre supernova orbit to simplify the equation.
We use this flag to determine, whether a post SN orbit is bound or disrupted. Hence, we classified too many eccentric pre SN cases as disrupted in the SN.
In a circular pre SN orbit, the correction factor would be 1, because it becomes Kepler's third law.

P.S.: I encountered the issue by making a check with 0 kick, where the two inequalities of the SNflag would become equal, hence no kick could hide the missing factor anymore. Maybe this would not mean that we had to many disrupted systems overall, but the kick range leading to disruptions was shifted surely.

### DIFF
--- a/posydon/binary_evol/SN/step_SN.py
+++ b/posydon/binary_evol/SN/step_SN.py
@@ -1723,8 +1723,13 @@ class StepSN(object):
 
                 # SNflag2: Equations 22-23, Willems, B., Henninger, M., Levin, T., et al. 2005, ApJ, 625, 324
                 # (see, e.g., Kalogera, V. & Lorimer, D.R. 2000, ApJ, 530, 890)
-                tmp1 = 2 - Mtot_pre / Mtot_post * (Vkick / Vr - 1) ** 2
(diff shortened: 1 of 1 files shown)


tassos25 (commented):
Can you add a reference, with specific equations if possible, from were you got the correction factor?

mkruckow:
> Can you add a reference, with specific equations if possible, from were you got the correction factor?

I derived it myself. Thus the reference would be: my notes. We can may add it in the appendix or a section in the v2 paper.

tassos25 (approved):

maxbriel (approved):
I would suggest adding the derivation to either the draft or to somewhere archivable.

astroJeff:
@mkruckow can you post your notes? I'm having trouble deriving this check.

tassos25 (approved):

PR #317, opened 2024-05-30
Further change to make MESA's and POSYDON's RLO detection the same

mkruckow:
The reason come from the recent repost-processing of the grids, see [slack](https://posydon.slack.com/archives/C04K8DQ4EMU/p1716634325558919).

Change `and` to `or` for the two RLO conditions: the mass-transfer rate and the Roche lobe-filling factor.


devimisra (approved):

tassos25 (approved):
Seems good.

Issue #307, opened 2024-05-22
Clarify logic for redirecting binaries in CO-H / He HMS grids.


ka-rocha:
https://github.com/POSYDON-code/POSYDON/blob/cf933e466b96ac70e9823b02ff52849585928a92/posydon/binary_evol/MESA/step_mesa.py#L1416-L1424

It is unclear what this `else` clause is doing and it is causing some binaries to go through loops where they are redirected multiple times back into the CO-HMS steps.

maxbriel:
This code should not be there anymore. I already removed it in PR #255.

It looks like it came back in PR #287, the error checking PR.
Additionally there's code in `step_mesa.py` that was added with PR #255, which made sure to catch all the different types of out-of-grid, such that no cyclical redirects should happen.

@celiotine and @ka-rocha, this should be an easy fix. Would you be able to add the code back in and remove the part that should have been removed?

PR #255, opened 2024-02-20
CO-H(e)MS period outside grid being failed fix

maxbriel:
The CO-H(e)MS looping fix was too strong and would cause too many systems to fail.

This fix checks if the period is outside the grid period and then redirects the binary back to detached.

example binary at 1Zsun
```
STAR1 = SingleStar(**{'mass': 58.947503,
                      'state': 'H-rich_Core_H_burning'})
STAR2 = SingleStar(**{'mass': 56.660506,
                      'state': 'H-rich_Core_H_burning'})

BINARY = BinaryStar(STAR1, STAR2,  
                    **{'time': 0.0, 'state': 'detached', 'event': 'ZAMS', 'orbital_period':211.300659, 'eccentricity': 0.0},
                    properties = sim_pop)
```

PR #320, opened 2024-05-31
Fixes to error checking

celiotine:
This PR restores code written by @maxbriel in PR #255 that was accidentally deleted from step_mesa. This code addresses issue #307. 
In addition, this PR adds traceback messages to the binary attributes when running a population and reformats some error messages.


posydon/binary_evol/MESA/step_mesa.py
```diff
@@ -1514,27 +1523,36 @@ def __call__(self, binary):
...
-                    raise GridError('CO_HeMS_RLO binary outside grid and coming from detached')
+        # period inside the grid, but m2 outside the grid with flipped stars
+        elif ((self.flip_stars_before_step and
+                self.p_min <= p <= self.p_max and
+                m2 < self.m1_min or m2 > self.m1_max)):
+            set_binary_to_failed(self.binary)
+            raise GridError(f'The mass of m1 ({m2}) is outside the grid,'
+                                ' while the period is inside the grid.')
 
+        else:
```
ka-rocha, line 1555: In this final `else` - is it known what binaries are expected to enter here? It is hard for me to tell if all the other `elif`'s are catching the expected behavior and this is a final fail safe or if this `else` is actually catching a known case that we understand. I think just some comment explaining the intended behavior may be nice!

ka-rocha (dismissed):

posydon/binary_evol/MESA/step_mesa.py
maxbriel: I tried to cover all case where something might not work because of the associated grid. The `else`, on the other hand, covers binaries, where we can redirect them to the detached step.
There are known cases of binaries going through the redirect. The example below is from an slack post on the developers channel from Zepei. This is/was intended behaviour, although naming could be better.

Screenshot 2024-04-11 at 08 19 32

We could change the logic to manually include these redirects and catch any unexpected system with the else instead (which might be better).

maxbriel (dismissed):
This looks good. We can put an issue in place in case we want to update the logic to be more selective in what gets redirected from this step

maxbriel (approved):

mpgalleg (approved):
Looks good to me!

Issue #318, opened 2024-05-30
matplotlib version issues


ezapartas:
Using the new posydon_v2_dev_env in quest by sourcing /projects/b1119/.bash_shared_v2_dev , I got this matplotlib error:

```
  File "/projects/b1119/POSYDON_SHARED_v2_dev/posydon/popsyn/synthetic_population.py", line 26, in <module>
    import posydon.visualization.plot_pop as plot_pop
  File "/projects/b1119/POSYDON_SHARED_v2_dev/posydon/visualization/plot_pop.py", line 19, in <module>
    cm = plt.cm.get_cmap('tab20')
         ^^^^^^^^^^^^^^^
AttributeError: module 'matplotlib.cm' has no attribute 'get_cmap'
```
This function got removed in matplotlib 3.9.0, hence we either need to limit the version or rewrite our code. (@mkruckow)
https://matplotlib.org/stable/api/prev_api_changes/api_changes_3.9.0.html#removals

sgossage:
Just posting this for posterity:

If we decide to stay with `matplotlib` version 3.9.0 with our new v2 conda environments, we will need to change the current function call to e.g.,

```
import matplotlib as mpl
cm = mpl.colormaps.get_cmap('tab20')
```

To stay with the current function call, we should downgrade `matplotlib` to something like version 3.7.1 (which is a version used in previous environments where this call is valid).

mkruckow:
For the moment we down graded the matplotlib version in #336 .
The way of #322 will be the better one after it got tested.

celiotine:
PR #322 was successfully merged, so this issue should be resolved.

PR #322, opened 2024-06-01
matplotlib changes

ezapartas:
quick solution to #318 issue of the newest version of matplotlib


ezapartas:
@ka-rocha consider whether we need to make a similar change in active_learning/psy_cris, where `get_cmap` is also found in the code.

tassos25:
We should first make a decision about which matplotlib version we want to use, and if it is 3.9, we should specify it in the requirements files. Only then this PR should be accepted.

sgossage:
The easy route is likely downgrade our new v2 environments to matplotlib 3.7.1 which we know works, otherwise we should probably run tests to ensure 3.9.0 does not break our code in any additional unexpected ways

mkruckow:
We made a downgrade of the required matplotlib in preparation for the f2f-meeting, see #336 . Hence, please merge the current development into your branch and free the installation restriction again within this PR.

astroJeff:
Update (July 25): Manos to include updates to setup.py (or equivalent) the installed/required version of matplotlib. Can we also check if this code is backwards compatible with older matplotlib versions?

posydon/visualization/plot_pop.py
```diff
@@ -2,10 +2,10 @@
     "Simone Bavera <Simone.Bavera@unige.ch>",
     "Max Briel <max.briel@unige.ch>",
     ]
-    
+
 import os
 import numpy as np
-import matplotlib.pyplot as plt
```
mkruckow, line 8: This can't be removed, `plt` is still used at other places.

ezapartas: It exists again a few lines below, it was duplicate there, probably by an earlier error

mkruckow: I'd probably remove the other instance, to keep the matplotlib imports together.

astroJeff:
Update (Aug 1): @ezapartas to meet with @dimsour94 to update tomorrow - hopefully accept in the next couple days.

mkruckow:
I made a test with this branch to create some plots with the pipeline on the CO-HMS grid: all fine!

mkruckow (approved):
I made a test with this branch to create some plots with the pipeline on the CO-HMS grid: all fine!

dimsour94 (approved):

Issue #303, opened 2024-05-13
Incorrect classification/remnant masses for BHs when using interpolated values in SN step


celiotine:
When running with `use_interp_values = True` in the SN settings, there is incorrect handling of some remnant masses. Specifically, these objects are either being misclassified as BHs or are assigned incorrect remnant masses. An example binary where this behavior is present is:

```
star_1 = SingleStar(**{'mass': 38.741115, 'state': 'H-rich_Core_H_burning',\
                'natal_kick_array': [21.113771, 2.060135, 2.224789, 4.089729]})
star_2 = SingleStar(**{'mass': 27.776178, 'state': 'H-rich_Core_H_burning',\
                'natal_kick_array': [282.712103, 0.296252, 1.628433, 5.623812]})

binary = BinaryStar(star_1, star_2,
                   **{'time': 0.0, 
                      'state': 'detached', 
                      'event': 'ZAMS',
                      'orbital_period': 93.387072, 
                      'eccentricity': 0.0},
                    properties = sim_prop)
```
Quest .ini file path: `/projects/b1119/cliotine/dev_general/debugging/params_Monica.ini`

This is the behavior when running with `use_interp_values = True` (S2 BH is way too low mass or is actually a NS):

<img width="652" alt="Screen Shot 2024-05-13 at 3 56 59 PM" src="https://github.com/POSYDON-code/POSYDON/assets/13874913/33fd494f-d4e0-4f4f-bb2c-cadb73c7eb57">

This is the behavior when running with `use_interp_values = False` (S2 BH looks fine):

<img width="658" alt="Screen Shot 2024-05-13 at 3 55 11 PM" src="https://github.com/POSYDON-code/POSYDON/assets/13874913/ff5b62d6-538d-4316-b05f-cf2c448b1e7b">

mkruckow:
- Could you please add the SN_type to your two examples?
- Can you may take the state binary at the "oRLO2" for your first example to run from there with the change of the `use_interp_values` to exclude that the slight change in primary mass during its CC don't move the secondary out of the problematic regime, which might be tight.

celiotine:
Sure, here is the evolution re-evolving from oRLO2. The outcome is similar to above:

With `use_interp_values = True`:
<img width="652" alt="Screen Shot 2024-05-17 at 10 25 38 AM" src="https://github.com/POSYDON-code/POSYDON/assets/13874913/b9eea071-296a-413a-a070-5095dfb2ddbd">

[...]

Issue #280, opened 2024-03-18
NS/WD NaN mass, SN_type & CO_type not consistent


maxbriel:
For systems with state=NS and SN_Type=WD, no mass is present in the MODELxx.
Is this expected behaviour for the models, when the state and SN type do not match?

Should we recalculate the mass with the state/SN_type based on the preSN model parameters?

```
STAR1 = SingleStar(**{'mass': 7.592921,
                      'state': 'H-rich_Core_H_burning'})
STAR2 = SingleStar(**{'mass':5.038679 ,
                      'state': 'H-rich_Core_H_burning'})

BINARY = BinaryStar(STAR1, STAR2,
                    **{'time': 0.0, 'state': 'detached', 'event': 'ZAMS', 'orbital_period':5.537807, 'eccentricity': 0.0},
                    properties = sim_pop)
```

 S1_mass | S2_mass | S1_he_core_mass | S1_co_core_mass | S1_state | S2_state | time | state | event | orbital_period | eccentricity
 -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
7.592921 | 5.038679 | NaN | NaN | H-rich_Core_H_burning | H-rich_Core_H_burning | 0.000000e+00 | detached | ZAMS | 5.537807 | 0.0
1.229538 | 5.234164 | 1.229538 | 1.120187 | stripped_He_Central_He_depleted | H-rich_Core_H_burning | 5.204840e+07 | detached | CC1 | 11.434389 | 0.0
NaN | 5.234164 | NaN | NaN | WD | H-rich_Core_H_burning | 5.204840e+07 | detached | <NA> | NaN | 0.0
NaN | 5.234164 | NaN | NaN | WD | H-rich_Core_H_burning | 5.204840e+07 | detached | <NA> | NaN | 0.0

maxbriel:
Small update: This also occurs for some BHs.

celiotine:
Adding an update: I have also noticed this behavior is happening for ECSN when `use_interp_values = True`. This is causing problems later in the evolution, because the resulting NS masses are NaN, which is crashing the evolution in the detached step.

maxbriel:
@philipp-rajah  Here's a full example. I've send you the `population.ini`

```
import numpy as np
import os
from posydon.config import PATH_TO_POSYDON, PATH_TO_POSYDON_DATA
import pandas as pd

from posydon.binary_evol.binarystar import BinaryStar, SingleStar
[...]

PR #324, opened 2024-06-04
step_SN mass check; force manual calculation if not matching

maxbriel:
This is a temporary fix for several issues:
1. Incorrect masses BH.   #303 
2. NS/WD NaN masses #280 

The previous code had the following issue:
The check for SN_type and CO_type would not recalculate the masses from the interpolator. Thus, giving a NaN or a wrong remnant mass from a different CO_type class.

Now, it forces all mismatching SN_type and CO_types to be fully recalculated with the given remnant mass prescriptions based on the preSN star. This calculation is the same as if `use_interp_values=False` or for stars evolved with `step_detached` or `step_disrupted`.


celiotine:
Ok I have some sad news. It seems like this error is occurring for binaries where the SN type and object type _do actually match_, so this fix is not catching everything. Here is an example problem binary where the SN type is ECSN and the object type is NS:

```
star_1 = SingleStar(**{'mass': 7.1125623905943485, 'state': 'H-rich_Core_H_burning',\
                'natal_kick_array': [19.068527695032785, 3.5633775052142633, 1.2378597315236688, 2.4627960384344374]})
star_2 = SingleStar(**{'mass': 5.041150596272908, 'state': 'H-rich_Core_H_burning',\

binary = BinaryStar(star_1, star_2,
                   **{'time': 0.0, 
                      'state': 'detached', 
                      'event': 'ZAMS',
                      'orbital_period': 35.66303442178471, 
                      'eccentricity': 0.0},
                    properties = sim_prop)
```
Evolving this binary throws the usual error resulting from NaN values:
Screen Shot 2024-06-04 at 12 26 31 PM

It looks like this is happening for most combinations of SN types and object types (e.g. CCSN/NS, CCSN/BH, etc.) For the temp fix we can explicitly check if the interpolated values are NaN regardless of mismatch, but it seems the deeper issue may be different than what we thought?

For binaries that do have a mismatch, the interpolated values are sometimes NaN and sometimes not:
Screen Shot 2024-06-04 at 2 34 55 PM
Screen Shot 2024-06-04 at 2 35 12 PM

celiotine:
My most recent commit adds an explicit check if the interpolated values are NaN (rn just checks if the mass is NaN). This is fixing the issue for now, I am not getting any more errors as I reported above.

celiotine (dismissed):

tassos25:
@celiotine will coordinate with @ZepeiX and @philipp-rajah to decide if this PR is required on top of #325

astroJeff (dismissed):

maxbriel:
Just to have it recorded: Even with retrained interpolators, the non-match situation still occurs and, thus, we should catch these situations and process them.
A warning is thrown as well to inform the user that something went wrong, but it will try to continue with the step_SN.

ZepeiX:
Regarding the mismatch between SN_type and CO_type, is that only 'WD' and 'NS'? 
I have some concerns about whether we should turn to use 'profiles' in this case. Not profiles actually, but the interpolated core properties. Because it might be a problem of the interpolation of SN_type not the CO properties, which could probably be fixed in the future.

maxbriel:
The issue is not only with WD's and NS's, but I've also seen discrepancies between BHs and NSs at their boundary.

I think the main question is: what interpolator/classifier do you trust more? The `SN_type` or the `CO_type`?
And if there's a mismatch do we still trust the `CO_interpolation_class` values?

This is similar to the already present check and warning is the model does not match up:
https://github.com/POSYDON-code/POSYDON/blob/c7ecd73ef4d5b05345cf9f762cf057f1251a3c6e/posydon/binary_evol/SN/step_SN.py#L524

maxbriel:
@ZepeiX and I had a short discussion on this and I dug a bit on why the discrepancy occurs.

The discrepancy between `SN_type` and `CO_type` originated from them being trained on different classifiers.
`SN_type` was trained on the mass transfer classification, while the `CO_type` was trained on its own classes.
See the code how it used to be: https://github.com/POSYDON-code/POSYDON/blame/3ca9243d3a2b459a5a9d4ca341c611e1365d0582/bin/run-pipeline#L342-L353

The `CO_type` got changed to `CO_interpolation_class`, but `CO_type` and `SN_type` were now both trained within the mass transfer classes. It's unclear how sparse this could become, possibly resulting in more issues with the values.

With PR #325, this makes `CO_type` and `SN_type` be trained per `CO_interpolation_class`.
This should also remove discrepancies between the `CO_type` and `SN_type`.
However, @ZepeiX has noticed a binary "going out of hull" and giving a `CO_type` and `SN_type` discrepancy. This indicates another potential issue. 

**I think a fallback, as introduced here, would still be useful in such rare cases. Since a warning is given, this is clear to the user something happened, although I've got no strong opinions.**

ZepeiX (approved):

astroJeff (approved):

PR #325, opened 2024-06-06
Zepei fix interpolator

ZepeiX:
fix NS mass above 1.5
fix NaN mass

### DIFF
--- a/bin/posydon-run-pipeline
+++ b/bin/posydon-run-pipeline
@@ -421,7 +421,9 @@ def train_interpolators(i, path_to_csv_file, verbose=False):
                     "interp_classes": interp_classes,
                     "out_keys": out_keys,
                     "class_method": "kNN",
-                    "c_keys": [f'S{i}_{MODEL_NAME}_CO_interpolation_class'],
+                    "c_keys": [f'S{i}_{MODEL_NAME}_CO_interpolation_class', 
+                               f'S{i}_{MODEL_NAME}_CO_type',
+                               f'S{i}_{MODEL_NAME}_SN_type'],
                     "c_key": f'S{i}_{MODEL_NAME}_CO_interpolation_class'
                 },   
             )
--- a/posydon/interpolation/IF_interpolation.py
+++ b/posydon/interpolation/IF_interpolation.py
@@ -397,10 +397,6 @@ class BaseIFInterpolator:
             A list of classes that the simulation tracks can
             fall into. Usually specified as the mass transfer type. This only
             needs be specified if interp_method is a list.
-        valid_classes : list of strings
-            A list of mass-transfer classes that are kept for interpolations 
-            (exclude e.g. 'not_converged' class).
-            valid_classes = ['stable_MT', 'unstable_MT', 'no_MT', 'stable_reverse_MT']
         class_method : string or list of strings
             The classification method to be used, either kNN or
             1NN.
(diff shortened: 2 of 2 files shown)


tassos25:
@ZepeiX and @philipp-rajah will meet to review this PR and give an update on slack. @maxbriel will also have a quick look.

posydon/interpolation/IF_interpolation.py
```diff
@@ -397,6 +397,10 @@ def __init__(self, grid=None, in_keys=None, out_keys=None, in_scaling=None,
             A list of classes that the simulation tracks can
             fall into. Usually specified as the mass transfer type. This only
             needs be specified if interp_method is a list.
+        valid_classes : list of strings
```
mkruckow: Why do you add a docstring for a Parameter, which is not supported by/add to this function?

posydon/interpolation/IF_interpolation.py
```diff
@@ -467,7 +471,7 @@ def __init__(self, grid=None, in_keys=None, out_keys=None, in_scaling=None,
             self.interp_in_q = self._interpIn_q(grid)
             self.XT, self.YT = self._grid2array(grid)
             self.valid = self._setValid(
-                grid.final_values[self.c_key], self.XT)
+                grid.final_values["interpolation_class"], self.XT)
```
mkruckow, line 477: Could you please comment, why we need to have here a hard coded "interpolation class"? For me it feels in this way you make the use case of c_key obsolete. @philipp-rajah, could you please check this change in detail?

posydon/interpolation/IF_interpolation.py
```diff
@@ -598,8 +602,8 @@ def _setValid(self, ic, X):
         which = np.isnan(np.sum(X, axis=1))
         valid[which] = -1
         print(f"Discarded {np.sum(which)} binaries with nans in input values.")
-
-        for i, flag in enumerate(self.interp_classes):
+        valid_classes = ['stable_MT', 'unstable_MT', 'no_MT', 'stable_reverse_MT']
```
mkruckow: Again, this hard codes stuff and takes flexibility away. @philipp-rajah, please check that this is what we want.

posydon/interpolation/IF_interpolation.py
ZepeiX: This is the problem I found. When training CC_quantities, self.c_key=['BH','NS','WD','None']. But we were using "interpolation_class" to remove failed binaries. So self.c_key doesn't work for CC_quantities and we got Nan mass because we interpolate mass including not converged binaries.

posydon/interpolation/IF_interpolation.py
mkruckow: Wouldn't it be better to add a parameter to the interpolation class, which contains the `valid_classes`?
I'm not sure how others think about, but I'd may say it would be better to have a negative version, hence which of the interpolation classes to exclude from validation. Opinions?

ZepeiX: invalid_classes = ['not_converged', 'ignored_no_BH', 'ignored_no_RLO','initial_MT']
valid_classes = np.unique(ic)
for i in invalid_classes:
    while( i in valid_classes):
        valid_classes.remove(i)

posydon/interpolation/IF_interpolation.py
maxbriel: @ZepeiX This is a good explanation of what's going on.

Was there a change in how we process non-valid models?
As in, originally we used to exclude them in the parameter calculation, while now we run them through it?
As such, we get final values for non-valid models?

bin/posydon-run-pipeline
maxbriel, line 1: This looks like it's more in-line with what was there before PR #237. [see here](https://github.com/POSYDON-code/POSYDON/blame/3ca9243d3a2b459a5a9d4ca341c611e1365d0582/bin/run-pipeline#L349)

maxbriel (approved):

tassos25 (approved):

PR #329, opened 2024-06-08
Stretch the pipeline to avoid too busy filesystem

mkruckow:
Matplotlib likes to reuse some cache files but at the same time requests them to be readable in unlocked way from disk within 5 seconds, this can cause trouble in case of large arrays.
- [x] Use slurm ID to let jobs wait instead of random time. This scales better with a larger number of slurm jobs.


astroJeff (approved):
I like this idea

maxbriel (approved):

PR #335, opened 2024-06-18
Update common_functions.py

ZepeiX:
setting lambda=np.nan raises the error
ValueError("lambda_CE has a negative value")


mkruckow (approved):

mkruckow:
A case were too much negation hurts. Using `if (lambda_CE < -tolerance):` would have worked, too. Now it's more explicit to allow `np.nan`.

ezapartas (approved):

Issue #341, opened 2024-06-25
IF interpolation all NaN slices


maxbriel:
@ZepeiX showed that lambda values for grid slices in the CO_HeMS(_RLO) grid have lambda values ( ie `S1_lambda_CE_pure_He_star_1cent`) that are **all** NaN. PR #338 still has NaN slices in training.

**Current behaviour:**
As such, these parameters are not included in the `BaseIFInterpolator.evaluate()` output. This function returns interpolator values from `out_keys` and `out_nan_keys` stored in the BaseIFInterpolator.
Because we set `out_keys` as an input based on the presence of at least 1 non-nan value in the grid, the key is excluded and`out_nan_keys` is not calculated, since it's not part of `BaseIFInterpolator.__init__`.

**Expected behaviour**
We want the interpolator to return these grid values to allow for CE evolution.
As such, I've added the `out_nan_keys` as an input to the `BaseIFInterpolator.__init__` in PR #342

celiotine:
This should have been resolved with PR #342

PR #342, opened 2024-06-25
IF interpolator `out_nan_keys` as an input

maxbriel:
See Issue #341 

Adds the `out_nan_keys` to the training parameters and as an input to the `BaseIFInterpolator.__init__`.

### DIFF
--- a/bin/posydon-run-pipeline
+++ b/bin/posydon-run-pipeline
@@ -437,7 +437,9 @@ def train_interpolators(i, path_to_csv_file, verbose=False):
                     "out_keys": out_keys,
                     "out_nan_keys": out_nan_keys,
                     "class_method": "kNN",
-                    "c_keys": [f'S{i}_{MODEL_NAME}_CO_interpolation_class',],
+                    "c_keys": [f'S{i}_{MODEL_NAME}_CO_interpolation_class', 
+                               f'S{i}_{MODEL_NAME}_CO_type',
+                               f'S{i}_{MODEL_NAME}_SN_type'],
                     "c_key": f'S{i}_{MODEL_NAME}_CO_interpolation_class'
                 },   
             )
--- a/bin/posydon-run-pipeline
+++ b/bin/posydon-run-pipeline
@@ -376,6 +376,11 @@ def train_interpolators(i, path_to_csv_file, verbose=False):
                                    and any(~np.isnan(grid.final_values[key]))
                                    and "MODEL" not in key)]
     
+    out_nan_keys = [key for key in grid.final_values.dtype.names
+                    if type(grid.final_values[key][0]) != np.str_
+                    and all(np.isnan(grid.final_values[key]))
+                ]
+    
     # define all string keys in final values
     c_keys = ['interpolation_class', 'S1_state', 'S2_state', 'mt_history']
     for MODEL_NAME in MODELS.keys():
@@ -389,6 +394,7 @@ def train_interpolators(i, path_to_csv_file, verbose=False):
             "interp_method": interp_method,
             "interp_classes": interp_classes,
             "out_keys": out_keys,
+            'out_nan_keys': out_nan_keys,
             "class_method": "kNN",
             "c_keys": c_keys,
             "c_key": "interpolation_class"
@@ -407,6 +413,12 @@ def train_interpolators(i, path_to_csv_file, verbose=False):
                                    (type(grid.final_values[key][0]) != np.str_)
                                    and any(~np.isnan(grid.final_values[key]))
                                    and f"S{i}_{MODEL_NAME}" in key)]
+            
+            out_nan_keys = [key for key in grid.final_values.dtype.names
+                            if type(grid.final_values[key][0]) != np.str_
+                            and all(np.isnan(grid.final_values[key]))
+                        ]
+            
             # get interpolations classes dynamically
             interp_method = []
             interp_classes = []
@@ -420,6 +432,7 @@ def train_interpolators(i, path_to_csv_file, verbose=False):
                     "interp_method": interp_method,
                     "interp_classes": interp_classes,
                     "out_keys": out_keys,
+                    "out_nan_keys": out_nan_keys,
                     "class_method": "kNN",
                     "c_keys": [f'S{i}_{MODEL_NAME}_CO_interpolation_class'],
                     "c_key": f'S{i}_{MODEL_NAME}_CO_interpolation_class'
--- a/bin/posydon-run-pipeline
+++ b/bin/posydon-run-pipeline
@@ -376,10 +376,12 @@ def train_interpolators(i, path_to_csv_file, verbose=False):
                                    and any(~np.isnan(grid.final_values[key]))
                                    and "MODEL" not in key)]
     
-    out_nan_keys = [key for key in grid.final_values.dtype.names
-                    if type(grid.final_values[key][0]) != np.str_
-                    and all(np.isnan(grid.final_values[key]))
-                ]
+
+    out_nan_keys = [key for key in grid.final_values.dtype.names if (
+                             key != "model_number" and
+                            (type(grid.final_values[key][0]) != np.str_)
+                            and all(np.isnan(grid.final_values[key]))
+                            and "MODEL" not in key)]
     
     # define all string keys in final values
     c_keys = ['interpolation_class', 'S1_state', 'S2_state', 'mt_history']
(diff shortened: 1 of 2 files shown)


bin/posydon-run-pipeline
```diff
@@ -376,6 +376,11 @@ def train_interpolators(i, path_to_csv_file, verbose=False):
                                    and any(~np.isnan(grid.final_values[key]))
                                    and "MODEL" not in key)]
     
+    out_nan_keys = [key for key in grid.final_values.dtype.names
+                    if type(grid.final_values[key][0]) != np.str_
+                    and all(np.isnan(grid.final_values[key]))
+                ]
```
mkruckow: Why did you exclude the check for "model_number" and "MODEL"? The first one is a MESA specific column to exclude the second one are columns from the supernova models, which should not be in there for the interpolator based on the mass transfer.

bin/posydon-run-pipeline
```diff
@@ -407,6 +413,12 @@ def train_interpolators(i, path_to_csv_file, verbose=False):
                                    (type(grid.final_values[key][0]) != np.str_)
                                    and any(~np.isnan(grid.final_values[key]))
                                    and f"S{i}_{MODEL_NAME}" in key)]
+            
+            out_nan_keys = [key for key in grid.final_values.dtype.names
+                            if type(grid.final_values[key][0]) != np.str_
+                            and all(np.isnan(grid.final_values[key]))
+                        ]
```
mkruckow: Here we need to request that the values belong to the supernova models. We don't want to include here stuff which are only keys for the mass transfer classes.

posydon/interpolation/IF_interpolation.py
```diff
@@ -359,7 +359,7 @@ def save(self, filename):
 class BaseIFInterpolator:
     """Class handling the initial-final interpolation for POSYDON."""
 
-    def __init__(self, grid=None, in_keys=None, out_keys=None, in_scaling=None,
+    def __init__(self, grid=None, in_keys=None, out_keys=None, out_nan_keys=None, in_scaling=None,
```
mkruckow, line 362: Have you checked that later functions can handle a None as input for `out_nan_keys`? So far it got initialized as an empty list in case it wasn't set.

maxbriel: This is fine, since it's set to a value in this function, if it's `None`. 
See [line 463.](https://github.com/POSYDON-code/POSYDON/blob/191f056e4fa35e9b63324cf8fffba5dfe03486c0/posydon/interpolation/IF_interpolation.py#L463)

bin/posydon-run-pipeline
maxbriel: fixed

bin/posydon-run-pipeline
maxbriel: fixed

maxbriel:
Test training:

```
from posydon.grids.psygrid import PSyGrid
import numpy as np
from posydon.interpolation.IF_interpolation import IFInterpolator
from posydon.grids.MODELS import MODELS

# example file
interp_path = './CO_HeMS_RLO_1e+00_Zsun.h5'

grid = PSyGrid(interp_path)
interpolation_method = 'linear_RLO'
grid_type = 'CO-HeMS'

method = interpolation_method
if '_RLO' in method:
    method = method.split('_RLO')[0]
    
    
interp_method = [method, method]
interp_classes = ["stable_MT", "unstable_MT"]
if '_RLO' not in interpolation_method:
    interp_method += [method]
    interp_classes += ["no_MT"]
if 'CO' not in grid_type:
    interp_method += [method]
    interp_classes += ["stable_reverse_MT"]

out_keys = [key for key in grid.final_values.dtype.names if (
                                key != "model_number" and
                                (type(grid.final_values[key][0]) != np.str_)
                                and any(~np.isnan(grid.final_values[key]))
                                and "MODEL" not in key)]


out_nan_keys = [key for key in grid.final_values.dtype.names if 
                         (key != "model_number" and
                         (type(grid.final_values[key][0]) != np.str_)
                         and all(np.isnan(grid.final_values[key]))
                         and "MODEL" not in key)
                    ]

c_keys = ['interpolation_class', 'S1_state', 'S2_state', 'mt_history']
        
interpolators = [
    {
        "interp_method": interp_method,
        "interp_classes": interp_classes,
        "out_keys": out_keys,
        "out_nan_keys" : out_nan_keys,
        "class_method": "kNN",
        "c_keys": c_keys,
        "c_key": "interpolation_class"
    },
]
interp = IFInterpolator(grid=grid, interpolators=interpolators)
interp.train()

for i in interp.interpolators:
    print("Out keys:", i.out_keys)
    print("NaN keys:", i.out_nan_keys)
```

posydon/interpolation/IF_interpolation.py
mkruckow: True, my question was, whether you checked, that all function calls are fine.
E.g. there is one in line 446: has this no problems, with changing the type from list to None-type.
After having a look at this function, yes it looks to be OK with this change.

bin/posydon-run-pipeline
```diff
@@ -434,9 +437,7 @@ def train_interpolators(i, path_to_csv_file, verbose=False):
                     "out_keys": out_keys,
                     "out_nan_keys": out_nan_keys,
                     "class_method": "kNN",
-                    "c_keys": [f'S{i}_{MODEL_NAME}_CO_interpolation_class', 
-                               f'S{i}_{MODEL_NAME}_CO_type',
-                               f'S{i}_{MODEL_NAME}_SN_type'],
+                    "c_keys": [f'S{i}_{MODEL_NAME}_CO_interpolation_class',],
```
mkruckow: This reverts a change done in PR #325 :
@maxbriel , could you please comment, why this change need to be reverted?
@ZepeiX , could you please comment, why you needed this change, and whether it is still needed?

maxbriel:
One additional change here:

I've removed the `c_keys` for `f'S{i}_{MODEL_NAME}_SN_type'` and `f'S{i}_{MODEL_NAME}_CO_type'` from when we're training IFinterpolators for each `MODEL_NAME` core-collapse quantities.

The `c_keys` at [line 401](https://github.com/POSYDON-code/POSYDON/blob/191f056e4fa35e9b63324cf8fffba5dfe03486c0/bin/posydon-run-pipeline#L399-L400) already contains the different `f'S{i}_{MODEL_NAME}_CO_type'` and `f'S{i}_{MODEL_NAME}_SN_type'`. 

Since the classifier is independent of the `c_key` class, there is no need to include it again, when we're creating the IFinterpolators for the core-collapse MODELS. These classifiers are already being trained. There's no difference in the training between them.

bin/posydon-run-pipeline
ZepeiX: This was to fix the inconsistency between CO_type and CO_interpolation_class. I think it's still needed. 
Perhaps remove the SN_type and CO_type at line 401 when training for mass transfer cases.

maxbriel: @ZepeiX I don't understand how this can fix the inconsistency after conversations with @philipp-rajah, since the classifiers are independent of the interpolation class and of each other. The interpolator is conditional to `c_key` classes, but the classifiers (`c_keys`) are not. The code seems to indicate this too.
I think it's your other change of the valid classes that fixed it, since that does play an important role in the classifier training.

I will train the interpolators with this branch for an additional check.

bin/posydon-run-pipeline
```diff
@@ -389,6 +396,7 @@ def train_interpolators(i, path_to_csv_file, verbose=False):
             "interp_method": interp_method,
             "interp_classes": interp_classes,
             "out_keys": out_keys,
+            'out_nan_keys': out_nan_keys,
```
mkruckow: Very minor: could you please use double quotes here for consistency?

mkruckow (dismissed):

maxbriel:
I've successfully ran a population with this. The SN_type + CO_state seems to match; also as an affect of the forced check in the evolution we implemented in PR #324. I did notice other issues but not related to this one.

bin/posydon-run-pipeline
ZepeiX: @maxbriel @mkruckow 
Did you see NSs and BHs with wrong masses again with your test?  I did find training the CO_type again gives different results. 
Can you add it back and train the interpolators again and to see if the wrong masses are fixed?

mkruckow: It looks like you are only going with try and error here.
@ZepeiX, please explain, what it does differently by adding this, it looks wrong to us to put it in there again.
@philipp-rajah, could you may have a look at it, too? All this discussion may relate to unclear differences between `c_keys` and `c_key`. Unfortunately, the docstrings only include `c_keys` but don't explain `c_key`.

maxbriel: I would like to add another argument to at least including it in the first training:

`S2_MODEL_CO_type` and `S2_MODEL_SN_type` classifiers are **not** included in the secondary training loop.

Also I am finding some NS with BH masses (4.5 Msun) but only for the detached step. The state is then corrected in `step_END` to a BH. I didn't encounter any WD with wrong masses.

astroJeff (approved):

mkruckow (approved):
We should discuss the interpolation keys later on.

elizabethteng (approved):

PR #344, opened 2024-06-27
Replace scipy.integrate.simps call with scipy.integrate.simpson call

maxbriel:
Replaces the `simps` function call with `simpson`.
This is the only call to this function.

This call has been deprecated in scipy 1.14.0.

No additional version limit is required than our current 1.10.1, since `simpson` is already present in 1.7.0 of scipy.


mkruckow (approved):
Looks fine, the only change I have seen between the old and the new function is the dropping of an optional parameter, which we don't use.

astroJeff (approved):

PR #352, opened 2024-07-09
Use correct PISN mass range

mkruckow:
Use He-core mass range for PISN.
(previously the CO-core mass value was compared to He-core mass)

### DIFF
--- a/posydon/binary_evol/SN/step_SN.py
+++ b/posydon/binary_evol/SN/step_SN.py
@@ -887,7 +887,7 @@ class StepSN(object):
                     # this is the 8th-order polynomial fit of table 1
                     # value, see COSMIC paper (Breivik et al. 2020)
                     polyfit = (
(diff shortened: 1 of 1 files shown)


maxbriel (approved):
Just clarifying, the upper limit here comes from the direct collapse value in Marchant et al. (2019), while the lower limit is from the PPI fit Breivik et al. (2020) performed?

mkruckow:
> Just clarifying, the upper limit here comes from the direct collapse value in Marchant et al. (2019), while the lower limit is from the PPI fit Breivik et al. (2020) performed?

All the limits are in the Marchant et al. (2019) just taking the edge simulations in table 1. Breivik et al. (2020) quoted the ranges (unfortunately switched there between He core and Co core mass). The only think new in Breivik is the polynomial fit.

astroJeff (approved):

PR #357, opened 2024-07-12
replace np.NaN and np.NAN by np.nan

mkruckow:
It looks like that the [development version](https://numpy.org/devdocs/reference/constants.html#numpy.nan) of numpy doesn't support the aliases for nan anymore.

Hence, we need to only use `np.nan` in the future.
- [x] replace `np.NaN`
- [x] replace `np.NAN`


ezapartas (approved):

maxbriel (approved):
Looks like all of them!

astroJeff (approved):

PR #358, opened 2024-07-14
Fixing a small typo in grids IO module.

kkovlakas:
Correcting a copy-paste mistake, possibly from the PRs related to `PSyGrid` reading zipped MESA files.

The path of the MESA model (`*.mod`) for *star 2* was stored as if it was *star 1*'s model path. Consequently, the information was lost, and depending on the order of reading the MESA files, substituting the correct path to *star 1*'s model.


mkruckow (approved):
Thanks for clarifying, that it is a typo and not put in by intention to swap the definitions of star1 and star2.

astroJeff (approved):

PR #365, opened 2024-07-30
co_core_mass option added

dimsour94:
Just added the co_core_mass option for the matching.


ezapartas (approved):

mkruckow:
Is that tested?
I guess, it will may cause issue for stars, where the CO core is not well defined, e.g. low mass or less evolved stars.

On the other hand, this more looks like a list of allowed keys. Why do have a limitation here in the first place and not allow all keys, we have in the grid?

ezapartas:
for now, covered in #311 . Allowing for all keys may be an option, if we are sure that there is no  reason to be there in the first place

PR #371, opened 2024-08-19
Fix RLO warnings in detached step

celiotine:
This PR adds catching for invalid inputs to four highly used RLO functions in the detached step. With the POSYDON warnings PR, python warnings are no longer suppressed when running binary populations, and these functions are throwing a lot of python warnings (for divide-by-zero, etc.) that should not occur. Thus the error files for running populations are large without this fix.
This PR should not be accepted until the POSYDON warnings PR is accepted, so that the warnings added here can be correctly integrated into the code with the new warnings framework.

### DIFF
--- a/posydon/utils/common_functions.py
+++ b/posydon/utils/common_functions.py
@@ -207,15 +207,15 @@ RL -> Roche lobe radius in similar units as a_orb
 '''
 
 
-def roche_lobe_radius(q, a_orb=1):
+def roche_lobe_radius(m1, m2, a_orb=1):
     """Approximate the Roche lobe radius from [1]_.
 
     Parameters
     ----------
-    q : float
-        Dimensionless mass ratio = MRL/Mcomp, where
-        MRL is the mass of the star we calculate the RL and
-        Mcomp is the mass of its companion star.
+    m1 : float
+        the mass of the star for which we calculate the Roche lobe
+    m2: float
+        the mass of the companion star
     a_orb : float
         Orbital separation. The return value will have the same unit.
 
@@ -228,6 +228,20 @@ def roche_lobe_radius(q, a_orb=1):
     .. [1] Eggleton, P. P. 1983, ApJ, 268, 368
 
     """
+
+    if isinstance(a_orb, np.ndarray):
+        Pwarn("Trying to compute RL radius for binary with no separation", "EvolutionWarning")
+        a_orb = np.nan
+                          
+    if m1 <=0:
+        Pwarn("Trying to compute RL radius for binary with nonexistent companion", "EvolutionWarning")
+        m1 = np.nan
+
+    if m2 <=0:
+        Pwarn("Trying to compute RL radius for binary with nonexistent companion", "EvolutionWarning")
+        m2 = np.nan
+
+    q = m1/m2
     RL = a_orb * (0.49 * q**(2. / 3.)) / (
         0.6 * q**(2. / 3.) + np.log(1 + q**(1. / 3))
     )
--- a/posydon/utils/common_functions.py
+++ b/posydon/utils/common_functions.py
@@ -195,52 +195,54 @@ def rzams(m, z=0.02, Zsun=0.02):
     return r
 
 
-'''
-
-
-Receives:
-q ->
-a_orb ->
-
-Returns:
-RL -> Roche lobe radius in similar units as a_orb
-'''
-
-
 def roche_lobe_radius(m1, m2, a_orb=1):
     """Approximate the Roche lobe radius from [1]_.
 
     Parameters
     ----------
-    m1 : float
+    m1 : float, array of floats
         the mass of the star for which we calculate the Roche lobe
-    m2: float
+    m2 : float, array of floats
         the mass of the companion star
-    a_orb : float
+    a_orb : float, array of floats
         Orbital separation. The return value will have the same unit.
 
     Returns
     -------
-    float
+    float, array of floats
         Roche lobe radius in similar units as a_orb
     References
     ----------
     .. [1] Eggleton, P. P. 1983, ApJ, 268, 368
 
     """
-
-    if isinstance(a_orb, np.ndarray):
+    
+    ## catching if a_orb is an empty array or is an array with invalid separation values
+    if isinstance(a_orb, np.ndarray) and (not np.any(a_orb) or np.any(a_orb <=0)):
+        Pwarn("Trying to compute RL radius for binary with no separation", "EvolutionWarning")
+        a_orb = np.full_like(a_orb, np.nan)
+    ## catching if a_orb is a float with invalid separation value
+    elif a_orb <=0: 
         Pwarn("Trying to compute RL radius for binary with no separation", "EvolutionWarning")
         a_orb = np.nan
-                          
-    if m1 <=0:
-        Pwarn("Trying to compute RL radius for binary with nonexistent companion", "EvolutionWarning")
-        m1 = np.nan
 
-    if m2 <=0:
-        Pwarn("Trying to compute RL radius for binary with nonexistent companion", "EvolutionWarning")
+
+    if isinstance(m1, np.ndarray) and (not np.any(m1) or np.any(m1 <=0)):                    
+        Pwarn("Trying to compute RL radius for nonexistent object", "EvolutionWarning")
+        m1 = np.full_like(m1, np.nan)
+    elif m1 <=0:
+        Pwarn("Trying to compute RL radius for nonexistent object", "EvolutionWarning")
+        m1 = np.nan
+    
+
+    if isinstance(m2, np.ndarray) and (not np.any(m2) or np.any(m2 <=0)):                    
+        Pwarn("Trying to compute RL radius for nonexistent companion", "EvolutionWarning")
+        m2 = np.full_like(m2, np.nan)
+    elif m2 <=0:
+        Pwarn("Trying to compute RL radius for nonexistent companion", "EvolutionWarning")
         m2 = np.nan
 
+    
     q = m1/m2
     RL = a_orb * (0.49 * q**(2. / 3.)) / (
         0.6 * q**(2. / 3.) + np.log(1 + q**(1. / 3))
--- a/posydon/utils/common_functions.py
+++ b/posydon/utils/common_functions.py
@@ -200,16 +200,16 @@ def roche_lobe_radius(m1, m2, a_orb=1):
 
     Parameters
     ----------
-    m1 : float, array of floats
+    m1 : float, ndarray of floats
         the mass of the star for which we calculate the Roche lobe
-    m2 : float, array of floats
+    m2 : float, ndarray of floats
         the mass of the companion star
-    a_orb : float, array of floats
+    a_orb : float, ndarray of floats
         Orbital separation. The return value will have the same unit.
 
     Returns
     -------
-    float, array of floats
+    float, ndarray of floats
         Roche lobe radius in similar units as a_orb
     References
     ----------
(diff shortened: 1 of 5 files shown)


astroJeff:
Update (Aug 22): We discussed this PR. There are a couple ways it could possibly be streamlined: we suggested moving the checks into the common_functions.py routine, rather than here.

posydon/utils/common_functions.py
```diff
@@ -228,6 +228,20 @@ def roche_lobe_radius(q, a_orb=1):
     .. [1] Eggleton, P. P. 1983, ApJ, 268, 368
 
     """
+
+    if isinstance(a_orb, np.ndarray):
+        Pwarn("Trying to compute RL radius for binary with no separation", "EvolutionWarning")
```
mkruckow: I'd more go for a `InappropriateValueWarning` here. You don't know, whether this function will only be called when evolving a binary.
Additionally, I don't see a good reason to set here an input value to `nan`. This will cause the output to be always `nan`, hence we can directly return a `nan`.

posydon/utils/common_functions.py
```diff
@@ -207,15 +207,15 @@ def rzams(m, z=0.02, Zsun=0.02):
...
 
     Parameters
     ----------
-    q : float
-        Dimensionless mass ratio = MRL/Mcomp, where
-        MRL is the mass of the star we calculate the RL and
-        Mcomp is the mass of its companion star.
+    m1 : float
+        the mass of the star for which we calculate the Roche lobe
+    m2: float
```
mkruckow: There is a space missing between `m2` and the colon.

posydon/utils/common_functions.py
```diff
@@ -228,6 +228,20 @@ def roche_lobe_radius(q, a_orb=1):
     .. [1] Eggleton, P. P. 1983, ApJ, 268, 368
 
     """
+
+    if isinstance(a_orb, np.ndarray):
```
mkruckow, line 221: Shouldn't we better check whether a variable is not a `float`? (same applies to `m1` ans `m2`) Shell we disallow negative orbital separations?

celiotine: Checking if m1 or m2 are floats causes problems because sometimes they are passed to the function as np.ndarray and sometimes they are np.float64. If we restrict to only floats then integration fails in the detached step later on.

celiotine: If we want to be strict about keeping m1, m2, and a_orb to float types only, then more changes need to be made in other places in the code, because currently they all have inconsistent data types.

celiotine: Or, we can expand the function to allow for floats and arrays of floats, and then change the type checking accordingly

mkruckow: Yes, we can take floats and numpy-arrays of floats, but this needs then to be represented in:

- the checking (not included yet); if we allow arrays we need to ensure, that the arrays have the same size, because we need a unique size for the output
- the docstring (states to allow floats only)

The more important thing is that the docstring needs to be correct about outputs. Hence, allowing numpy arrays as input will cause the output to be a numpy array.
We should think about, whether we'd like to allow a mixture e.g. m1 and m2 numpy arrays and a_orb being a float.

posydon/utils/common_functions.py
```diff
@@ -228,6 +228,20 @@ def roche_lobe_radius(q, a_orb=1):
...
+        Pwarn("Trying to compute RL radius for binary with no separation", "EvolutionWarning")
+        a_orb = np.nan
+                          
+    if m1 <=0:
```
astroJeff: try this:

`if np.any(np.atleast_1d(m1) <= 0):`

posydon/utils/common_functions.py
celiotine: Currently a mixture of data types are being passed to this function, and the calculation of RL is handling the mixed-types accordingly. In the detached step, m1, m2, and a_orb are coming from the interpolator, so I think it is easiest to allow for mixed types, unless someone wants to standardize the outputs of the interpolator.

posydon/utils/common_functions.py
```diff
@@ -195,52 +195,54 @@ def rzams(m, z=0.02, Zsun=0.02):
...
-'''
-
-
 def roche_lobe_radius(m1, m2, a_orb=1):
     """Approximate the Roche lobe radius from [1]_.
 
     Parameters
     ----------
-    m1 : float
+    m1 : float, array of floats
```
mkruckow: Could you please use the term `ndarray` instead of `array`, because python does not have a real `array` type. Some code parts already use the correct terminology.
P.S.: But there are some others using the wrong one, I guess that's something to correct in a different PR.

posydon/utils/common_functions.py
```diff
@@ -195,52 +195,54 @@ def rzams(m, z=0.02, Zsun=0.02):
...
     References
     ----------
     .. [1] Eggleton, P. P. 1983, ApJ, 268, 368
 
     """
-
-    if isinstance(a_orb, np.ndarray):
+    
+    ## catching if a_orb is an empty array or is an array with invalid separation values
+    if isinstance(a_orb, np.ndarray) and (not np.any(a_orb) or np.any(a_orb <=0)):
```
mkruckow: Can you remind me, what `not np.any(a_orb)` is good for? Hence, what is covered by that, that's not covered by `np.any(a_orb <=0)`? I guess, you look for an empty array, but there you should use `len(a_orb)==0`, which should be more efficient because it doesn't need to touch all the values stored in the ndarray, additionally `np.any` would fail if 0 would be an allowed value.
Instead, we should have a check on the dtype of the ndarray.

posydon/utils/common_functions.py
```diff
@@ -195,52 +195,54 @@ def rzams(m, z=0.02, Zsun=0.02):
...
+        Pwarn("Trying to compute RL radius for binary with no separation", "EvolutionWarning")
+        a_orb = np.full_like(a_orb, np.nan)
```
mkruckow: If you like to only set input values to nan. I'd suggest to only set the inappropriate ones to nan, hence if there is only one bad value in an array, you can get all other values correctly calculated.

posydon/utils/common_functions.py
celiotine: Yes, np.any() will check for an empty array. I would not use len() in this case because it will return the number of dimensions for a numpy array if the array has more than one dimension, which is not what we want. 
If the entries of a_orb are zero, that should be caught by the second condition.

celiotine: I think a_orb.size will do this check better, I am changing to that

mkruckow: Yes, using the `size` is what I intended when suggesting `len`. `np.any` has always the issue, that is returns `False` on both an empty ndarray and an ndarray containing only values which are convertible to False individually.

posydon/utils/common_functions.py
```diff
@@ -218,26 +218,40 @@ def roche_lobe_radius(m1, m2, a_orb=1):
...
-    if isinstance(a_orb, np.ndarray) and (not np.any(a_orb) or np.any(a_orb <=0)):
-        Pwarn("Trying to compute RL radius for binary with no separation", "EvolutionWarning")
-        a_orb = np.full_like(a_orb, np.nan)
+    if isinstance(a_orb, np.ndarray):
+        ## if array is empty (or all zeros), fill with NaN values
+        if a_orb.size == 0:
+            Pwarn("Trying to compute RL radius for binary with invalid separation", "EvolutionWarning")
+            a_orb = np.full_like(a_orb, np.nan, dtype=np.float64)
+        ## if array contains invalid values, replace with NaN 
+        elif np.any(a_orb < 0):
```
mkruckow, line 227: Here we should use `a_orb <= 0`. The same two lines later.
(Alternatively we should use <0 for floats as well, in case we'd like to allow for a 0 separation. But I think we don't want that.)

posydon/utils/common_functions.py
```diff
@@ -218,26 +218,40 @@ def roche_lobe_radius(m1, m2, a_orb=1):
...
     
     ## catching if a_orb is an empty array or is an array with invalid separation values
-    if isinstance(a_orb, np.ndarray) and (not np.any(a_orb) or np.any(a_orb <=0)):
-        Pwarn("Trying to compute RL radius for binary with no separation", "EvolutionWarning")
-        a_orb = np.full_like(a_orb, np.nan)
+    if isinstance(a_orb, np.ndarray):
+        ## if array is empty (or all zeros), fill with NaN values
+        if a_orb.size == 0:
+            Pwarn("Trying to compute RL radius for binary with invalid separation", "EvolutionWarning")
+            a_orb = np.full_like(a_orb, np.nan, dtype=np.float64)
```
mkruckow: This isn't doing anything, because an empty ndarray stays empty (the same applies to `m1` and `m2`). If you like to get np.nan values in there you'd need to change the shape to make all dimensions with 0 shape to become a shape 1.
e.g. `a_orb = np.full([1 if s==0 else s for s in a_orb.shape], np.nan, dtype=np.float64)`

posydon/utils/common_functions.py
```diff
@@ -195,39 +195,69 @@ def rzams(m, z=0.02, Zsun=0.02):
...
+            m2 = np.full_like(m2, np.nan, dtype=np.float64)
+        elif np.any(m2 <= 0):
+            Pwarn("Trying to compute RL radius for nonexistent companion", "EvolutionWarning")
+            m2[m2 <= 0] = np.nan
+    elif m2 <=0:
+        Pwarn("Trying to compute RL radius for nonexistent companion", "EvolutionWarning")
+        m2 = np.nan
+
+    
+    q = m1/m2
```
mkruckow, line 259: Should we check for compatible shapes before combining to objects? Otherwise we risk numpy errors.

posydon/utils/common_functions.py
celiotine: We can allow for the separation to be zero, it will just cause RL to be zero

posydon/utils/common_functions.py
celiotine: We want those errors to be thrown so that people know they are using inputs of different shapes.

posydon/utils/common_functions.py
celiotine: If the shape is empty (not zero) the size will be greater than zero, so the fix I have implemented should work.

mkruckow: But here we are behind a `a_orb.size == 0`, thus the ndarray is empty. But there is another strange thing in the latest changes, see other comment.

posydon/utils/common_functions.py
mkruckow: My request was to turn the errors into warnings not to suppress them.

posydon/utils/common_functions.py
```diff
@@ -222,7 +223,7 @@ def roche_lobe_radius(m1, m2, a_orb=1):
         ## if array is empty, fill with NaN values
         if a_orb.size == 0:
             Pwarn("Trying to compute RL radius for binary with invalid separation", "EvolutionWarning")
-            a_orb = np.full_like(a_orb, np.nan, dtype=np.float64)
+            a_orb = np.full_like(a_orb.shape, np.nan, dtype=np.float64)
```
mkruckow: This won't work for ndarrays of and combination of dimensions:
`np.full_like` needs an array like object as first argument, while `np.full` takes a shape as first argument.

posydon/utils/common_functions.py
celiotine: Please see the changes that have been made in the code before commenting on threads that have been resolved.

posydon/utils/common_functions.py
celiotine: That was not made clear in you comment. We should let those errors be thrown by python to prevent broken code from causing more errors down the line.

posydon/utils/common_functions.py
celiotine: The code I have written does work, I have tested it, and the code you have suggested does not work. Please see a screenshot below as to why if you are struggling to understand:

Screen Shot 2024-09-07 at 8 40 08 PM

celiotine: This does work for certain arrays (namely np.array([]) objects) because shape itself is array-like. I realized it does not work for arrays with more dimensions, please see most recent changes for that.

posydon/utils/common_functions.py
mkruckow: The consequence of this will be that we can never remove the dirty try statements in the code. Hence, we can't differentiate between an error caused by a bug in our code and an error caused by passing inappropriate objects to imported modules, which aren't written by us.
If we'd like to limit us in this way, OK.

mkruckow (approved):
Looks good to me.

astroJeff (approved):

PR #374, opened 2024-08-28
TransientPopulation overplot crash fix

maxbriel:
When no binary is present in met_indices, the selection function crashes.

This checks if the array is empty and returns an empty Dataframe with the appropriate columns.
With this fix, you're able to plot empty slices.

### DIFF
--- a/posydon/visualization/plot_pop.py
+++ b/posydon/visualization/plot_pop.py
@@ -247,10 +247,10 @@ def plot_popsyn_over_grid_slice(pop, grid_type, met_Zsun, slices=None, channel=N
         if 'HMS-HMS' in grid_path:
             slice_3D_var_range = (dq_edges[i],dq_edges[i+1])
             if len(met_indices) == 0:
-                data = pd.DataFrame(columns=['S1_mass','S2_mass' ,'orbital_period'])
+                data = pd.DataFrame(columns=['S1_mass','S2_mass', 'orbital_period'])
             else:
                 sel = 'index in '+str(met_indices)+' & event == "ZAMS"'
-                data = pop.history.select(where=sel, columns=['S1_mass','S2_mass' ,'orbital_period'])
+                data = pop.history.select(where=sel, columns=['S1_mass','S2_mass', 'orbital_period'])
 
             q = data['S2_mass'].values/data['S1_mass'].values
             q[q>1] = 1./q[q>1]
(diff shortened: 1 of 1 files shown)


posydon/visualization/plot_pop.py
```diff
@@ -245,9 +246,11 @@ def plot_popsyn_over_grid_slice(pop, grid_type, met_Zsun, slices=None, channel=N
         met_indices = np.where(pop.select(where=channel_sel, columns=['metallicity']) == met_Zsun)[0].tolist()
         if 'HMS-HMS' in grid_path:
             slice_3D_var_range = (dq_edges[i],dq_edges[i+1])
-
-            sel = 'index in '+str(met_indices)+' & event == "ZAMS"'
-            data = pop.history.select(where=sel, columns=['S1_mass','S2_mass' ,'orbital_period'])
+            if len(met_indices) == 0:
+                data = pd.DataFrame(columns=['S1_mass','S2_mass' ,'orbital_period'])
```
mkruckow: Wouldn't this be a good place to add a warning, that there is no data to plot? (the same for the other two cases)

posydon/visualization/plot_pop.py
```diff
@@ -245,9 +246,11 @@ def plot_popsyn_over_grid_slice(pop, grid_type, met_Zsun, slices=None, channel=N
...
+            else:
+                sel = 'index in '+str(met_indices)+' & event == "ZAMS"'
```
mkruckow, line 252: How well is the select to deal with long statements? I was wondering, whether we shouldn't hard code the selection in the binary evolution but instead call the flow chart to extract all cases, where a binary will get send to a step with a given binary grid?

posydon/visualization/plot_pop.py
```diff
@@ -245,9 +246,11 @@ def plot_popsyn_over_grid_slice(pop, grid_type, met_Zsun, slices=None, channel=N
...
+                data = pop.history.select(where=sel, columns=['S1_mass','S2_mass' ,'orbital_period'])
```
mkruckow: Just style: is there a reason for the space behind `'S2_mass'`? Would it be useful to define a common column set once and always use this?

posydon/visualization/plot_pop.py
mkruckow: Additionally, do we still want the plotting or should we skip it (just adding a `continue` instead of creating a new data frame), like `# TODO: skip plotting slice if there are no data` suggests?

maxbriel: I find it useful to see the slice even if no binaries are on it, since this helps with verification.

posydon/visualization/plot_pop.py
maxbriel: This is a good question. Especially with the other two selections, I'm not sure they're always correct. I have not changed the type of selection in this PR nor in the previous PR that touched this code.

posydon/visualization/plot_pop.py
mkruckow: OK, please check, whether the other condition `if slices is not None and round(var,2) not in np.around(slices,2)`, should be modified, to produce a plot there, too.

posydon/visualization/plot_pop.py
mkruckow: You were too quick: now there is a space behind the second comma but not behind the first one, why?

astroJeff (approved):
This fixes one bug, and uncovers another: we should not necessarily hard code the `event` for each of the grids - they should be taken from the flow, regardless of any user's flowchart. That has been added as a new issue to be dealt with in the future.

posydon/visualization/plot_pop.py
mkruckow: Please open an issue in case you don't want to address this here.

kasdaglie (approved):

PR #376, opened 2024-08-29
New normalized_pop_mass.py file

dimsour94:
The new normalized_pop_mass.py file estimates the underlying mass of a population based on your simulated binary fraction (as defined in your ini file) and any specified desired binary fraction you want the "nature" population to have.

### DIFF
--- a/posydon/popsyn/binarypopulation.py
+++ b/posydon/popsyn/binarypopulation.py
@@ -531,7 +531,7 @@ class BinaryPopulation:
                     
                     simulated_mass_binaries += np.nansum(filtered_data_binaries[["S1_mass_i", "S2_mass_i"]].to_numpy())
                     simulated_mass_single += np.nansum(filtered_data_single[["S1_mass_i"]].to_numpy())
-                    simulated_mass += simulated_mass_single + simulated_mass_binaries
+                    simulated_mass = simulated_mass_single + simulated_mass_binaries
                        
                     if 'metallicity' not in oneline.columns:
                         met_df = pd.DataFrame(data={'metallicity': [self.metallicity] * len(oneline)}, index=oneline.index)
--- a/posydon/popsyn/binarypopulation.py
+++ b/posydon/popsyn/binarypopulation.py
@@ -510,9 +510,10 @@ class BinaryPopulation:
         
         
         with pd.HDFStore(absolute_filepath, mode=mode, complevel=complevel, complib=complib) as store:
+            simulated_mass = 0.0
+            simulated_mass_single = 0.0
+            simulated_mass_binaries = 0.0
             number_of_systems=0
-            
-            
             
             for f in file_names:
                 # strings itemsize set by first append max value,
@@ -522,18 +523,15 @@ class BinaryPopulation:
                                  min_itemsize=history_min_itemsize)
                     
                     oneline = pd.read_hdf(f, key='oneline')
-                    if "initially_single_star" in oneline["state_i"].values:
-                        mask= oneline["state_i"]=="initially_single_star"
-                        filtered_data_single = oneline[mask]
-                        filtered_data_binaries = oneline[~mask]
-    
-                        simulated_mass_binaries = np.sum(filtered_data_binaries[["S1_mass_i", "S2_mass_i"]].to_numpy())
-                        simulated_mass_single = np.sum(filtered_data_single[["S1_mass_i"]].to_numpy())
-                        simulated_mass = simulated_mass_binaries + simulated_mass_single
-                    else:
-                        simulated_mass_binaries = np.sum(oneline[["S1_mass_i", "S2_mass_i"]].to_numpy())
-                        simulated_mass_single = 0.0
-                        simulated_mass = simulated_mass_binaries
+                    
+                    # split weight between single and binary stars
+                    mask = oneline["state_i"] == "initially_single_star"
+                    filtered_data_single = oneline[mask]
+                    filtered_data_binaries = oneline[~mask]
+                    
+                    simulated_mass_binaries += np.nansum(filtered_data_binaries[["S1_mass_i", "S2_mass_i"]].to_numpy())
+                    simulated_mass_single += np.nansum(filtered_data_single[["S1_mass_i"]].to_numpy())
+                    simulated_mass += simulated_mass_single + simulated_mass_binaries
                        
                     if 'metallicity' not in oneline.columns:
                         met_df = pd.DataFrame(data={'metallicity': [self.metallicity] * len(oneline)}, index=oneline.index)
@@ -556,7 +554,8 @@ class BinaryPopulation:
             tmp_df = pd.DataFrame(
                 index=[self.metallicity],
                 data={'simulated_mass': simulated_mass,
-                      'underlying_mass': initial_total_underlying_mass(df=simulated_mass,df1=simulated_mass_single, df2= simulated_mass_binaries, **self.kwargs)[0], 
+                      'simulated_mass_single': simulated_mass_single,
+                      'simulated_mass_binaries': simulated_mass_binaries,
                       'number_of_systems': number_of_systems})
             tmp_df.index.name = 'metallicity'
             store.append('mass_per_metallicity', tmp_df)
--- a/posydon/popsyn/binarypopulation.py
+++ b/posydon/popsyn/binarypopulation.py
@@ -510,8 +510,10 @@ class BinaryPopulation:
         
         
         with pd.HDFStore(absolute_filepath, mode=mode, complevel=complevel, complib=complib) as store:
-            simulated_mass = 0
-            number_of_systems = 0
+            number_of_systems=0
+            
+            
+            
             for f in file_names:
                 # strings itemsize set by first append max value,
                 # which may not be largest string
@@ -520,7 +522,19 @@ class BinaryPopulation:
                                  min_itemsize=history_min_itemsize)
                     
                     oneline = pd.read_hdf(f, key='oneline')
-                    simulated_mass += oneline['S1_mass_i'].sum() + oneline['S2_mass_i'].sum()
+                    if "initially_single_star" in oneline["state_i"].values:
+                        mask= oneline["state_i"]=="initially_single_star"
+                        filtered_data_single = oneline[mask]
+                        filtered_data_binaries = oneline[~mask]
+    
+                        simulated_mass_binaries = np.sum(filtered_data_binaries[["S1_mass_i", "S2_mass_i"]].to_numpy())
+                        simulated_mass_single = np.sum(filtered_data_single[["S1_mass_i"]].to_numpy())
+                        simulated_mass = simulated_mass_binaries + simulated_mass_single
+                    else:
+                        simulated_mass_binaries = np.sum(oneline[["S1_mass_i", "S2_mass_i"]].to_numpy())
+                        simulated_mass_single = 0.0
+                        simulated_mass = simulated_mass_binaries
+                       
                     if 'metallicity' not in oneline.columns:
                         met_df = pd.DataFrame(data={'metallicity': [self.metallicity] * len(oneline)}, index=oneline.index)
                         oneline = pd.concat([oneline, met_df], axis=1)
@@ -542,7 +556,7 @@ class BinaryPopulation:
             tmp_df = pd.DataFrame(
                 index=[self.metallicity],
                 data={'simulated_mass': simulated_mass,
-                      'underlying_mass': initial_total_underlying_mass(df=simulated_mass, **self.kwargs)[0], 
+                      'underlying_mass': initial_total_underlying_mass(df=simulated_mass,df1=simulated_mass_single, df2= simulated_mass_binaries, **self.kwargs)[0], 
                       'number_of_systems': number_of_systems})
             tmp_df.index.name = 'metallicity'
             store.append('mass_per_metallicity', tmp_df)
--- a/posydon/popsyn/normalized_pop_mass.py
+++ b/posydon/popsyn/normalized_pop_mass.py
@@ -5,6 +5,8 @@ __authors__ = [
     "Devina Misra <devina.misra@unige.ch>",
     "Simone Bavera <Simone.Bavera@unige.ch>",
     "Konstantinos Kovlakas <Konstantinos.Kovlakas@unige.ch>",
+    "Matthias Kruckow <Matthias.Kruckow@unige.ch>",
+    "Dimitris Souropanis <dsouropanis@ia.forth.gr>",
 ]
 
 
--- a/posydon/popsyn/normalized_pop_mass.py
+++ b/posydon/popsyn/normalized_pop_mass.py
@@ -19,9 +19,8 @@ def initial_total_underlying_mass(df=None, df1=None, df2=None, **kwargs):
 
     Parameters
     ----------
-    df : DataFrame
-        Data frame of a population class. If nothing is provided the code will
-        sample initial conditions.
+    df,df1,df2 : float
+        Total simulated mass, simulated mass of binary systems and simulated mass of single stars, respectively.
     primary_mass_min : type
         Description of parameter `primary_mass_min`.
     primary_mass_max : type
@@ -49,13 +48,11 @@ def initial_total_underlying_mass(df=None, df1=None, df2=None, **kwargs):
         initial_ZAMS_mass_2 = initial_ZAMS_mass[3]
         initial_ZAMS_TOTAL_mass = (sum(initial_ZAMS_mass_1)
                                    + sum(initial_ZAMS_mass_2))
-    elif isinstance(df, float):
-        initial_ZAMS_TOTAL_mass = df
-        initial_ZAMS_TOTAL_single=df1
-        initial_ZAMS_TOTAL_binaries=df2
     else:
-        sel = df['event'] == 'ZAMS'
-        initial_ZAMS_TOTAL_mass = sum(df['S1_mass'][sel]+df['S2_mass'][sel])
+        initial_ZAMS_TOTAL_mass = df
+        initial_ZAMS_TOTAL_single= df1
+        initial_ZAMS_TOTAL_binaries= df2
+    
 
     def imf_part_1(m, m_min, alpha1):
         return (m/m_min)**-alpha1
--- a/posydon/popsyn/normalized_pop_mass.py
+++ b/posydon/popsyn/normalized_pop_mass.py
@@ -5,17 +5,14 @@ __authors__ = [
     "Devina Misra <devina.misra@unige.ch>",
     "Simone Bavera <Simone.Bavera@unige.ch>",
     "Konstantinos Kovlakas <Konstantinos.Kovlakas@unige.ch>",
-    "Matthias Kruckow <Matthias.Kruckow@unige.ch>",
 ]
 
 
-import numpy as np
-from posydon.utils.posydonwarning import Pwarn
 from posydon.popsyn import independent_sample
 from scipy.integrate import quad
 
 
-def initial_total_underlying_mass(df=None, **kwargs):
+def initial_total_underlying_mass(df=None, df1=None, df2=None, **kwargs):
     """Compute the initial total mass of the population.
 
     Parameters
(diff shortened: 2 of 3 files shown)


posydon/popsyn/normalized_pop_mass.py
```diff
@@ -107,30 +108,74 @@ def fraction_simulated(f_bin, m_1, m_2, m_min, m_max,
         alpha2 = 2.35
         alpha3 = 2.35
     else:
-        Pwarn("Scheme not included yet: primary_mass_scheme="
-              f"{kwargs['primary_mass_scheme']}, secondary_mass_scheme"
-              f"={kwargs['secondary_mass_scheme']}", "UnsupportedModelWarning")
-        return np.nan, np.nan, np.nan
-#        raise ValueError("Scheme not included yet")
+        raise ValueError("Scheme not included yet")
```
mkruckow: I'd prefer to keep a warning here. Because there might be use cases where you don't care about the underlying mass for a different model. But raising an error, will make it impossible to use data without underlying mass, because the code stops here.

astroJeff:
Update (Aug 29): This looks like a necessary and important improvement - needs to be checked carefully.

maxbriel:
@dimsour94 asked me how we should pass the f_bin_nature to the calculation.
I want to make the following suggestion:
1. When creating the main metadata when concatenating the files, we remove the calculation. Thus not having an underlying mass in the output.
2. We add a function to the population class that will calculate the underlying mass for a given input: f_bin_nature, m_1_max, and m_2_max. If no input is given the defaults are used.

posydon/popsyn/synthetic_population.py
```diff
@@ -992,16 +992,38 @@ def __init__(
...
 
-            simulated_mass = np.sum(self.oneline[["S1_mass_i", "S2_mass_i"]].to_numpy())
+            try:
+        # Filter the data based on the condition
+                filtered_data = self.oneline[self.oneline["state_i"] == "initially_single_star"]
+                filtered_data1 = self.oneline[self.oneline["state_i"] != "initially_single_star"]
+    
+        # Calculate the simulated masses
+                simulated_mass_binaries = np.sum(filtered_data1[["S1_mass_i", "S2_mass_i"]].to_numpy())
+                simulated_mass_single = np.sum(filtered_data[["S1_mass_i", "S2_mass_i"]].to_numpy())
```
mkruckow: Is "S2_mass_i" always well defined for an initial single star? Might it be better to only sum over "S1_mass_i" for single stars?

posydon/popsyn/synthetic_population.py
```diff
@@ -992,16 +992,38 @@ def __init__(
...
+                simulated_mass = np.sum(self.oneline[["S1_mass_i", "S2_mass_i"]].to_numpy())
```
mkruckow: Wouldn't it be more efficient to simply add the `simulated_mass_binaries` and `simulated_mass_single` instead of summing the full dataset again?

posydon/popsyn/synthetic_population.py
```diff
@@ -992,16 +992,38 @@ def __init__(
                 Pwarn(f"{filename} already contains a mass_per_metallicity "
                       "table. Overwriting the table!", "OverwriteWarning")
 
-            simulated_mass = np.sum(self.oneline[["S1_mass_i", "S2_mass_i"]].to_numpy())
+            try:
+        # Filter the data based on the condition
+                filtered_data = self.oneline[self.oneline["state_i"] == "initially_single_star"]
+                filtered_data1 = self.oneline[self.oneline["state_i"] != "initially_single_star"]
```
mkruckow: Why don't you call them `filtered_data_binaries` and `filtered_data_single` as you do later for the masses?

mkruckow: Additionally, do we need to duplicate all the data here? Wouldn't it be sufficient to save only a mask and using this and it's inverse later on?

posydon/popsyn/synthetic_population.py
```diff
@@ -992,16 +992,38 @@ def __init__(
...
+    
+        # Calculate the simulated masses
+                simulated_mass_binaries = np.sum(filtered_data1[["S1_mass_i", "S2_mass_i"]].to_numpy())
+                simulated_mass_single = np.sum(filtered_data[["S1_mass_i", "S2_mass_i"]].to_numpy())
+                simulated_mass = np.sum(self.oneline[["S1_mass_i", "S2_mass_i"]].to_numpy())
+    
+            except TypeError as e:
```
mkruckow: This exception looks odd to me. What do you try to catch with it and wouldn't it be possible to turn it into a check with two options, hence a `if-else` instead of `try-except`?

posydon/popsyn/synthetic_population.py
```diff
@@ -992,16 +992,38 @@ def __init__(
...
+                filtered_data1 = self.oneline[(self.oneline["state_i"] != "initially_single_star")]
+                simulated_mass_binaries = np.sum(filtered_data1[["S1_mass_i", "S2_mass_i"]].to_numpy())
+                simulated_mass_single = None
+                simulated_mass = np.sum(self.oneline[["S1_mass_i", "S2_mass_i"]].to_numpy())
+                print(1)
+            
             underlying_mass = initial_total_underlying_mass(
-                df=simulated_mass, **self.ini_params
-            )[0]
+            df=simulated_mass, df1=simulated_mass_single, df2= simulated_mass_binaries, **self.ini_params
```
mkruckow: According to the function's doc-string `df` should be a pandas data frame, but you pass here a float.

posydon/popsyn/normalized_pop_mass.py
```diff
@@ -6,16 +6,15 @@
...
 
 
-import numpy as np
-from posydon.utils.posydonwarning import Pwarn
 from posydon.popsyn import independent_sample
 from scipy.integrate import quad
 
 
-def initial_total_underlying_mass(df=None, **kwargs):
+def initial_total_underlying_mass(df=None, df1=None, df2=None, **kwargs):
```
mkruckow: You should add df1 and df2 to the doc-string.

posydon/popsyn/normalized_pop_mass.py
```diff
@@ -52,6 +51,8 @@ def initial_total_underlying_mass(df=None, **kwargs):
                                    + sum(initial_ZAMS_mass_2))
     elif isinstance(df, float):
         initial_ZAMS_TOTAL_mass = df
+        initial_ZAMS_TOTAL_single=df1
+        initial_ZAMS_TOTAL_binaries=df2
     else:
```
mkruckow, line 87: You need to handle `df`, `df1`, and `df2` separately there is no guarantee that they will always have the same type.

posydon/popsyn/normalized_pop_mass.py
```diff
@@ -65,28 +66,28 @@ def imf_part_2(m, m_1, m_min, alpha1, alpha2):
...
+        mean_mass = f0 * (1 + (f_bin_nature/2)) * (
             (1/(2-a1)) * ((m_1**(2-a1) - m_min**(2-a1))/(m_min**-a1))
             + ((1/(2-a2))
                * (m_1/m_min)**-a1 * ((m_2**(2-a2) - m_1**(2-a2))/(m_1**-a2)))
             + ((1/(2-a3)) * (m_1/m_min)**-a1 * (m_2/m_1)**-a2
                * ((m_max**(2-a3) - m_2**(2-a3))/(m_2**-a3))))
         return mean_mass
 
-    def mean_mass_simulated(alpha3, m_a, m_b):
+    def mean_mass_simulated(alpha3, m_a, m_b, f_bin_simulated):
```
mkruckow, line 113: Do we need this function? Wouldn't it be better to get the mean_mass_simulated=mass_simulated/count_simulated?

posydon/popsyn/normalized_pop_mass.py
```diff
@@ -107,30 +108,74 @@ def fraction_simulated(f_bin, m_1, m_2, m_min, m_max,
...
+        underlying_total_mass=initial_ZAMS_TOTAL_binaries/f_corr_binaries
+
+        f_corr_single_stars=0
+
+    if (f_bin_simulated != 1): #when you have modeled both binary and single stars) 
+        if (f_bin_nature == 0): #you want the underlying mass for a population consisting of only single stars
+
+            
+
+            f_corr_single_stars = (fraction_simulated(1, m_1, m_2, m_min, m_max, alpha1, alpha2, alpha3)
```
mkruckow: Why is the first argument here `1` and not `f_bin_nature`? (same in line 152)

posydon/popsyn/normalized_pop_mass.py
```diff
@@ -107,30 +108,74 @@ def fraction_simulated(f_bin, m_1, m_2, m_min, m_max,
...
+            f_corr_binaries = (fraction_simulated(1, m_1, m_2, m_min, m_max, alpha1, alpha2, alpha3)
+                  * mean_mass_simulated(alpha3, m_a, m_b, f_bin_simulated)
+                  / mean_mass_of_binary(f0, f_bin_nature, m_1, m_2, m_min, m_max,
+                                        alpha1, alpha2, alpha3))
+            
+            underlying_total_mass=initial_ZAMS_TOTAL_binaries/f_corr_binaries
+            f_corr_single_stars=0
+
+       
+        else:  #you want the underlying mass for a population consisting of both single and binary stars
```
mkruckow: Why do we need the other cases? Wouldn't it be OK, to just go with this general case always?

I think the only case is if one of the corrections get 0. Hence, I'd go with the general case and use
```
underlying_total_mass=0.0
if (f_corr_single_stars>0):
    underlying_total_mass+=(initial_ZAMS_TOTAL_single/f_corr_single_stars)
if (f_corr_binaries>0):
    underlying_total_mass+=(initial_ZAMS_TOTAL_binaries/f_corr_binaries)
```

posydon/popsyn/normalized_pop_mass.py
dimsour94: In this case, our simulations model both binary systems and single stars, but the desired (in nature) population consists solely of single stars (f_bin_nature=0). So we have modeled the following fraction of all systems (1-fbin_nature)=1.

posydon/popsyn/normalized_pop_mass.py
dimsour94: In my opinion, it's just a different approach. However, you might be right—it does seem much more straightforward to use this.

posydon/popsyn/synthetic_population.py
```diff
@@ -992,16 +992,31 @@ def __init__(
...
+                mask=self.oneline["state_i"]=="initially_single_star"
+                filtered_data_single = self.oneline[mask]
+                filtered_data_binaries = self.oneline[~mask]
+    
+                simulated_mass_binaries = np.sum(filtered_data_binaries[["S1_mass_i", "S2_mass_i"]].to_numpy())
+                simulated_mass_single = np.sum(filtered_data_single[["S1_mass_i"]].to_numpy())
+                simulated_mass = simulated_mass_binaries + simulated_mass_single
+            else:
+                simulated_mass_binaries = np.sum(self.oneline[["S1_mass_i", "S2_mass_i"]].to_numpy())
+                simulated_mass_single = None
```
mkruckow: May just a style issue, but wouldn't it be saver to `set simulated_mass_single = 0.0` here, to ensure that it is a float and not a none type object. Usually, we use None for undefined objects, but here I'd say the simulated mass of single stars is well defined to be zero, because they didn't get simulated.

posydon/popsyn/binarypopulation.py
```diff
@@ -510,8 +510,10 @@ def combine_saved_files(self, absolute_filepath, file_names, **kwargs):
         
         
         with pd.HDFStore(absolute_filepath, mode=mode, complevel=complevel, complib=complib) as store:
-            simulated_mass = 0
-            number_of_systems = 0
+            number_of_systems=0
```
mkruckow, line 516: I think you shouldn't remove the initialization of the `simulated_mass`, here. It will be undefined in case your try later fails, while it will always being accessed in `data={'simulated_mass': simulated_mass, ...`.

posydon/popsyn/binarypopulation.py
```diff
@@ -520,7 +522,19 @@ def combine_saved_files(self, absolute_filepath, file_names, **kwargs):
...
                     oneline = pd.read_hdf(f, key='oneline')
-                    simulated_mass += oneline['S1_mass_i'].sum() + oneline['S2_mass_i'].sum()
+                    if "initially_single_star" in oneline["state_i"].values:
+                        mask= oneline["state_i"]=="initially_single_star"
+                        filtered_data_single = oneline[mask]
+                        filtered_data_binaries = oneline[~mask]
+    
+                        simulated_mass_binaries = np.sum(filtered_data_binaries[["S1_mass_i", "S2_mass_i"]].to_numpy())
+                        simulated_mass_single = np.sum(filtered_data_single[["S1_mass_i"]].to_numpy())
+                        simulated_mass = simulated_mass_binaries + simulated_mass_single
```
mkruckow: The simulated mass should be a sum over several files (as before), which are looped here. (The same applied to the else case.)
Additionally, I guess, you'd like to use a sum for the new `simulated_mass_binaries` and `simulated_mass_single`, too.

posydon/popsyn/normalized_pop_mass.py
```diff
@@ -19,9 +19,8 @@ def initial_total_underlying_mass(df=None, df1=None, df2=None, **kwargs):
 
     Parameters
     ----------
-    df : DataFrame
-        Data frame of a population class. If nothing is provided the code will
-        sample initial conditions.
+    df,df1,df2 : float
+        Total simulated mass, simulated mass of binary systems and simulated mass of single stars, respectively.
```
mkruckow: If they aren't data frames anymore, I think we should consider to rename them.
I'd suggest to call them `simulated_mass`, `simulated_mass_single`, and `simulated_mass_binaries`, respectively.

posydon/popsyn/binarypopulation.py
```diff
@@ -520,7 +522,19 @@ def combine_saved_files(self, absolute_filepath, file_names, **kwargs):
                                  min_itemsize=history_min_itemsize)
                     
                     oneline = pd.read_hdf(f, key='oneline')
-                    simulated_mass += oneline['S1_mass_i'].sum() + oneline['S2_mass_i'].sum()
+                    if "initially_single_star" in oneline["state_i"].values:
```
maxbriel: At this stage, it's known from the ini file parameter what sort of run is being done, i.e. binaries only or not.
Also, does the split-up speed up the calculation?

Otherwise I would suggest:

```
mask = online['state_i'] == 'initially_single_star'
if np.any(mask);
   ...
else:
   ...
```

posydon/popsyn/binarypopulation.py
```diff
@@ -542,7 +556,7 @@ def combine_saved_files(self, absolute_filepath, file_names, **kwargs):
             tmp_df = pd.DataFrame(
                 index=[self.metallicity],
                 data={'simulated_mass': simulated_mass,
-                      'underlying_mass': initial_total_underlying_mass(df=simulated_mass, **self.kwargs)[0], 
+                      'underlying_mass': initial_total_underlying_mass(df=simulated_mass,df1=simulated_mass_single, df2= simulated_mass_binaries, **self.kwargs)[0],
```
maxbriel: I would suggest adding these as additional data, similar to the simulated mass. 
Also I would  not calculate the underlying mass here, since the underlying mass calculation requires a user-defined input, ie. f_bin_nature. :)
Instead, a new function as part of the population class would be useful. I'm happy to write this after the weekend.

maxbriel:
@dimsour94 I've changed how/where the calculation is done to allow for a user defined `f_bin_nature`, instead of doing the calculation when creating the file. I branched off your branch and implemented it here: [13ab4b0](https://github.com/POSYDON-code/POSYDON/commit/13ab4b037bae596507e9ae463e584972249d0376)

It calculates the simulated_mass, simulated_mass_single and simulated_mass_binaries and stores those values. I removed the `if` statement, because `np.nansum()` will allow you to sum over an empty dataframe and returns 0.0. :) 
I noticed there's some repeating code in `synthetic_population.py` and `binarypopulation.py` for this calculation. 
This might be combined into a function and placed in another file. Not sure about this though. 

The `calculate_underlying_mass()` function in the Population class will now allow the user to calculate the underlying mass and add it to the `Population.mass_per_metallicity`.

**You will still have to implement the passing of `f_bin_nature` to the `calculate_underlying_mass()` function.**

maxbriel:
I hope I didn't push to this branch. It wasn't my intention, sorry if I did that! I added 2 more commits on the other branch for some additional checks: f561455

posydon/popsyn/binarypopulation.py
```diff
@@ -522,18 +523,15 @@ def combine_saved_files(self, absolute_filepath, file_names, **kwargs):
...
-                        simulated_mass = simulated_mass_binaries
+                    
+                    # split weight between single and binary stars
+                    mask = oneline["state_i"] == "initially_single_star"
+                    filtered_data_single = oneline[mask]
+                    filtered_data_binaries = oneline[~mask]
+                    
+                    simulated_mass_binaries += np.nansum(filtered_data_binaries[["S1_mass_i", "S2_mass_i"]].to_numpy())
+                    simulated_mass_single += np.nansum(filtered_data_single[["S1_mass_i"]].to_numpy())
+                    simulated_mass += simulated_mass_single + simulated_mass_binaries
```
mkruckow: I think, this is wrong now, `simulated_mass_binaries` and `simulated_mass_single` are sums growing with each loop.
`simulated_mass` is supposed to be the sum of the two final sums and not summed again for each loop.
It should read: `simulated_mass = simulated_mass_single + simulated_mass_binaries`, which would be equivalent to `simulated_mass += np.nansum(filtered_data_binaries[["S1_mass_i", "S2_mass_i"]].to_numpy()) + np.nansum(filtered_data_single[["S1_mass_i"]].to_numpy())`

posydon/popsyn/synthetic_population.py
```diff
@@ -1033,6 +1025,48 @@ def __init__(
...
+            The underlying mass of the population.
+        """
+        
+        if 'underlying_mass' in self.mass_per_metallicity.columns:
+            Pwarn("underlying_mass already exists in the mass_per_metallicity table.")
+            if overwrite:
+                Pwarn("Overwriting the underlying_mass values.")
+            else:
+                Pwarn("Not overwriting the underlying_mass values.")
+                return
```
mkruckow, line 1057: Why do we always need two warnings? I'd suggest to have only one warning (moving text from the first to the later two). Additionally, we should specify the warning, e.g. the overwriting is an "OverwriteWarning", and with no overwriting it is an "IncompletenessWarning" as things get skipped.
Here an example (please insert line breaks where needed):
```
        if 'underlying_mass' in self.mass_per_metallicity.columns:
            warn_text="underlying_mass already exists in the mass_per_metallicity table."
            if overwrite:
                Pwarn(warn_text+" Overwriting the underlying_mass values.", "OverwriteWarning")
            else:
                Pwarn(warn_text+" Not overwriting the underlying_mass values, skipping it.", "IncompletenessWarning")
                return
```

mkruckow:
> I hope I didn't push to this branch. It wasn't my intention, sorry if I did that! I added 2 more commits on the other branch for some additional checks: [f561455](https://github.com/POSYDON-code/POSYDON/commit/f5614556c5fb9e98a9b9fb148ba8572bb33ba914)

It looks like you pushed to this branch, hence @dimsour94 please don't miss to pull to your local branch before making more changes, otherwise you'll create merge conflicts, which you'll need to resolve later.

maxbriel (dismissed):
Looks good to me. I tested it and it gives the simulated mass per system.
I have added that `calculate_underlying_mass()` save the calculated data to the Population file.

maxbriel:
Tutorials changed in PR #406 Ready for this PR to be accepted.

mkruckow (dismissed):
Beside a few typos, e.g. "stae" -> "star" and not always respecting the 80 characters line length, the PR looks complete now.

mkruckow (approved):

maxbriel (approved):

Issue #378, opened 2024-09-06
get_redshift_from_cosmic_time doe not work


maxbriel:
The `get_redshift_from_cosmic_time(t_cosm)` function calls `redshift_from_cosmic_time_interpolator(t_cosm)`

However, `redshift_from_cosmic_time_interpolator` does not take a function input. It returns the interpolator itself.
The `get_redshift_from_cosmic_time` function should be altered.

celiotine:
Branch created from development called "fix_redshift_cosmic_time"

celiotine:
See PR #381

PR #381, opened 2024-09-09
fix get_redshift_from_cosmic_time() function

celiotine:
This PR addresses Issue #378.
Update the t-z interpolator in rate_calculations.py to use CubicSpline instead of interp1d, which is now considered "legacy" functionality by SciPy.
Correct the call of the interpolator in the get_redshift_from_cosmic_time() function.

### DIFF
--- a/posydon/popsyn/rate_calculation.py
+++ b/posydon/popsyn/rate_calculation.py
@@ -10,7 +10,7 @@ from astropy import constants as const
 import numpy as np
 import scipy as sp
 from astropy.cosmology import z_at_value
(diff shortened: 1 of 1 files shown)


maxbriel (approved):
Looks good! Thank you! 

Just to note that the calls to get_redshift_from_cosmic_time will be slow due to always "training" the interpolator every time. When this is used in POSYDON, we request the interpolator once and call that object instead of this function. So no issues there and having this function is useful if you only want to do a few calculations.

astroJeff (approved):

Issue #243, opened 2024-01-22
Disrupted/detached step unable to get final values 


maxbriel:
The following binary fails in `step_detached` due to failing in the `get_star_final_values(secondary, secondary.htrack, m01)` function on line 1929, see details below.

<details><summary>Error Traceback</summary>
<p>

<img width="1063" alt="Screenshot 2024-01-22 at 16 48 03" src="https://github.com/POSYDON-code/POSYDON/assets/14039563/0aebbf35-10b0-4bbb-b6ae-d2669dc51dda">

</p>
</details> 

## Example binary

<img width="895" alt="Screenshot 2024-01-22 at 16 49 07" src="https://github.com/POSYDON-code/POSYDON/assets/14039563/4636f6c8-af64-4825-8802-074e34797d65">

```
STAR1 = SingleStar(**{'mass': 9,
                      'state': 'H-rich_Core_H_burning'})
STAR2 = SingleStar(**{'mass':0.8,
                      'state': 'H-rich_Core_H_burning'})

BINARY = BinaryStar(STAR1, STAR2,  
                    **{'time': 0.0, 'state': 'detached', 'event': 'ZAMS', 'orbital_period':4513.150157, 'eccentricity': 0.0},
                    properties = sim_pop)
```

maxbriel:
The issue originates from line 605 in `interpolation.py`.

```
         while (kvalue_low is None or np.isnan(kvalue_low)):
                mass_low = np.max(self.grid_mass[mass_low > self.grid_mass])
                try:
                    kvalue_low = self.grid_final_values[mass_low][key]
                except KeyError:
                    self.load_grid(mass_low)
                    kvalue_low = self.grid_final_values[mass_low][key]
```

`kvalue_low = self.grid_final_values[mass_low][key]` stays `nan`. 

The key is`S1_avg_c_in_c_core_at_He_depletion` and it cannot find any masses. 

@mkruckow could it be that these models do not reach He depletion?

mkruckow:
For low mass stars it makes sense that there is no well defined value at He depletion, because they never get to He depletion.
I think, we should stop the loop as soon as it reached the lowest mass in the grid and at this point keep the nan value.

celiotine:
Looks like this bug is being caused by the same code described in Issue #286

maxbriel:
I've tried exiting the loop when the mask becomes empty. 
```
if np.sum(mass_low > self.grid_mass) == 0:
    break
```

[...]

PR #385, opened 2024-09-13
Add an escape option if value cannot be found in the single star grids

maxbriel:
Matching is still performed, but values are allowed to be skipped.
This solves issue: https://github.com/POSYDON-code/POSYDON/issues/243


mkruckow (approved):
It a very pythonic way to sum over booleans, I'd used `not np.any`, but OK.

celiotine (approved):
I ran a small population (10k binaries) on quest with this PR and it seems it fixes the issue(s) we discussed!

Issue #353, opened 2024-07-11
large memory usage on column load of a Population


maxbriel:
Because data_columns=True was removed, the memory-usage of loading a single column from the hdf5 file is equivalent to loading the complete file.
While a user can loop over the binaries and get the indices of the models they like, the behaviour from pandas reading from an hdf5 file is unexpected.

The solution is either to chunk read the file and only retain the column.
Or switch storage methods, away from pandas hdf5 format and to a h5py format, which does allow for column readouts.

maxbriel:
Should be fixed with PR #386

PR #386, opened 2024-09-13
Population() fix for large memory usage when reading files

maxbriel:
When using pd.read_hdf(), pandas reads in the complete Dataframe into memory in a weird way that uses much more memory than even the file is (~14GB of RAM usage for a 3GB file).
This issue originated from the choice that we made for `data_columns=False`. 

This PR does the following:
1. Create a new class `DFInterface` for interfacing with the Oneline and History dataframes in the file.
     - Contains the `select()`, `head()` and `tail()` functions that are the same for both above systems
     - More can be generalised, but I wasn't sure how to generalise a built-in function and give it an input, that is `self.__repr__`. (added) 
2. The new select function reads the file in chunks and concatenates the result. This limits the total memory usage to the given chunksize. The default parameters on my machine use about 3.6GB, but usage varies a bit.
3. The `History` and `Oneline` class now use `self.select` in the `__getitem__` function to read data from the file. (except for history[boolean_mask], I couldn't find a good way to implement this with the `self.select`)

NOTE: contains PR #385 

and should fix Issue: #353

### DIFF
--- a/posydon/popsyn/synthetic_population.py
+++ b/posydon/popsyn/synthetic_population.py
@@ -706,9 +706,9 @@ class Oneline(DFInterface):
                 chunk = self.number_of_systems
             else:
                 chunk = key.stop - pre
-            return pd.read_hdf(
-                self.filename, key="oneline", start=pre, stop=pre + chunk
-            )            
+            indices = list(range(pre, pre + chunk))
+            return self.select(where=f'index in {indices}')
+        
         elif isinstance(key, int):
             return self.select(where=f"index == {key}")
         elif isinstance(key, list) and all(isinstance(x, int) for x in key):
--- a/posydon/popsyn/synthetic_population.py
+++ b/posydon/popsyn/synthetic_population.py
@@ -731,7 +731,7 @@ class Oneline(DFInterface):
                 raise ValueError(f"{key} is not a valid column!")
         elif isinstance(key, list) and all(isinstance(x, str) for x in key):
             if all(x in self.columns for x in key):
-                self.select(columns=key)
+                return self.select(columns=key)
             else:
                 raise ValueError(f"Not all columns in {key} are valid column names!")
         else:
--- a/posydon/popsyn/synthetic_population.py
+++ b/posydon/popsyn/synthetic_population.py
@@ -592,6 +592,7 @@ class History(DFInterface):
         ----------
         where : str, optional
             A string representing the query condition to apply to the data.
+            It is only possible to query on the index or string columns. 
         start : int, optional
             The starting index of the data to select.
         stop : int, optional
@@ -810,7 +811,8 @@ class Oneline(DFInterface):
         Parameters
         ----------
         where : str, optional
-            A condition to filter the rows of the oneline table. Default is None.
+            A condition to filter the rows of the oneline table. Default is None. 
+            It is only possible to query on the index or string columns. 
         start : int, optional
             The starting index of the subset. Default is None.
         stop : int, optional
@@ -826,7 +828,7 @@ class Oneline(DFInterface):
         Examples
         --------
         # Select rows based on a condition
-        >>> df = Oneline.select(where="S2_mass_i > 30")
+        >>> df = Oneline.select(where="event == 'ZAMS'")
 
         # Select rows from index 10 to 20
         >>> df = Oneline.select(start=10, stop=20)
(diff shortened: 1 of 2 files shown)


posydon/popsyn/synthetic_population.py
```diff
@@ -224,8 +224,84 @@ def merge_parallel_runs(self, pop):
 # Helper classes #
 ##################
 
+class DFInterface:
+    """A class to handle the interface between the population file and the History and Oneline classes."""
+    
+    def __init__():
+        pass
```
mkruckow: The `__init__` should make sure that all elements, this class needs are created: `self.filename` and `self.chunksize`. Maybe a `self.verbose` will be useful, like the subclasses have.

posydon/popsyn/synthetic_population.py
```diff
@@ -224,8 +224,84 @@ def merge_parallel_runs(self, pop):
 # Helper classes #
 ##################
 
+class DFInterface:
```
mkruckow, line 227: The `DFInterface` class looks very general, hence we should consider to move this to the utils, e.g. if we'd like to use it for the grids, too.

posydon/popsyn/synthetic_population.py
```diff
@@ -224,8 +224,84 @@ def merge_parallel_runs(self, pop):
...
+            A list of column names to select.
+        
+        Returns
+        -------
+        pandas.DataFrame
+            The selected data as a DataFrame.
+        '''
+        # we have to chunk the read because of memory issues
+        with pd.HDFStore(self.filename, mode="r") as store:
+            iterator = store.select("history", where=where, start=start, stop=stop, columns=columns, chunksize=self.chunksize)
```
mkruckow: Shouldn't we have `key` as the first argument instead of "history"?
In the current way I'd expect it to fail for the Online subclass.

maxbriel: changed

posydon/popsyn/synthetic_population.py
maxbriel: Added

posydon/popsyn/synthetic_population.py
maxbriel: Fair. I have no real preference for either.

posydon/popsyn/synthetic_population.py
```diff
@@ -725,10 +834,11 @@ def select(self, where=None, start=None, stop=None, columns=None):
         # Select specific columns
         >>> df = Oneline.select(columns=['S1_mass_i', 'S1_mass_f'])
         """
-        with pd.HDFStore(self.filename, mode="r") as store:
-            return store.select(
-                "oneline", where=where, start=start, stop=stop, columns=columns
-            )
+        return super().select(key='oneline',
```
mpgalleg, line 839: I was trying to tests the .select functionality and couldn't get `pop.oneline.select(where="S2_mass_i > 30")` to work. This is an example given on line 829. Am I missing something? 

The error message said it "contains an invalid variable reference ...The currently defined references are: index,columns,state_i,state_f,event_i,event_f,step_names_i,step_names_f,S1_state_i,S1_state_f,S2_state_i,S2_state_f,S1_SN_type,S2_SN_type,interp_class_HMS_HMS,interp_class_CO_HeMS,interp_class_CO_HMS_RLO,interp_class_CO_HeMS_RLO,mt_history_HMS_HMS,mt_history_CO_HeMS,mt_history_CO_HMS_RLO,mt_history_CO_HeMS_RLO
"

We should add it to the defined references or remove that example from the documentation.

posydon/popsyn/synthetic_population.py
mpgalleg: What do you mean by interface? Functions that would apply to the population file, history, and oneline? It is super minor but this word seems a little opaque to me

posydon/popsyn/synthetic_population.py
maxbriel: That came from the switch from using `data_columns=True` to  `data_columns=False`. It removes the option to use `where=" "` non-string variables.
This was the choice we made when we wanted to have faster file merging after a population run. 

However, the info in the documentation should indeed be removed. I have removed the example from the docstring and added a statement in the docstring on what you can search. It should not be in the tutorial anymore. There is should explicitly state that it can only select on string columns.

posydon/popsyn/synthetic_population.py
maxbriel: The `History` and `Oneline` classes use it to interface with the history and oneline Dataframes in the population file. Those classes use it to interact with the population file (in most cases).

mkruckow (approved):

astroJeff (approved):

PR #390, opened 2024-09-17
Small Population fixes

maxbriel:
Fixes 3 small things related to Populations:

1. Throw a warning if no values have been written to the transient population is present and return a None.
2. Remove compulsory output on `calculate_intrinsic_rate_density()` function.
3. Fix overplotting of a specific channel on a grid slice.

### DIFF
--- a/posydon/popsyn/synthetic_population.py
+++ b/posydon/popsyn/synthetic_population.py
@@ -1483,8 +1483,8 @@ class Population(PopulationIO):
 
         Returns
         -------
-        TransientPopulation
-            A TransientPopulation object for interfacing with the transient population
+        TransientPopulation or None
+            A TransientPopulation object for interfacing with the transient population or None if no systems are present in the TransientPopulation.
 
         Raises
         ------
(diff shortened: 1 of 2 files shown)


celiotine (dismissed):

posydon/popsyn/synthetic_population.py
```diff
@@ -1577,6 +1577,12 @@ def create_transient_population(
                 )
 
             previous = end
+            
+        # it can happen that no systems are selected, in which case nothing has been appended to the file in the loop
+        with pd.HDFStore(self.filename, mode="r") as store:
+            if '/transients/'+transient_name not in store.keys():
+                Pwarn("No systems selected for the transient population!", "NoSystemsSelectedWarning")
```
celiotine: I think you have to add a "NoSystemsSelectedWarning" to the posydonwarning.py file because we don't currently have one.

mkruckow: FYI: any not defined warning specified by a string, will default to the base POSYDONwarning. Thus, the warning works, but will be issued as a general POSYDONwarning.

posydon/popsyn/synthetic_population.py
```diff
@@ -1577,6 +1577,12 @@ def create_transient_population(
...
+                return None
```
mkruckow, line 1585: Please add to the doc-string, that the returned object could be of None type.

posydon/popsyn/synthetic_population.py
maxbriel: Renamed to a general warning for clarity.

mkruckow (approved):
It would be nicer to have a more specific warning, but otherwise OK.

celiotine (approved):

Issue #388, opened 2024-09-16
missing/undefined states in flow


celiotine:
When evolving a binary population, many binaries are failing due to undefined next states in the flow. In some cases, it looks like the binary states have not been added to the flow at all, and in others it is more unclear why the binaries are not being directed in the flow correctly. Here are a couple examples:

In this case, the state 'disrupted (GridMatchingFailed)' has not been added to the flow:
```
Traceback (most recent call last):
  File "/home/cel2717/.conda/envs/posydon-source/lib/python3.11/site-packages/posydon/popsyn/binarypopulation.py", line 326, in _safe_evolve
    binary.evolve()
  File "/home/cel2717/.conda/envs/posydon-source/lib/python3.11/site-packages/posydon/binary_evol/binarystar.py", line 210, in evolve
    self.run_step()
  File "/home/cel2717/.conda/envs/posydon-source/lib/python3.11/site-packages/posydon/binary_evol/binarystar.py", line 229, in run_step
    raise ValueError("Undefined next step given stars/binary states {}.".format(total_state))
ValueError: Undefined next step given stars/binary states ('NS', 'H-rich_Core_He_burning', 'disrupted (GridMatchingFailed)', None).

Failed Binary Initial Conditions:
S1 mass: 13.395757072267616
S2 mass: 13.049711269195887
S1 state: H-rich_Core_H_burning
S2 state: H-rich_Core_H_burning
orbital period: 106.85569570285801
eccentricity: 0.0
binary state: detached
binary event: detached
S1 natal kick array: [307.4048771456759 2.896410103661572 2.868709495271407 5.028051522385029]
S2 natal kick array: [None None None None]
```

In this next case, it seems this combination of star/binary states is in the flow, so not sure what is happening:
```
Traceback (most recent call last):
  File "/home/cel2717/.conda/envs/posydon-source/lib/python3.11/site-packages/posydon/popsyn/binarypopulation.py", line 326, in _safe_evolve
    binary.evolve()
  File "/home/cel2717/.conda/envs/posydon-source/lib/python3.11/site-packages/posydon/binary_evol/binarystar.py", line 210, in evolve
    self.run_step()
  File "/home/cel2717/.conda/envs/posydon-source/lib/python3.11/site-packages/posydon/binary_evol/binarystar.py", line 229, in run_step
    raise ValueError("Undefined next step given stars/binary states {}.".format(total_state))
[...]

PR #391, opened 2024-09-17
remove failed binary states from flow chart

celiotine:
This addresses part of Issue #388.
There are some binary states missing from the flow chart because the binary state strings are being created dynamically. This is causing some binaries to fail and throw errors.
This PR adds the missing states to the flow chart and adds checks to make sure that states exist in the flow chart before the binary state string is updated.
After looking into it, I believe this is the best workaround because it would be cumbersome and generally not good coding practice to have dynamically-matched keys for some binary states in the flow chart, unless we want to change the data structure of the entire flow chart.

One question: in the detached step, if there is an integration failure, we send it to step_end. However, in step_dco, if there is an integration failure, we raise a NumericalError. Is this the behavior we want? As someone who did not write this code it appears inconsistent to me.

### DIFF
--- a/posydon/binary_evol/DT/double_CO.py
+++ b/posydon/binary_evol/DT/double_CO.py
@@ -105,7 +105,7 @@ class DoubleCO:
 
         if s.status == -1:
             binary.state += ' (Integration failure)'
-            raise NumericalError("Integrations failed: " + s.message)
+            raise NumericalError("Integration failed: " + s.message)
 
         elif s.status == 1:
 
--- a/posydon/binary_evol/DT/double_CO.py
+++ b/posydon/binary_evol/DT/double_CO.py
@@ -104,8 +104,9 @@ class DoubleCO:
             t_final = [t_inspiral]
 
         if s.status == -1:
-            binary.state += ' (Integration failure)'
-            raise NumericalError("Integration failed: " + s.message)
+            failed_state = binary.state
+            set_binary_to_failed(binary)
+            raise NumericalError(f"Integration failed for {failed_state} DCO!")
 
         elif s.status == 1:
 
--- a/posydon/binary_evol/DT/double_CO.py
+++ b/posydon/binary_evol/DT/double_CO.py
@@ -106,7 +106,7 @@ class DoubleCO:
         if s.status == -1:
             failed_state = binary.state
             set_binary_to_failed(binary)
-            raise NumericalError(f"Integration failed for {failed_state} DCO!")
+            raise NumericalError(f"Integration failed for {failed_state} DCO ({self.state1}, {self.state2}): ", s.message)
 
         elif s.status == 1:
 
--- a/posydon/binary_evol/DT/step_detached.py
+++ b/posydon/binary_evol/DT/step_detached.py
@@ -1174,11 +1174,16 @@ class detached_step:
 
 
         if interp1d_sec is None or interp1d_pri is None:
-            # binary.event = "END"
-            binary.state += " (GridMatchingFailed)"
-            if self.verbose or self.verbose == 1:
-                print("Failed matching")
-            return
+
+            if binary.state in ['detached', 'disrupted', 'merged']:
+                binary.state += " (GridMatchingFailed)"
+                if self.verbose or self.verbose == 1:
+                    print("Failed matching")
+                return       
+            else:
+                raise ValueError("need to add new binary state to flow chart for evolution to end correctly ",
+                                "when grid matching fails: ", binary.state," (GridMatchingFailed)")           
+                   
         t0_sec = interp1d_sec["t0"]
         t0_pri = interp1d_pri["t0"]
         m01 = interp1d_sec["m0"]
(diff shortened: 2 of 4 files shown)


posydon/binary_evol/DT/step_detached.py
```diff
@@ -1174,11 +1174,16 @@ def get_star_data(binary, star1, star2, htrack,
...
         if interp1d_sec is None or interp1d_pri is None:
-            # binary.event = "END"
-            binary.state += " (GridMatchingFailed)"
-            if self.verbose or self.verbose == 1:
-                print("Failed matching")
-            return
+
+            if binary.state in ['detached', 'disrupted', 'merged']:
+                binary.state += " (GridMatchingFailed)"
+                if self.verbose or self.verbose == 1:
```
mkruckow: I know, you haven't change that code part, but out of curiosity: Why do we check `self.verbose` for the boolean representation and whether it is 1? Usually, if the second condition is fulfilled the first one should be as well. (There are ways in python to create objects, which would need the second condition to capture it, because the first one doesn't. But those would be a very weird object, which needs to be explicitly designed that way, no common object should behave like that.)

posydon/binary_evol/DT/step_detached.py
```diff
@@ -1503,10 +1508,13 @@ def get_omega(star, is_secondary = True):
...
                 print("Integration failed", s.message)
-                binary.state += ' (Integration failure)'
-                # binary.event = "END"
+
+                if binary.state in ['detached']:
+                    binary.state += ' (Integration failure)'
+                else:
+                    raise ValueError("need to add new binary state to flow chart for evolution to end correctly ",
+                                "when detached integration fails: ", binary.state, " (Integration failure)")
                 return
```
mkruckow: Just a style point: why do you have the return here behind the raising of the error, while in line 1182 you put it only in the other case? I don't have a strong preference, but it should be consistent, hence may check which way is otherwise used in the code.

posydon/binary_evol/DT/step_detached.py
celiotine: I'm not sure why this is written this way. I've noticed it used in other places, like line 1505:
```
if self.verbose and self.verbose != 1:
                print("ODE solver duration: "
                      f"{t_after_ODEsolution-t_before_ODEsolution:.6g}")
```
This example in particular does not make sense to me.

According to the .ini file, this parameter should be a boolean, but it looks like the code is also allowing it to be 0/1 (presumably). Overall its usage in the detached step seems to be inconsistent throughout.
It should be simple to go through and fix this, I could do that in a different PR.

posydon/binary_evol/DT/step_detached.py
celiotine: Yeah, I agree it should be consistent, I've moved it inside the code block.

mkruckow (dismissed):
Look good to me.

astroJeff:
Update (Sept 19): We discussed how we want to properly cause a binary to fail due to integration failures

posydon/binary_evol/DT/double_CO.py
```diff
@@ -104,8 +104,9 @@ def ev_contact(t, y):
             t_final = [t_inspiral]
 
         if s.status == -1:
-            binary.state += ' (Integration failure)'
-            raise NumericalError("Integration failed: " + s.message)
+            failed_state = binary.state
+            set_binary_to_failed(binary)
+            raise NumericalError(f"Integration failed for {failed_state} DCO!")
```
mkruckow: Why don't we keep the message from the solver added to the error message?

Similarly, it would be nice to add more information to all the errors, especially stuff which is otherwise not preserved in the binary object, e.g. `primary.co`.

posydon/binary_evol/DT/step_detached.py
```diff
@@ -1169,16 +1169,16 @@ def get_star_data(binary, star1, star2, htrack,
             interp1d_pri = get_star_data(
                 binary, primary, secondary, primary.htrack, False)[0]
         else:
-            raise MatchingError("During matching, the primary should either be normal or not normal."
-                                "`non_existent_companion` should be zero.")
+            raise ValueError("During matching, the primary should either be normal (stellar object) or ",
```
maxbriel, line 1172: This is a standard `ValueError`?
I'm not too up-to-date with the warning capturing during a population run.
Does that mean this will stop a POSYDON run?

mkruckow: First, it's an error not a warning.
Second, no error in a evolution step can stop POSYDON when running a binary population, because of the try around it with a general exception catch, see [lines 325-341 in binarypopulation.py](https://github.com/POSYDON-code/POSYDON/blob/development/posydon/popsyn/binarypopulation.py)

maxbriel: The point/question I was trying to make is that there's a distinctly different treatment of POSYDON errors and general errors in binarypopualtion runs. This should be a POSYDON error, not a standard ValueError, right?
The error before was a POSYDON error.

mkruckow: When you just have a test message, POSYDON errors behave the same way as any standard errors. POSYDON errors have the additional ability to deal with POSYDON objects `SingleStar` or `BinaryStar`.
Hence, I'd say here it doesn't make any difference what error type we have. But I suggest to use the one with most appropriate meaning.

celiotine: I changed this from a POSYDON error to a value error because I think it is something that should never happen? So if it does happen the python error should be thrown.

maxbriel: Okay! This makes sense. Thank you for the additional clarification :)

maxbriel (approved):

mkruckow (approved):

Issue #238, opened 2024-01-11
replace np.isnan with pd.isna


tassos25:
pd.isna is more robust as it catches both None and np.nan. Consider using pd.isna  throughout the code.

mkruckow:
Here some important info for replacements:

- `np.isnan` -> `pd.isna`
- `~np.isnan` or `not np.isnan` -> `pd.notna` this is needed because `np.isnan` and `pd.isna` give different types in case of a scalar (`np.bool_` vs. `bool`), maybe there are differences for the array like parts as well

celiotine:
@mkruckow started doing this in PR #482, clearly it has been causing issues

PR #397, opened 2024-09-26
update profile interpolation for consistency with paper

elizabethteng:
- [x] update method in CompileData to use pd.concat() instead of append()
- [x] update assignment of total mass to being based on mass profile to avoid inconsistency with final values total mass
- [x] fix definition of omega profile
- [x] added posydon warning for nans in the final values (check with Camille)
- [x] add ability to set random seed to standardize results
- [x] add arguments for customizing neural network size, learning rate
- [x] simplify PCA step
- [x] make models for each MT class in a grid
- [x] add code for customizing neural network size
- [x] log vs linear -- make sure this is consistent with IF interpolator
- [x] update composition method
- [x] separate composition and density training
- [x] address comments from Matthias

### DIFF
--- a/posydon/interpolation/profile_interpolation.py
+++ b/posydon/interpolation/profile_interpolation.py
@@ -662,7 +662,6 @@ class Composition:
                     hbounds,nonflat_a = calc_bevel_bounds(h_prof)
                     valid_hbounds,valid_nonflat_a = calc_bevel_bounds(valid_h_prof)
                     pbounds,nonflat_p = calc_bevel_bounds(plus_prof)
-                    print(pbounds)
                     valid_pbounds,valid_nonflat_p = calc_bevel_bounds(valid_plus_prof)
                     nonflat = np.intersect1d(nonflat_a,nonflat_p)
                     valid_nonflat = np.intersect1d(valid_nonflat_a,valid_nonflat_p)
--- a/posydon/interpolation/profile_interpolation.py
+++ b/posydon/interpolation/profile_interpolation.py
@@ -350,8 +350,10 @@ class Density:
         Args:
             initial (array-like) : log-space initial conditions for training data.
             profiles (array-like) : final density profiles for training data. 
+            mt (array-like) : mass transfer classes corresponding to training set binaries. 
             valid_initial (array-like) : log-space initial conditions for validation data.
             valid_profiles (array-like) : final density profiles for validation data.
+            valid_mt (array-like): mass transfer classes corresponding to validation set binaries. 
             IF_interpolator (string) : path to .pkl file for IF interpolator for central density, final mass values
             n_comp (int) : number of PCA components. 
             hms_s2 (Boolean) : option to do profiles of star 2 in HMS-HMS grid
--- a/posydon/interpolation/profile_interpolation.py
+++ b/posydon/interpolation/profile_interpolation.py
@@ -260,7 +260,7 @@ class ProfileInterpolator:
     def predict(self,inputs):
         """Predict density, H mass fraction, and He mass fraction profiles from inputs.
         Args:
-            inputs (array-like) : log-space initial conditions of N binaries to predict, shape (N,3).
+            inputs (array-like) : linear-space initial conditions of N binaries to predict, shape (N,3).
         Returns:
             mass_coords (array-like) : linear-scale mass enclosed profile coordinates.
             density_profiles (array-like) : log-scale density profile coordinates.
@@ -416,18 +416,18 @@ class Density:
     def predict(self,inputs):
         """Predicts profile for n sets of given inputs, in array of shape (n,3).
         Args: 
-            inputs (array-like) : log-space initial conditions of N binaries to predict, shape (N,3).
+            inputs (array-like) : linear-space initial conditions of N binaries to predict, shape (N,3).
         Returns:
             mass_coords (array-like) : linear-scale mass enclosed profile coordinates.
             density_profiles (array_like) : log-scale density profile coordinates.
         """
         # predict PCA weights
         regress_prof = lambda x: self.model_prof(x)
-        pca_weights_pred = regress_prof(inputs).numpy()
+        pca_weights_pred = regress_prof(np.log10(inputs)).numpy()
                 
         # predict surface density
         regress_rho = lambda x: self.model_rho(x)
-        min_rho = regress_rho(inputs).numpy()[:,0]
+        min_rho = regress_rho(np.log10(inputs)).numpy()[:,0]
         
         # IF interpolate final mass, center density 
         if self.hms_s2==False:
@@ -436,8 +436,8 @@ class Density:
         else:
             m_ind = self.model_IF.interpolators[0].out_keys.index("star_2_mass")
             center_ind = self.model_IF.interpolators[0].out_keys.index('S2_log_center_Rho')
-        max_rho = self.model_IF.interpolators[0].test_interpolator(10**inputs)[:,center_ind]                    
-        pred_mass = self.model_IF.interpolators[0].test_interpolator(10**inputs)[:,m_ind]
+        max_rho = self.model_IF.interpolators[0].test_interpolator(inputs)[:,center_ind]                    
+        pred_mass = self.model_IF.interpolators[0].test_interpolator(inputs)[:,m_ind]
             
         # reconstruct profile
         norm_prof = self.pca.inverse_transform(pca_weights_pred*self.scaling)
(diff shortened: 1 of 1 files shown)


posydon/interpolation/profile_interpolation.py
```diff
@@ -134,9 +137,14 @@ def scrape(self,grid,ind,hms_s2):
         
         if 'omega' in self.names:
             profiles[-1]= profiles[self.names.index('omega')]/  \
-                          grid.final_values['S1_surf_avg_omega_div_omega_crit'][ind]
+                          (grid.final_values['S1_surf_avg_omega'][ind]/ \
+                           grid.final_values['S1_surf_avg_omega_div_omega_crit'][ind])
         
-        return scalars, profiles
+        if not np.isnan(grid.final_values['S1_surf_avg_omega_div_omega_crit'][ind]):
```
mkruckow: When I remember correctly, at some point we said to not use `np.isnan` but `pd.isna` instead. See issue #238 , thus I'd encourage to not add new `np.isnan`s to the code.

posydon/interpolation/profile_interpolation.py
```diff
@@ -134,9 +137,14 @@ def scrape(self,grid,ind,hms_s2):
...
+            return scalars, profiles
+        else:
+            Pwarn("nan in final values, binary will not be"
+                  "included in file","InappropriateValueWarning")
```
mkruckow: I'm not sure, but maybe an "IncompletenessWarning" is better here, because you like to point out, that stuff won't be added and not that there are `nan` values in the data. This is a borderline case, where both warning types apply and you should chose the one you think is more appropriate to transport your main aim.

posydon/interpolation/profile_interpolation.py
```diff
@@ -149,13 +157,21 @@ def save(self, filename):
                    for key in SAVE_ATTRS if key not in DONT_SAVE}
         with open(filename, 'wb') as f:
             pickle.dump(myattrs, f)
-            
+
             
 class ProfileInterpolator:
     
-    def __init__(self):
+    def __init__(self,seed_value):
```
mkruckow: I suggest to set `seed_value=None` as default, so you can call the function without specifying it.
Please add to the doc-string, that None is allow beside an int. Add the default value to the doc-string as well, when you decide to set it.

posydon/interpolation/profile_interpolation.py
```diff
@@ -192,60 +208,86 @@ def load_profiles(self,filename,valid_split=0.2):
...
             density_patience (int) : patience parameter for NN callback in density profile model
             comp_bounds_epochs (int) : number of epochs used to train composition profiles model
             comp_bounds_patience (int) : patience parameter for NN callback in composition profiles model
             loss_history (Boolean) : option to return training and validation loss histories
             hms_s2 (Boolean) : option to do profiles of star 2 in HMS-HMS grid
+            depth (int) : depth of neural network for principal component weights
+            width (int) : width of neural network for principal component weights
+            depthn (int) : depth of neural network for normalizing value
+            widthn (int) : width of neural network for normalizing value
+            lr (float) : learning rate for neural network training
```
mkruckow: Try to avoid that short variable names like `lr`, may just call it `learning_rate`.
P.S.: the same applies in other cases later.

posydon/interpolation/profile_interpolation.py
```diff
@@ -192,60 +208,86 @@ def load_profiles(self,filename,valid_split=0.2):
...
         if loss_history==True:
-            return self.comp.loss_history, self.dens.loss_history
-
-    def predict(self,inputs):
+            if comp==True and density==True:
+                return self.dens.loss_history, self.comp.loss_history
+            elif comp!=True and density==True:
+                return self.dens.loss_history
+            elif comp==True and density!=True:
+                return self.comp.loss_history
```
mkruckow, line 276: Please reflect the three return options in the doc-string. I think it is important to specify well, which order the two histories will have in case both are returned. (it looks that the order switched)
Additionally, please consider to add a general else case with a return to not let the function ever exit without a return.
P.S.: the same applies to the predict function below.

posydon/interpolation/profile_interpolation.py
```diff
@@ -311,54 +352,57 @@ def __init__(self,initial,profiles,valid_initial,
             IF_interpolator (string) : path to .pkl file for IF interpolator for central density, final mass values
             n_comp (int) : number of PCA components. 
             hms_s2 (Boolean) : option to do profiles of star 2 in HMS-HMS grid
+            depth (int) : depth of neural network for principal component weights
+            width (int) : width of neural network for principal component weights
+            depthn (int) : depth of neural network for normalizing value
+            widthn (int) : width of neural network for normalizing value
+            lr (float) : learning rate for neural network training
```
mkruckow: There are not mentioned variables: `mt`, `valid_mt` and a not required variable `lr`.

Beside that, it looks like, what you call `mt` here is in fact the interpolation class. Could you please double check what you are actually using here?

posydon/interpolation/profile_interpolation.py
```diff
@@ -367,39 +411,44 @@ def train(self,loss=losses.MeanSquaredError(),prof_epochs=3000,prof_patience=200
...
+        self.model_norm.fit(self.initial,self.surf_val,epochs=300,callbacks=[callback],
+                            validation_data = (self.valid_initial,self.valid_surf_val))
                                      
         print("done training")
     
     def predict(self,inputs):
         """Predicts profile for n sets of given inputs, in array of shape (n,3).
         Args: 
-            inputs (array-like) : log-space initial conditions of N binaries to predict, shape (N,3).
+            inputs (array-like) : linear-space initial conditions of N binaries to predict, shape (N,3).
```
mkruckow: May add that the input should be positive (>0), because you'll take the log10 of them later.

posydon/interpolation/profile_interpolation.py
```diff
@@ -492,20 +545,41 @@ def __init__(self,initial,h_profiles,he_profiles,star_state,
...
+            # i.e. boundaries of 'bevel' shape
+            nonflat=[] # ensures that training data only has the correct "non-flat" shape
+            bounds = []   # collect information on parameters for given H and He profile shapes
+            for i in range(len(profs)):
+                diff = np.where(profs[i][1:]-profs[i][:-1]>0.002)[0]
+                if len(diff>=2):
+                    bounds.append(np.array([diff[0]+1,diff[-1]+1])/200)
+                    nonflat.append(i)
+                else:
+                    bounds.append([np.nan,np.nan])
```
mkruckow: Just for consistency: I'd suggest to either always append a numpy array or a list. line 577 vs. 580.

posydon/interpolation/profile_interpolation.py
```diff
@@ -650,93 +695,115 @@ def predict_single(self,initial,center_H,surface_H,center_He,surface_He,star_sta
...
             H = np.zeros(200) # stripped Helium stars have no Hydrogen
         
-        if star_state == "stripped_He_non_burning":
+        # if there is no training data in the state, return nans
+        if len(self.sort_ind[star_state])==0: 
+            H = np.ones(200)*np.nan
+            He = np.ones(200)*np.nan
+            
+        elif star_state == "stripped_He_non_burning":
             He = np.ones(200)*surface_He # non-burning stars have a flat He profile
```
mkruckow: For all the states, where you expect a flat profile, wouldn't it be better to check that on the difference between the surface and central value?
When they differ, wouldn't it be better to fill the profile with the average of the central and surface value instead of the pure surface value?

posydon/interpolation/profile_interpolation.py
elizabethteng: fixed, thanks!

posydon/interpolation/profile_interpolation.py
elizabethteng: good point. I'll change to IncompletenessWarning.

posydon/interpolation/profile_interpolation.py
elizabethteng: good idea. Done!

posydon/interpolation/profile_interpolation.py
elizabethteng: done!

posydon/interpolation/profile_interpolation.py
elizabethteng: good point. I've put in your suggestion and added a warning.

posydon/interpolation/profile_interpolation.py
elizabethteng: good catch. just fixed this!

posydon/interpolation/profile_interpolation.py
elizabethteng: done!

posydon/interpolation/profile_interpolation.py
elizabethteng: done

posydon/interpolation/profile_interpolation.py
elizabethteng: done

mkruckow (dismissed):

mkruckow (approved):
Looks good.

Here some suggestions for enhancements (feel free to disagree or put to an issue for later investigation):

- Make sure, that the __init__ creates all attributes which are assumed by other functions to exist, e.g. `ProfileInterpolator.train_density` expected in `ProfileInterpolator.predict` but not existing for a fresh instance of the class.
- Document data structures a bit more, e.g. which are the scalars nested in an additional list, see lines 118-123.
- Think, whether you really need two load functions, e.g. `ProfileInterpolator.load_profiles` vs. `ProfileInterpolator.load`. In case you keep both, the doc-strings should tell the user clearly, when to use which of them and may think about building one up on the other and don't use different imports when loading the same file type.

philipp-rajah (approved):
I am not sure what the general expectation for reviews is but I know that this is time sensitive so I will approve it. I have not gone through the process of running the code so I will assume that that has been done. 

In general everything here makes sense to me given it is implemented how it is written in the paper. However it would be nice to see some more modularization in the future especially for the network architecture which is hard coded. Something else to consider is normalization which is not carried out in this implementation but is widely used in ML as standard practice. A final thought is in regards to an evaluation pipeline which I didn't see and might be nice to include for anyone else looking to use or assess these models.

PR #399, opened 2024-09-26
Update IF_interpolation.py

philipp-rajah:
Updated k to be a minimum of 3 to prioritize smoothness

### DIFF
--- a/posydon/interpolation/IF_interpolation.py
+++ b/posydon/interpolation/IF_interpolation.py
@@ -1430,7 +1430,10 @@ class KNNClassifier(Classifier):
                     acc[n - 1] += balanced_accuracy_score(
                         yT[itest], ki.predict(XT[itest, :])) / nfolds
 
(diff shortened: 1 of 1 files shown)


mkruckow (dismissed):
Usually, the change should be fine as long as we are sure that `balanced_accuracy_score` won't always return 0 for all our `XY` and `yT`. In the case this can happen this change won't guarantee `n_opt>2`, because `np.argmax` will return the lowest index, which is 0, in case all entries of `acc` keep zero. Thus a saver way would be to change `n_opt = np.argmax(acc) + 1` to `n_opt = np.argmax(acc[2:]) + 3`.

philipp-rajah:
@mkruckow I think if that happens then we have bigger issues to worry about but in principle you are correct. I just made the changes, thanks for catching that edge case.

mkruckow (approved):

maxbriel (approved):

elizabethteng (approved):

PR #403, opened 2024-10-02
small fix to BBH selection function

maxbriel:
The build-in selection function crashed due to the name change of the spin_orbit_tilt parameter. This changes the name for this function.

### DIFF
--- a/posydon/popsyn/transient_select_funcs.py
+++ b/posydon/popsyn/transient_select_funcs.py
@@ -193,7 +193,7 @@ def BBH_selection_function(history_chunk, oneline_chunk, formation_channels_chun
     return df_transients
     
 
-def DCO_detactability(sensitivity, transient_pop_chunk, z_events_chunk, z_weights_chunk, verbose=False):
+def DCO_detectability(sensitivity, transient_pop_chunk, z_events_chunk, z_weights_chunk, verbose=False):
     '''Calculate the observability of a DCO population.
     
     Parameters
@@ -208,6 +208,18 @@ def DCO_detactability(sensitivity, transient_pop_chunk, z_events_chunk, z_weight
         
         GW detector sensitivity and network configuration you want to use, see arXiv:1304.0670v3
         detector sensitivities are taken from: https://dcc.ligo.org/LIGO-T2000012-v2/public
+        
+    Note: The population must have the following columns:
+    - S1_mass : the mass of the first BH
+    
+    For the following columns, the function will try to calculate them if they are not present:
+    - q : the mass ratio
+    - chi_eff : the effective spin of the BHs
+    
+    For q, S2_mass must be present.
+    For chi_eff, S1_Mass, S2_mass, S1_spin, S2_spin, S1_spin_orbit_tilt_at_merger, S2_spin_orbit_tilt_at_merger must be present.
+
+    These have to be present and a valid value. If not, the function will raise an error!        
     
     '''
     available_sensitiveies = ['O3actual_H1L1V1', 'O4low_H1L1V1', 'O4high_H1L1V1', 'design_H1L1V1']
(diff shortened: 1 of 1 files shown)


celiotine (dismissed):

elizabethteng (dismissed):

posydon/popsyn/transient_select_funcs.py
```diff
@@ -179,12 +179,12 @@ def BBH_selection_function(history_chunk, oneline_chunk, formation_channels_chun
     df_transients['S2_mass'] = history_chunk[mask]['S2_mass']
     df_transients['S1_spin'] = history_chunk[mask]['S1_spin']
     df_transients['S2_spin'] = history_chunk[mask]['S2_spin']
-    df_transients['S1_spin_orbit_tilt'] = oneline_chunk['S1_spin_orbit_tilt']
-    df_transients['S2_spin_orbit_tilt'] = oneline_chunk['S2_spin_orbit_tilt']
+    df_transients['S1_spin_orbit_tilt'] = oneline_chunk['S1_spin_orbit_tilt_second_SN']
+    df_transients['S2_spin_orbit_tilt'] = oneline_chunk['S2_spin_orbit_tilt_second_SN']
```
mkruckow: The naming changed because we split it into `spin_orbit_tilt_first_SN` and `spin_orbit_tilt_second_SN`. Thus, should we include the tilt from the first SN here as well?
Is there a reason, why the name in the `df_transients` didn't get updated as well?

posydon/popsyn/transient_select_funcs.py
```diff
@@ -179,12 +179,12 @@ def BBH_selection_function(history_chunk, oneline_chunk, formation_channels_chun
...
     df_transients['orbital_period'] = history_chunk[mask]['orbital_period']
     df_transients['chirp_mass'] = m_chirp(history_chunk[mask]['S1_mass'], history_chunk[mask]['S2_mass'])
     df_transients['mass_ratio'] = mass_ratio(history_chunk[mask]['S1_mass'], history_chunk[mask]['S2_mass'])
-    df_transients['chi_eff'] = chi_eff(history_chunk[mask]['S1_mass'], history_chunk[mask]['S2_mass'], history_chunk[mask]['S1_spin'], history_chunk[mask]['S2_spin'], oneline_chunk['S1_spin_orbit_tilt'], oneline_chunk['S2_spin_orbit_tilt'])
+    df_transients['chi_eff'] = chi_eff(history_chunk[mask]['S1_mass'], history_chunk[mask]['S2_mass'], history_chunk[mask]['S1_spin'], history_chunk[mask]['S2_spin'], oneline_chunk['S1_spin_orbit_tilt_second_SN'], oneline_chunk['S2_spin_orbit_tilt_second_SN'])
```
mkruckow, line 187: There is another calculation of `chi_eff` in lines 232-237, shouldn't we update the naming there, too?
I guess, there we'd only need the update, when we'd update the key in the `df_transients`, is there a reason to not use there the same naming as in the oneline?

maxbriel: That one actually depends on the transient_population value. If the values are not present inside population, this function will not work anyway.

I will add additional docstring info to clarify the requirements to use these functions.

posydon/popsyn/transient_select_funcs.py
maxbriel: I kept them different, because for the BBH selection we only care about the `spin_orbit_tilt` at merger, so I kept that one naming convention here.

mkruckow: How about calling them then `S1_spin_orbit_tilt_at_merger`?
So, the user will know that it's already taken care of using the appropriate one from the oneline.

maxbriel: Sounds good! Done!

mkruckow (dismissed):

maxbriel:
Tutorials updated in PR #406

celiotine (approved):

elizabethteng (approved):

mkruckow (approved):
WIth the updates of the documentation in PR #406

sgossage (approved):

PR #404, opened 2024-10-02
rename "stripped_He_Core_H_burning" state and add to flow chart

celiotine:
The star state 'stripped_He_Core_H_burning' was previously missing in the flow chart. This was contributing to some of the problems discussed in Issue #388.

### DIFF
--- a/posydon/binary_evol/DT/step_detached.py
+++ b/posydon/binary_evol/DT/step_detached.py
@@ -1395,7 +1395,6 @@ class detached_step:
 
         if (ev_rlo1(binary.time, [binary.separation, binary.eccentricity]) >= 0
                 or ev_rlo2(binary.time, [binary.separation, binary.eccentricity]) >= 0):
-            print("setting initial_RLOF 5")
             binary.state = "initial_RLOF"
             return
             
--- a/posydon/binary_evol/DT/step_detached.py
+++ b/posydon/binary_evol/DT/step_detached.py
@@ -40,7 +40,7 @@ import posydon.utils.constants as const
 from posydon.utils.posydonerror import NumericalError, MatchingError, POSYDONError
 from posydon.utils.posydonwarning import Pwarn
 
-LIST_ACCEPTABLE_STATES_FOR_HMS = ["H-rich_Core_H_burning"]
+LIST_ACCEPTABLE_STATES_FOR_HMS = ["H-rich_Core_H_burning", "accreted_He_Core_H_burning"]
 
 LIST_ACCEPTABLE_STATES_FOR_postMS = [
     "H-rich_Shell_H_burning",
@@ -48,8 +48,7 @@ LIST_ACCEPTABLE_STATES_FOR_postMS = [
     "H-rich_Central_He_depleted",
     "H-rich_Core_C_burning",
     "H-rich_Central_C_depletion",
-    "H-rich_non_burning",
-    "accreted_He_Core_H_burning"]
+    "H-rich_non_burning"]
 
 LIST_ACCEPTABLE_STATES_FOR_HeStar = [
     'stripped_He_Core_He_burning',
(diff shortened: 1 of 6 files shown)


celiotine:
Should we also remove "stripped_He_Core_H_burning" from the list of H-rich star states (line 50 in the flow chart)?

mkruckow:
> Should we also remove "stripped_He_Core_H_burning" from the list of H-rich star states (line 50 in the flow chart)?

I think, we shouldn't have that state as it is named right now, because it is not a consequence of stripping, but of accretion. As it is caused by accretion, the overall star will still be H-rich.

posydon/binary_evol/DT/step_detached.py
```diff
@@ -48,7 +48,8 @@
     "H-rich_Central_He_depleted",
     "H-rich_Core_C_burning",
     "H-rich_Central_C_depletion",
-    "H-rich_non_burning"]
+    "H-rich_non_burning",
+    "accreted_He_Core_H_burning"]
```
mkruckow: Shouldn't we add this to the `LIST_ACCEPTABLE_STATES_FOR_HMS` instead of `LIST_ACCEPTABLE_STATES_FOR_postMS`, because the `Core_H_burning` should be MS.

posydon/visualization/plot_defaults.py
```diff
@@ -404,7 +404,7 @@
             ['o', 2, 'tab:brown', 'stripped He-star C depletion'],
         'stripped_He_non_burning':
             ['o', 2, 'gray', 'stripped He-star non burning'],
-        'stripped_He_Core_H_burning':
+        'accreted_He_Core_H_burning':
             ['x', 1, 'black', 'unknown'],
```
mkruckow: Please keep the old entry and add a new one for `accreted_He_Core_H_burning`, which is a known state. (The same applies to the other case, where you replaced `stripped_He_Core_H_burning` in the plot defaults.

celiotine: Didn't we remove the "stripped_He_Core_H_burning" state so it no longer exists? So we can replace this and then yes also edit the label too

celiotine:
Here is an example of how this evolution looks for systems with accreted_He_Core_H_burning stars that are successfully matched in the detached step:
Screen Shot 2024-10-08 at 1 26 21 PM

posydon/visualization/plot_defaults.py
mkruckow: It is not removed. We replaced it in the cases, where it was wrongly assigned.
It is still possible in a non-physical situation, hence getting the unknown label in the plotting.

posydon/binary_evol/flow_chart.py
```diff
@@ -74,7 +75,7 @@
 # we infer the state, if core_definition_H_fraction=0.1 then surface_h1=0.1
 # and the state is H-rich_non_burning which we stil want to evolve thorugh
 # the step_CO_HeMS
-STAR_STATES_HE_RICH_EVOLVABLE.append('H-rich_non_burning')
+STAR_STATES_HE_RICH_EVOLVABLE.extend(['H-rich_non_burning', 'accreted_He_non_burning'])
```
mkruckow: According to the comment, I'd not add `accreted_He_non_burning` here, because this shouldn't be a post-CE state and surface_h1=0.1 won't lead to `accreted_He`.
But maybe it is needed when using 1percent for the core boundary. Could you please recheck this?

celiotine:
Here is an example of how this evolution looks for systems with an "accreted_He_Core_He_burning star" (run with IF):
Screen Shot 2024-10-14 at 1 14 25 PM

And "accreted_He_non_burning" (run with IF):
Screen Shot 2024-10-14 at 1 15 59 PM

mkruckow (approved):

maxbriel (approved):

PR #405, opened 2024-10-03
Add "last" to TF12 for stable MT

mkruckow:
- [x] Add "last" to TF12 for stable MT

### DIFF
--- a/posydon/visualization/plot_defaults.py
+++ b/posydon/visualization/plot_defaults.py
@@ -466,50 +466,50 @@ DEFAULT_MARKERS_COLORS_LEGENDS = {
         'Stable contact':
             ['s', 2, list_of_colors[3], 'Stable contact phase'],
         'Stable case A':
-            ['s', 2, list_of_colors[2], 'Stable RLOF during MS'],
+            ['s', 2, list_of_colors[2], 'Last stable RLOF during MS'],
         'Stable case B':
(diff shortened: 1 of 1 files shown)


maxbriel (approved):

sgossage:
This looks good, but I'm wondering whether we should make the other "termination flags" more descriptive as well. At the moment users need to keep in mind or look up in our table/docs what each TF means. Should we add a summary label under each TF's legend title? Something like

1. MESA termination_code
2. RLO state history
3. Final stellar state of S1
4. Final stellar state of S2

and for TF12 something like "Final MT state"

or might this be too crowded?

mkruckow:
You can already do that by changing the `PLOT_PROPERTIES['legend2D']['title']`, when you make a plot.

Surely, we can think of setting up there some defaults for plots created with the pipeline, but that's outside the scope of this PR, because this would more be an update on the pipeline instead of the default keys in the legend. Feel free to open an issue and assign me. Additionally, may think about whether there are other PLOT_PROPERTIES, we'd like to change for the plots created by the pipeline.

sgossage:
OK -- I've opened [issue 411](https://github.com/POSYDON-code/POSYDON/issues/411).

My last comment is that this change may become redundant if we do introduce default legend labels, but I'm fine with sorting that out later as well.

sgossage (approved):

Issue #413, opened 2024-10-15
mislabeling of "BH" remnants as "NS"


celiotine:
I thought this issue had previously been fixed, but I am finding that SN remnants that should definitely be BHs (mass >> 2.5 Msun) are being labeled as NSs, and then this label is changed later-on in the history:

<img width="767" alt="Screen Shot 2024-10-15 at 11 14 50 AM" src="https://github.com/user-attachments/assets/61b1a3d6-e35a-482f-9e4d-8a43f466a5fd">

Here are the initial conditions for re-running this binary with `/projects/b1119/cliotine/pulsar_dev/pulsar_pops/runs/oct_2024/Sukhbold+16/grid_reruns/params.ini`:

```
sim_kwargs_IF = simprop_kwargs_from_ini(ini_filepath_IF)
sim_prop_IF = SimulationProperties(**sim_kwargs_IF)
sim_prop_IF.load_steps(verbose=True)

star_1 = SingleStar(**{'mass':  16.563968, 'state': 'H-rich_Core_H_burning',
                'natal_kick_array': [250.739341, 0.378998, 1.529432, 1.176576]})
star_2 = SingleStar(**{'mass': 10.345701, 'state': 'H-rich_Core_H_burning',
                'natal_kick_array': [34.498945, 0.283107, 2.017168, 2.053916]})
binary_IF = BinaryStar(star_1, star_2, **{'time': 0.0,  'state': 'detached',  'event': 'ZAMS', 
                        'orbital_period': 5.995067,'eccentricity': 0.0}, properties = sim_prop_IF)
binary_IF.evolve()
```
This is running with IF and using IF to get the SN remnants.

celiotine:
After discussing with @ZepeiX, it might be possible the new interpolators for the most recent grids were trained with an old pipeline script, I am looking into this now (should have been resolved w/ Issue #303).

Also, it is unclear to me if this issue is different than that in Issue #345. There are more of these systems in populations run with the 241011 grids compared to the 240920 grids, and BHs being mislabeled as NSs are only present in populations run with the 241011 grids.

celiotine:
temporarily resolved with PR #416, but classification constraints still need to be added to the interpolator re discussion with @maxbriel @mkruckow @philipp-rajah

Issue #345, opened 2024-06-27
step_disrupted does not update stellar/CO state


maxbriel:
The detached step does not check for the physical accuracy of its components:
`step_end` calls `update_star_states` to update the state of S1 to the correct type.
It looks like the step_disrupted does not update the stellar state

Expected behaviour: S1_state is updated when entering step_disrupted.

Potential solution: Force a update_star_states of CO_type, when entering the `disrupted/detached_step` or in `step_SN`.

| Binary Index | S1 State | S2 State | S1 Mass | S2 Mass | Event | State | Orbital Period |
| --- | --- | --- | --- | --- | --- | --- | --- |
| 532 | H-rich_Core_H_burning | H-rich_Core_H_burning | 26.213338 | 24.474420 | ZAMS | detached | 579.504315 |
| 532 | stripped_He_Central_C_depletion | H-rich_Core_He_burning | 9.193318 | 13.938683 | CC1 | detached | 1381.985011 |
| 532 | BH | H-rich_Core_He_burning | 1.423424 | 13.938683 | NaN | disrupted | NaN |
| 532 | BH | stripped_He_Central_C_depletion | 1.423424 | 9.171099 | CC2 | disrupted | NaN |
| 532 | BH | NS | 1.423424 | 1.430361 | NaN | disrupted | NaN |
| 532 | NS | NS | 1.423424 | 1.430361 | END | disrupted | NaN |

celiotine:
@maxbriel do you have the full initial conditions of this binary for testing purposes? I will look into addressing this issue.
 Also, it looks like there may be a bigger classification issue here that we have discussed before (I think), so I assume this is meant to be an extra check/fix on top of a bigger fix? Ideally we wouldn't have 1.4Msun objects being labeled as BHs in the first place.

maxbriel:
@celiotine What additional info do you want? I believe the metallicity was at solar.

Looking at the original code, I believe that in step_end, `binary.update_star_states()` is called, which correctly sets the type. This is never done when entering the step_disrupted for the first time.

It probably has something to do with the interpolator, since this was from an IF population.
The IF interpolation is done using the CO_interpolation_class, while the state comes from the CO_type classifier.
See here: https://github.com/POSYDON-code/POSYDON/blob/251b3c6d895e0ef19169f2f34af968937c53b308/posydon/binary_evol/MESA/step_mesa.py#L965-L983


where `fv` is defined as the final_values output of the IF interpolator.

[...]

PR #416, opened 2024-10-16
update star states in step_SN

celiotine:
This PR addresses Issue #345 and Issue #413 by updating the binary stellar states after core collapse in step_SN.
Here is an example of a binary where the stellar state was not updated previously:
Screen Shot 2024-10-16 at 12 34 39 PM
And now with the fix:
Screen Shot 2024-10-16 at 12 35 00 PM
This appears to fix all mislabeling of SN remnants in a small test population of binaries, I am running a larger population to be sure.


maxbriel (approved):
This looks good to me 👍

mkruckow (approved):

PR #417, opened 2024-10-18
Pulsational Pair-Instability top-down mass loss approach

maxbriel:
I implemented the prescription by [Hendriks et al. (2023)](https://ui.adsabs.harvard.edu/abs/2023MNRAS.526.4130H/abstract) for pulsational pair-instability mass loss before remnant mass determination.

It creates a 'pre-explosion' stellar structure and then calculates the remnant mass using the chosen remnant mass prescription.
Because of this, the user can pick their own remnant mass prescription without worrying about whether the remnant mass determination from the PPI prescription matches the core-collapse SN determination.
PISN occurs when the remnant mass is below 10 Msun, as per the paper.
It also allows for shifting of the CO mass regime and additional mass loss.

One thing I would like some input on is the implementation of `hydrogen_conserve=True`.
I applied this for this prescription because I wanted to see the effect of the additional hydrogen present if PPI was not able to unbind that material.  But the prescription is made with helium stars!

In [Woosley (2017)](http://arxiv.org/abs/1608.08939) only in the lowest mass models where pulsation pair-instability occurs, the outer hydrogen envelope is able to stay. But above M_CO~35 Msun, this is fully ejected.
As such, removing the `hydrogen_conserve` here feels okay.
Should I add an additional parameter for the hydrogen_conserved in the (P)PISN prescription?


posydon/binary_evol/SN/step_SN.py
```diff
@@ -927,6 +930,55 @@ def PISN_prescription(self, star):
...
+                delta_M_PPI_extra_ML = self.PPI_extra_mass_loss if self.PPI_extra_mass_loss is not None else 0.0
+                if ((m_CO_core >= 38 + delta_M_CO_shift) 
+                    and m_CO_core <= 114 + delta_M_CO_shift):
+
+                    delta_M_PPI = (
+                        (0.0006 * np.log10(star.metallicity * const.Zsun) + 0.0054)
+                        * (m_CO_core - delta_M_CO_shift - 34.8)**3
+                        - 0.0013 * (m_CO_core - delta_M_CO_shift - 34.8)**2
+                        + delta_M_PPI_extra_ML
+                    )
```
mkruckow, line 965: - May add a comment here, which equation this is in the paper.
- May consider to define a variable for `m_CO_core - delta_M_CO_shift`, which looks to be often used (incl. the condition) and the relevant mass.
- Shell we limit the metallicity? Because currently for `Z -> 0` we get `delta_M_PPI -> - infinity`. In the paper it looks like they only considered log10(Z)>=-4. An alternative could be log10(Z)>−8.9818, because there the first derivative at the upper edge (114Msun) becomes 0, thus ensuring to have the minimum of the function inside the mass range. May even more log10(Z)>−8.549 to keep the minimum before of the lower edge (38Msun).

posydon/binary_evol/SN/step_SN.py
```diff
@@ -927,6 +930,55 @@ def PISN_prescription(self, star):
...
+                    else:
+                        m_PISN = m_He_core
+                else:
+                    if self.conserve_hydrogen_envelope:
+                        m_PISN = m_star - delta_M_PPI
+                    else:
+                        m_PISN = m_He_core - delta_M_PPI
+                    
+                    if m_PISN < 0.0:
+                        m_PISN = np.nan
```
mkruckow, line 985: Isn't this the case, where the PI leaves no remnant? Hence, shouldn't the following calculation of `m_rembar` (lines 970-981) only be done otherwise (having it in an else case)?
At the moment the calculation of `m_rembar` can overwrite the full disruption of the star.

posydon/popsyn/population_params_default.ini
```diff
@@ -212,7 +212,13 @@
     # 'N20' for 'Sukhbold+16-engine',
     # 'Patton&Sukhbold20-engine' or None for the others
   PISN = "Marchant+19"
-    # None, "Marchant+19"
+    # None, "Marchant+19", "Hendriks+23"
+  PISN_CO_shift = 0.0
+    # float (-inf,inf)
+    # Only when using Hendriks+23
+  PPI_extra_mass_loss = 0.0
+    # float (0,inf)
```
mkruckow, line 220: At the moment the code doesn't ensure that the extra mass loss is not negative. May add a check for this in the code.

mkruckow:
About your questions:

1. In case `hydrogen_conserve=True`, throw a warning. Explaining, what is not done accurately.
2. No, I'd suggest to not introduce another flag. With 1. the user should know, what's happening. I think, there is no good reason for a case to have the hydrogen conservation flag on for some SN parts but not for others depending on mass.

maxbriel:
From a physics perspective, it is reasonable to not conserve the hydrogen for the pulsational pair-instability, since these will eject the envelope before the actual core collapse takes place. For stars not undergoing PPI, you could expect their envelope to stay. Since it's a different mechanism for removing the hydrogen, it feels reasonable to implement them separately.

posydon/binary_evol/SN/step_SN.py
maxbriel: - paper equation added.
- I think this is closer/nicer to see original equation.
- Limited to 1e-4 as per paper

posydon/binary_evol/SN/step_SN.py
maxbriel: added an `else` statement although the `m_rembar` is set to `m_PISN` inside the `compute_m_rembar` so it only speeds it up :)

posydon/popsyn/population_params_default.ini
maxbriel: This would be a good one to add for when we redo the "run setup"

posydon/binary_evol/SN/step_SN.py
mkruckow: Beside that it doesn't rely on the treatment of the `compute_m_rembar` function, which may change in the future.

mkruckow (dismissed):

maxbriel:
In the end, I decided to add a new parameter for hydrogen conservation during PPI.
Since this is a separate physical process from hydrogen conservation during core-collapse, this feels reasonable and is something we required for a project.

mkruckow (dismissed):

posydon/binary_evol/SN/step_SN.py
```diff
@@ -927,6 +931,66 @@ def PISN_prescription(self, star):
...
+            elif self.PISN == 'Hendriks+23':
+                # Hendriks et al. 2023 PISN prescription
+                # Shifting PPI and PISN gap
+                # works by removing delta_M_PPI from the star
+                # and then applying any remnant mass prescription
+                
+                delta_M_CO_shift = self.PISN_CO_shift if self.PISN_CO_shift is not None else 0.0
+                delta_M_PPI_extra_ML = self.PPI_extra_mass_loss if self.PPI_extra_mass_loss is not None else 0.0
+                if ((m_CO_core >= 38 + delta_M_CO_shift) 
+                    and m_CO_core <= 114 + delta_M_CO_shift):
```
astroJeff: These numbers: 38 Msun to 114 Msun are hard coded in. I think I saw they are listed in the Hendriks paper, but consider adding another two variables, something like `m_CO_core_PISN_min` and `m_CO_core_PISN_max`. It's cumbersome, but if they are defined up at the top of the function, then it makes it a bit clearer what's going on.

posydon/binary_evol/SN/step_SN.py
```diff
@@ -927,6 +931,66 @@ def PISN_prescription(self, star):
...
+                        if PISN_star.he_core_mass > m_PISN:
+                            PISN_star.he_core_mass = m_PISN
+                        if PISN_star.co_core_mass > m_PISN:
+                            PISN_star.co_core_mass = m_PISN
+                        m_rembar, _, _ = self.compute_m_rembar(PISN_star, m_PISN)
+                    
+                        if m_rembar < 10:
+                            m_PISN = np.nan
+                        else:
+                            m_PISN = m_rembar
```
astroJeff, line 998: Does this come from Hendriks et al.?

maxbriel: Yes. They define as any remnant mass below 10 to be a PISN

posydon/binary_evol/SN/step_SN.py
maxbriel: Do you just mean like commit f475744?

mkruckow (dismissed):

mkruckow (approved):

sgossage (approved):

Issue #370, opened 2024-08-15
remove "redirect" steps from history dataframe


celiotine:
Per our discussion at the developers meeting today as well as in the [slack](https://posydon.slack.com/archives/CDCS3MS31/p1723489363940419), we agree the history output would be more clear to users if the "redirect" steps are not included in the history dataframe. Here is an example of what the evolution for a redirected binary currently looks like:
<img width="760" alt="Screen Shot 2024-08-15 at 10 26 35 AM" src="https://github.com/user-attachments/assets/82470c19-1fa8-4938-80ea-e9336b4ee6fd">

celiotine:
Issue addressed in PR #421

PR #421, opened 2024-10-22
remove "redirect" steps from history dataframe

celiotine:
This PR addresses Issue #370. It removes all "redirect" steps from the history dataframe to clarify the binary evolution for users. The redirect steps still exist internally (e.g. in the flow), only now they are simply not saved in the binary history.

Here is an example history of a binary that was previously undergoing a redirect step:
Screen Shot 2024-10-22 at 1 57 06 PM

And this is what the history looks like when running the same binary with this PR:
Screen Shot 2024-10-22 at 1 54 16 PM


celiotine:
Per discussion on 10/24: add "history_verbose" option as binary attribute if people want to keep the redirect steps in the output

mkruckow:
Please add the `history_verbose` to our [default ini](https://github.com/POSYDON-code/POSYDON/blob/camille_remove_redirect_from_df/posydon/popsyn/population_params_default.ini).

posydon/binary_evol/simulationproperties.py
```diff
@@ -255,13 +253,21 @@ def pre_step(self, binary, step_name):
 
     def post_step(self, binary, step_name):
         """Record the duration of the step."""
+
+        ## do not record redirect step times if history_verbose=False
+        if not binary.history_verbose and binary.event is not None:
+            if "redirect" in binary.event:
+                return binary
```
mkruckow: I wondering, whether it is the right place here. Because it requires to repeat it for every hook.
The other option would be to put it up one level (in `post_step` of `SimulationProperties`), and apply it before the `post_step` function of any gets called.

mkruckow:
For discussion: Should we have a record in the oneline when history entries got skipped? I'd more tend to yes. If so, how?

1. Putting up a POSYDON warning, hence getting it included in the warnings records. This goes with the assumption, that a redirect is a non well treated behavior. (Most simple)
2. Potentially adding a new field to the oneline, e.g. badnesses (in analogy to badboxes in LaTeX), where we record it. (Giving it less priority then a warning)

posydon/binary_evol/simulationproperties.py
celiotine: I would think we don't want to call this before every hook, because hooks could be doing anything. For example, what if someone creates a hook meant to do something with the binary _only_ for redirect steps?

mkruckow: Wouldn't someone writing a hook to act on the redirect steps want to have `history_verbose=True` anyways?

So, my motivation is: Who should take care about leaving/removing the redirect steps in the coding? A person writing a random hook or a person writing a hook on redirects?

mkruckow: If keeping it to be added to any hook, we should get a documentation/instruction for this, so that everyone writing a hook knows about.

celiotine: Ok, from that perspective I think your suggested change makes sense. I think that the instances where people will create hooks completely independent of the history will be rare.

celiotine:
@mkruckow I think adding a warning may potentially create confusion as I believe @astroJeff mentioned, because nothing about omitting the history steps from the output is incorrect with regards to the binary evolution. So, it would be difficult to determine from the oneline if something strange happened with the evolution, or if simply part of the history output was not written to the dataframe.
If we do add some record to the oneline, I would create a separate field. I am also wondering if this is needed, as someone who really cares could just set history_verbose=True with little change in the runtime/output file size?

posydon/binary_evol/simulationproperties.py
```diff
@@ -147,6 +147,11 @@ def pre_step(self, binary, step_name):
         binary : instance of <class, BinaryStar>
 
         """
+        ## do not call extra step hooks if history_verbose=False
+        if not binary.history_verbose and binary.event is not None:
+            if "redirect" in binary.event:
+                return binary
```
mkruckow: Are you sure, we need this here? I have my doubts, that `binary.event` is well defined pre step?
I'd guess, it's either undefined or copied from the previous step, hence you'd may skip the pre_step in case the previous step was a redirect. Please, double check whether this is intended behavior.

maxbriel:
@celiotine Thank you for this PR! Sorry for only getting around to it now!
I agree with you that the warning component adds confusion. If the user wants to verify they can turn on `history_verbose`.

One thing I'm confused about is how we treat pre and post-step hooks in the `history_verbose=False` case.
We're still moving through the step despite being redirected, so I feel like we should apply the pre/post steps + hooks.
However, understand that this will probably cause issues with the actual size of the final data frame, right?
Or is the point that this is not actually the step where we should be and it shouldn't exist at all?

Would implementing it in another location maybe be easier? For example, in the `BinaryStar.to_df()` function as a mask on the "full history"?
A cleaned-up data frame is given here:
https://github.com/POSYDON-code/POSYDON/blob/532f620557cf1b197ee731386aea81f59914333d/posydon/binary_evol/binarystar.py#L439-L443

Based on "history_verbose", you can mask and achieve a similar output data frame without changing the treatment of the steps.

```
if self.history_verbose:
     return binary_df
else:
    mask = ~ binary_df['event'].str.contains('redirect')
    return binary_df[mask]
```

If I missed something, please let me know. I haven't followed the discussion on this issue closely in the past weeks.

celiotine:
Thanks for the comment @maxbriel! Yes, the reason we were adding a check for history_verbose in the step hooks is because the hooks generally alter the history dataframe, e.g. the step_names and step_times hooks add those respective columns to the history. While those hooks do check if there is a mismatch in the history length, the final output will be incorrect if we still execute the full step hook. 

Your suggestion of doing this with a mask in the to_df() method is interesting! I think it is a reasonable suggestion, and I will think about if there are any drawbacks to treating things this way. If not, this may be the preferable way to do it.

mkruckow:
Here some ideas from my side:
One drawback is, that we keep the redirect stuff longer in memory (@maxbriel you should know best, whether this is important or not). Another drawback could be that we fix it on a binarystar class, hence we'd disallow any evolution e.g. with a singlestar class outside a binary, while the hooks go on the step/flow level.

posydon/binary_evol/simulationproperties.py
celiotine: I just checked and it looks like the binary event is not updated when the pre-step hooks are called (it has the event from the previous step), so I have removed this check from the hooks.

mkruckow (dismissed):

mkruckow:
One more useful addition could be to update the documentation accordingly.

maxbriel:
The additional memory footprint is insignificant compared to the other usage.
A similar approach works for the Singlestar, but I don't understand why this is relevant for the redirects, which are only in the binary grids.
I don't think there's a wrong or right approach, but a choice of how we want it implemented. Maybe we can have a chat on Thursday about it.

maxbriel (approved):
In the end, this does seem like the way to go

maxbriel (dismissed):

mkruckow (approved):

maxbriel (approved):
LGTM! Thank you for adding the documentation info

Issue #418, opened 2024-10-21
Undefined binaries in flow: ('stripped_He_Central_He_depleted', 'H-rich_Core_H_burning', 'RLO1', 'oRLO1')


celiotine:
Binaries with the parameters ('stripped_He_Central_He_depleted', 'H-rich_Core_H_burning', 'RLO1', 'oRLO1') are failing because they are not handled in the flow. An example of one such binary:

```
ini_filepath_IF = "/projects/b1119/cliotine/pulsar_dev/pulsar_pops/runs/oct_2024/Sukhbold+16/grid_reruns_final/params.ini"
sim_kwargs_IF = simprop_kwargs_from_ini(ini_filepath_IF)
sim_prop_IF = SimulationProperties(**sim_kwargs_IF)
sim_prop_IF.load_steps(verbose=True)

star_1 = SingleStar(**{'mass':  6.30415516982906, 'state': 'H-rich_Core_H_burning',
                'natal_kick_array': [None, None, None, None]})
star_2 = SingleStar(**{'mass': 3.19348803304828, 'state': 'H-rich_Core_H_burning',
                'natal_kick_array': [None, None, None, None]})
binary_IF = BinaryStar(star_1, star_2, **{'time': 0.0,  'state': 'detached',  'event': 'ZAMS', 
                        'orbital_period': 585.7165771197014,'eccentricity': 0.0}, properties = sim_prop_IF)
binary_IF.evolve()
```
<img width="681" alt="Screen Shot 2024-10-21 at 1 49 18 PM" src="https://github.com/user-attachments/assets/64065277-3431-47d8-b48f-f28008d992fc">

This is happening when running with IF but not NN.

celiotine:
Possibly related systems, though they are much rarer (these binaries also leave step_CE in RLO after step_HMS_HMS with an undefined state):

1)
 ('H-rich_Core_He_burning', 'H-rich_Shell_H_burning', 'RLO2', 'oRLO2')
```
star_1 = SingleStar(**{'mass':  6.629154054663811, 'state': 'H-rich_Core_H_burning',
                'natal_kick_array': [None, None, None, None]})
star_2 = SingleStar(**{'mass': 5.643884766762587, 'state': 'H-rich_Core_H_burning',
                'natal_kick_array': [None, None, None, None]})
binary_IF = BinaryStar(star_1, star_2, **{'time': 0.0,  'state': 'detached',  'event': 'ZAMS', 
[...]

PR #423, opened 2024-10-28
error catch for He-rich stars in RLO onto H-rich stars

celiotine:
This PR is a temporary fix for part of Issue #418. It adds an error catch in the detached step for binaries with an He-rich star in RLO onto an H-rich star, which account for <0.5% of all binaries at solar metallicity. Ideally we would have a grid to model this subpopulation of binaries, but for now we force them to fail with a POSYDON error.

Here is an example of the output for one such binary run with this PR:
Screen Shot 2024-10-28 at 3 06 20 PM
Screen Shot 2024-10-28 at 3 06 41 PM

### DIFF
--- a/posydon/binary_evol/DT/step_detached.py
+++ b/posydon/binary_evol/DT/step_detached.py
@@ -37,7 +37,7 @@ from posydon.utils.common_functions import (
 )
 from posydon.binary_evol.flow_chart import (STAR_STATES_CC, STAR_STATES_CO)
 import posydon.utils.constants as const
-from posydon.utils.posydonerror import NumericalError, MatchingError, POSYDONError, ModelError
+from posydon.utils.posydonerror import NumericalError, MatchingError, POSYDONError, FlowError
 from posydon.utils.posydonwarning import Pwarn
 
 LIST_ACCEPTABLE_STATES_FOR_HMS = ["H-rich_Core_H_burning", "accreted_He_Core_H_burning"]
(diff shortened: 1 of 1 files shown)


posydon/binary_evol/DT/step_detached.py
```diff
@@ -1926,6 +1926,19 @@ def get_star_profile(star, htrack, m0):
                     else:
                         binary.state = "RLO1"
                         binary.event = "oRLO1"
+                
+                if (binary.star_1.state in LIST_ACCEPTABLE_STATES_FOR_HeStar 
+                    and binary.star_2.state in STAR_STATES_H_RICH 
+                    and binary.state == "RLO1"):
+                        set_binary_to_failed(binary)
+                        raise ModelError("Evolution of He-rich stars in RLO onto H-rich stars after HMS-HMS not yet supported.")
```
mkruckow: I'd be more inclined to use a FlowError here.

posydon/binary_evol/DT/step_detached.py
```diff
@@ -1926,6 +1926,19 @@ def get_star_profile(star, htrack, m0):
                     else:
                         binary.state = "RLO1"
                         binary.event = "oRLO1"
+                
+                if (binary.star_1.state in LIST_ACCEPTABLE_STATES_FOR_HeStar 
+                    and binary.star_2.state in STAR_STATES_H_RICH 
+                    and binary.state == "RLO1"):
```
mkruckow: A performance improvement: I'd first check for the "RLO1" before checking the star states.

mkruckow (approved):

sgossage (approved):

PR #425, opened 2024-10-30
Automatic testing of installation

elizabethteng:
continuous_integration.yml: 
- is triggered for all PRs created on development or main
- tests on mac, ubuntu, windows
- runs python 3.11 (can add others)
- installs all requirements from setup.py
- I have currently commented out the option to run all tests in the unit_tests folder, but we can include that (or any combination of specific tests) later after Matthias' PR is accepted

install_extras.yml is same as above, except:
- is only triggered weekly by cronjob (takes much longer than regular installation)
- additionally installs all extras in setup.py: ml, hpc, docs, vis

### DIFF
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -10,7 +10,7 @@ jobs:
 
     strategy:
       matrix:
-        os: [macos-latest, ubuntu-latest, windows-latest]
+        os: [macos-latest, ubuntu-latest]
         python-version: ['3.11']  # can add more if we want to support
 
     steps:
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -17,10 +17,6 @@ jobs:
     - name: Checkout repository
       uses: actions/checkout@v2
 
-    - name: Normalize line endings on Windows
-      if: runner.os == 'Windows'
-      run: git config --global core.autocrlf false
-
     - name: Set up Python
       uses: actions/setup-python@v2
       with:
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -1,0 +1,36 @@
+name: POSYDON Continuous Integration
+
+on:
+  pull_request:
+    branches: [main, development]
+
+jobs:
+  test:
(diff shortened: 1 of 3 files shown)


.github/workflows/install_extras.yml
```diff
@@ -0,0 +1,32 @@
...
+  schedule:
+    - cron: "0 0 * * 0"  # Runs at 00:00 (UTC) every Sunday
+
+jobs:
+  test:
+    runs-on: ${{ matrix.os }}
+
+    strategy:
+      matrix:
+        os: [macos-latest, ubuntu-latest, windows-latest]
```
maxbriel: Do you also want to remove the windows option here?

.github/workflows/continuous_integration.yml
```diff
@@ -0,0 +1,36 @@
...
+        os: [macos-latest, ubuntu-latest]
+        python-version: ['3.11']  # can add more if we want to support
+
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v2
+
+    - name: Normalize line endings on Windows
+      if: runner.os == 'Windows'
+      run: git config --global core.autocrlf false
```
maxbriel: Is this still required?

.github/workflows/install_extras.yml
```diff
@@ -0,0 +1,32 @@
...
+        python-version: ['3.11']  # can add more if we want to support
+
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v2
+
+    - name: Normalize line endings on Windows
+      if: runner.os == 'Windows'
+      run: git config --global core.autocrlf false
```
maxbriel: Same as other comment :)

maxbriel (changes requested):
Looks good and I'm loving that it's already doing the installation test.
Just added some comments about the Windows components :)

maxbriel (approved):

maxbriel:
Thank you!

mkruckow (approved):
Look good.

Did you had a test, what happens when something fails?

elizabethteng:
> Look good.
> 
> Did you had a test, what happens when something fails?

Yes! I put a nonexistent package in the setup.py and both checks failed.

PR #428, opened 2024-11-07
update installation test to include M1 and intel chips for macos

elizabethteng:
add macos-14 github runner which uses the apple silicon M1 chip: https://github.blog/changelog/2024-01-30-github-actions-introducing-the-new-m1-macos-runner-available-to-open-source/

### DIFF
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -10,7 +10,7 @@ jobs:
 
     strategy:
       matrix:
-        os: [macos-latest, ubuntu-latest]
+        os: [macos-latest, macos-14, ubuntu-latest]
         python-version: ['3.11']  # can add more if we want to support
 
     steps:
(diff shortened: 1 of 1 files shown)


elizabethteng:
info about github runners [here](https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources)

elizabethteng:
may want to discuss github billing for extra minutes: https://docs.github.com/en/billing/managing-billing-for-your-products/managing-billing-for-github-actions/about-billing-for-github-actions#per-minute-rates

maxbriel:
Can we change this to `macos-13` and `macos-latest`?
Latest is already on the M1 chipset, `macos-13` allows testing on a x86 (intel) chipset. And this means we won't have to update it when new OS releases come out :)

elizabethteng:
good idea Max! I guess I had left open the possibility that later mac setups could go back to intel but now that I'm actually thinking about it it's unlikely to happen.

maxbriel (approved):
Great! Approved :)

mkruckow (approved):

Issue #419, opened 2024-10-22
Undefined binaries in flow: ('H-rich_Core_H_burning', 'H-rich_Core_H_burning', 'RLO2', 'oRLO2')


celiotine:
Binaries with the parameters ('H-rich_Core_H_burning', 'H-rich_Core_H_burning', 'RLO2', 'oRLO2') are failing because they are not handled in the flow. An example of one such binary:

```
ini_filepath_IF = "/projects/b1119/cliotine/pulsar_dev/pulsar_pops/runs/oct_2024/Sukhbold+16/grid_reruns_final/params.ini"
sim_kwargs_IF = simprop_kwargs_from_ini(ini_filepath_IF)
sim_prop_IF = SimulationProperties(**sim_kwargs_IF)
sim_prop_IF.load_steps(verbose=True)

star_1 = SingleStar(**{'mass':   6.263565862272097, 'state': 'H-rich_Core_H_burning',
                'natal_kick_array': [None, None, None, None]})
star_2 = SingleStar(**{'mass': 4.042250364920023, 'state': 'H-rich_Core_H_burning',
                'natal_kick_array': [None, None, None, None]})
binary_IF = BinaryStar(star_1, star_2, **{'time': 0.0,  'state': 'detached',  'event': 'ZAMS', 
                        'orbital_period':  2.3130949924692237, 'eccentricity': 0.0}, properties = sim_prop_IF)
binary_IF.evolve()
```
<img width="623" alt="Screen Shot 2024-10-22 at 9 38 00 AM" src="https://github.com/user-attachments/assets/283acebd-74cf-4f8b-9eaa-001ea0cf0c24">

celiotine:
Per discussion 10/24: this is most likely a bug with labeling the star type, should go to CC1 and form WD?

celiotine:
These binaries should be handled by PR #430.

celiotine:
In addition, I am wondering if something wonky is happening with the interpolation for these binaries? It is still unclear to me why they are entering RLO after HMS-HMS to begin with. This is the interpolation information for the binary above:

<img width="275" alt="Screen Shot 2024-11-19 at 2 41 35 PM" src="https://github.com/user-attachments/assets/bfeb7554-6158-4a31-b46e-78bfe5b13549">
 
So not sure if these systems should be entering CE?

maxbriel:
@celiotine Yep. That looks like another classifier inconsistency to me. I've seen similar things in this issue #402

[...]

PR #430, opened 2024-11-11
handle H-rich RLO binaries and clarify verbosity of detached step

celiotine:
The goal of this PR is to address the remaining failed binaries in Issue #418 that initiate RLO with H-rich binary components.
These systems leave step_CE with stripped_He primaries, which are then matched to H-rich stars in the detached step. 

This PR also cleans-up the matching code in the detached step and clarifies the verbose output for easier interpretation. This is an example of what running step_detached with verbose=True looks like with this PR:
Screen Shot 2024-11-11 at 2 28 28 PM


celiotine:
As a reminder, we added an error catch for stripped-He stars in RLO onto H-rich stars in PR #423, where presumably the primary stars were matched to other stripped-He stars

celiotine:
I've extended the error check for RLO binaries in the detached step to include:
H-rich stars in RLO onto He-rich stars
He-rich stars in RLO onto H-rich stars
H-rich stars in RLO onto H-rich stars
So, basically everything except He-rich stars in RLO onto He-rich stars, which we assume merge.
I matched the criteria for these binaries to that for the HMS_HMS_RLO step in the flow chart, and also added a check to see if this step exists in the flow chart (if it does, these binaries should be evolved correctly through that step).

So, this PR should now handle all the remaining failed binaries in Issue #418 and #419 until the HMS_HMS_RLO step is included in the POSYDON data.

In addition, I've fixed some minor issues in the detached step related to the "verbose" parameter and also tried to streamline some of the code to make it easier to read.

posydon/binary_evol/DT/step_detached.py
```diff
@@ -819,14 +818,14 @@ def get_MESA_labels(list_for_matching):
             # if matching is still not successful, set result to NaN:
             if (np.abs(sol.fun) > tolerance_matching_integration_hard or not sol.success):
                 if self.verbose:
-                    print("\nMatching result is NOT successful, with ",
-                          np.abs(sol.fun), ">", tolerance_matching_integration, "tolerance")
+                    print("\nMatching result is NOT successful, with tolerance ",
+                          np.abs(sol.fun), ">", tolerance_matching_integration_hard)
```
mkruckow, line 822: This corrections look needed, but it raised for me the question, whether there is a more fundamental problem in the matching here: For all the minimization tries always `tolerance_matching_integration` is used to reject a solution. Only after the last try we check for `tolerance_matching_integration_hard`. Why do we allow only the last try to pass on a different limit? E.g. in the case the solution in the first try was better then in the last one, but both above `tolerance_matching_integration` we accept the worse solution of the last try.
If we would take the solution for them first try, we could may avoid too many bad matches on the non-preferred grid.

posydon/binary_evol/DT/step_detached.py
```diff
@@ -857,12 +856,12 @@ def get_MESA_labels(list_for_matching):
...
-                    f'log_R = {star.log_R:.3f}',
-                    f'center_he4 = {star.center_he4:.4f}',
-                    f'surface_he4 = {star.surface_he4:.4f}',
-                    f'surface_h1 = {star.surface_h1:.4f}',
-                    f'he_core_mass = {star.he_core_mass:.3f}',
+                    f'mass = {star.mass:.3f}, ',
+                    f'log_R = {star.log_R:.3f}, ',
+                    f'center_he4 = {star.center_he4:.4f}, ',
+                    f'surface_he4 = {star.surface_he4:.4f}, ',
+                    f'\nsurface_h1 = {star.surface_h1:.4f}, ',
```
mkruckow: Why shell we have an additional line break here? Look arbitrary to me.

posydon/binary_evol/DT/step_detached.py
```diff
@@ -1886,21 +1805,18 @@ def get_star_profile(star, htrack, m0):
...
-                if (binary.state == "RLO1" 
-                    and binary.star_1.state in LIST_ACCEPTABLE_STATES_FOR_HeStar 
-                    and binary.star_2.state in STAR_STATES_H_RICH):
+                if ('step_HMS_HMS_RLO' not in BINARY_STEPS_ALL
+                    and ((binary.star_1.state in STAR_STATES_HE_RICH_EVOLVABLE 
+                         and binary.star_2.state in STAR_STATES_H_RICH_EVOLVABLE)
+                    or (binary.star_1.state in STAR_STATES_H_RICH_EVOLVABLE
+                         and binary.star_2.state in STAR_STATES_HE_RICH_EVOLVABLE)
+                    or (binary.star_1.state in STAR_STATES_H_RICH_EVOLVABLE
+                         and binary.star_2.state in STAR_STATES_H_RICH_EVOLVABLE))):
```
mkruckow: The `step_HMS_HMS_RLO` will only help for the case that both stars have a state in `STAR_STATES_H_RICH_EVOLVABLE`.
The cases where we have a `STAR_STATES_HE_RICH_EVOLVABLE` in there, the RLO grid of the HMS-HMS won't help either and the binaries still need to be send to failed.

posydon/binary_evol/flow_chart.py
```diff
@@ -91,6 +104,8 @@
     'initial_RLOF'
 ]
 
+BINARY_STATES_CC = BINARY_STATES_ALL.copy()
```
mkruckow, line 107: Should we allow a CC on an `initial_RLOF`? I'd say no.

posydon/binary_evol/flow_chart.py
```diff
@@ -388,6 +402,9 @@ def initial_eccentricity_flow_chart(FLOW_CHART=None, CHANGE_FLOW_CHART=None):
         # always take ZAMS binaries to step_detached
         if event == 'ZAMS' and state in BINARY_STATES_ZAMS: # check for event
             MY_FLOW_CHART[key] = 'step_detached'
+    
+    ## Add new step to binary step list
+    BINARY_STEPS_ALL.append('step_HMS_HMS_RLO')
```
mkruckow: That's the wrong place. This function is not called at the beginning of the simulation, but whenever the flow chart is requested.
I think the best place would be where the flow chart function is read in from the ini to modify the the steps list there.
(Maybe you can keep it here, but you should then add a check, that you only add the entry when it's not existing already otherwise this can become a memory leek.)

Edit: it might be best to create the list of steps in general at the place, where the ini file is read, hence after you know, which steps are enabled, in principle a user could decide to remove other steps.

posydon/binary_evol/DT/step_detached.py
celiotine: Yes, so I did try rewriting this so that the "hard" limit is tested against the _best_ matching solution of all the attempts. However, this was creating some new issues with the matching results, so I decided that reworking this should be done in a different PR.

I asked Manos about this "hard" limit and all he said was that it was meant to apply to the stripped_He stars to allow less of them to fail in the matching, so maybe going forward that is something we should try to implement correctly (in addition to reworking the matching for these stars in general as we discussed).

posydon/binary_evol/DT/step_detached.py
celiotine: It formats the output in a cleaner way, so that the lines are not too long and run over into the next line (I have tested this).

posydon/binary_evol/DT/step_detached.py
celiotine: So I did this to align with what is currently in the flow chart as you suggested, I saw that both the H and He-rich stars are being directed to step_HMS_HMS_RLO in the initial_eccentricity_flow_chart. If that is incorrect then it should also be changed.

posydon/binary_evol/flow_chart.py
celiotine: If that is the case we should change that in a different PR?

posydon/binary_evol/DT/step_detached.py
mkruckow: Maybe I wasn't clear enough, what I'd suggest as solution:
When can check against the stricter limit for each try. But instead of always replacing the solution by a subsequent try, I'd suggest to keep the better solution of the previous and current try. In this way we would compare the lest strict limit not against the last try only, but against the best try.
To do that, I'd not overwrite the data from the previous try in a new one, but instead create new variables in the new try and overwrite the previous try only if the new try has a better solution. So something like
```
sol = get_sol()
if sol.fun>tolerance:
    new_sol = get_sol2()
    if sol.fun>new_sol.fun:
        sol = new_sol
if sol.fun>tolerance:
    new_sol = get_sol3()
    if sol.fun>new_sol.fun:
        sol = new_sol
```
But yes, we can make a different PR for this, if so please open an issue.

posydon/binary_evol/DT/step_detached.py
mkruckow: This will depend on the display and the length of the values. Therefore I'd either suggest to allow for automatic instead of fixed line breaks or to set all the sizes. At the moment a different person or depending on the exact value you may get an automatic line break just short before the hard coded one.

posydon/binary_evol/DT/step_detached.py
mkruckow: Hm, yes the flow although redirects the He starts to that step. But in the step itself they directly fail, because they are not in `FOR_RLO_STATES`. Again, we can may address this in a different PR, if so please open an issue.

posydon/binary_evol/flow_chart.py
mkruckow: Sorry, I thought you newly introduced it here. But yes, if you only moved it, may open an issue to address it in a different PR.

posydon/binary_evol/DT/step_detached.py
celiotine: Yes, that is what I did! I can open an issue for it.

celiotine: Ok I've created an issue for this: #447

posydon/binary_evol/DT/step_detached.py
celiotine: I've created Issue #448 for this.

posydon/binary_evol/flow_chart.py
celiotine: Issue opened #449

posydon/binary_evol/flow_chart.py
celiotine: Ok, you will see in the changes that I ended-up getting the step names from the binary SimulationProperties attribute and creating a list of the names in the Simulation Properties class, rather than the flow.

celiotine:
After our discussion regarding Issue #419, I added a separate POSYDON error for binaries that are being misclassified and entering the detached step when they should be going to CE. However, someone should review this, especially as it relates to the HMS-HMS-RLO step.

posydon/unit_tests/utils/test_posydonerror.py
```diff
@@ -32,7 +32,7 @@ def SingleStar():
 class TestElements:
     # check for objects, which should be an element of the tested module
     def test_dir(self):
-        elements = ['BinaryStar', 'FlowError', 'GridError', 'MatchingError',\
+        elements = ['BinaryStar', 'ClassificationError', 'FlowError', 'GridError', 'MatchingError',\
```
mkruckow: The PR looks good now, there are some merge conflicts to resolve.

MINOR: Otherwise here is violation of the line length.

celiotine: Ok I have fixed the conflicts

mkruckow (dismissed):
Fine.
In case we like to stick with the aim of using a limit on the line length, we should make sure that PRs like this don't violate it, especially in files which have no violations in there yet (e.g. the unit test files).

celiotine:
I just fixed the line length if you want to approve again @mkruckow.
What is the agreed-upon formatting for line lengths in general? Did we decide this? If so I agree we should make an effort to follow this because it varies greatly throughout the code.

mkruckow (approved):

sgossage (approved):

PR #382, opened 2024-09-10
Start to get unit tests back: utils

mkruckow:
- create a new directory for unit tests (we will get a validation test later; together they will make the old test directory obsolete, but for the moment we keep the old files as reference)
- [x] [`README.md`](https://github.com/POSYDON-code/POSYDON/blob/matthias_tests/posydon/unit_tests/README.md) for the unit tests, to explain how to run them and where to find the documentation of the python `unittest` module.
- [x] template file `test_template.py` to provide general structure of new unit test files
- [x] `test_common_functions.py`
- [x] `test_configfile.py`
- [x] `test_constants.py`
- [x] `test_data_download.py`
- [x] `test_gridutils.py`
- [x] `test_ignorereason.py`
- [x] `test_limits_thresholds.py`
- [x] `test_posydonerror.py`
- [x] `test_posydonwarning.py`
- some improvements on the utils code:
  - `constants.py`: rename `rae2a` to `rad2a`
  - `ignorereason.py`: add `__authors__`
  - `posydonerror.py`: correct bug: `obj`->`self.objects`, edit docstrings and add type checks
  - `posydonwarning.py`: write destructor print to `stderr`, add `TypeError`s, create `stacklevel` is not in dictionary, skip coverage tests (`# pragma: no cover`) for failures of other modules
  - `data_download.py`: there was a deprecation warning from `progressbar` to not use `maxval` anymore but `max_value` instead.
  - `configfile.py`: use more detailed errors
  - `gridutils.py`: typo, deal with no path and no mass transfer rate
  - `common_functions.py`: several small corrections

Issue #187, opened 2023-11-08
Reformating and updating all docstrings + adding missing ones.


mkruckow:
With the automatic creation of the reference from the docstrings we need to check them:

- check format:
  - here I already checked the format of citations
  - mainly checking the correct indentation, missing black lines, and other warnings and errors
- update information, e.g. some have more parameters then explained, or some contain outdated information or the information is too much/less detailed
- add missing ones

astroJeff:
Need to identify and check the docstrings throughout the code. @elizabethteng can you help with this?

PR #433, opened 2024-11-12
replace scipy interp1d with alternatives

### LINKED ISSUE #408: Scipy interp1d now deprecated
interp1d is a [legacy class](https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.interp1d.html#scipy.interpolate.interp1d) and will no longer be updated. Scipy [recommends](https://docs.scipy.org/doc/scipy/tutorial/interpolate/1D.html#tutorial-interpolate-1dsection) switching to numpy.interp for basic linear interpolation.

celiotine:
This PR addresses Issue #408, which recognizes that the scipy interp1d function is now deprecated.
Here, we replace instances of interp1d with alternatives:
1) np.interp() (the recommended replacement by scipy, however this does not return an interpolator object)
2) PchipInterpolator (for cases where the interpolator object itself is heavily built into the existing POSYDON functionality, we should discuss if this is reasonable with the developers)

@elizabethteng will look at how things should be updated in the profile interpolation.


mkruckow:
This PR should wait until PR #382 is merged and can serve as a first and simple test case, that changes won't make the unit tests fail.

astroJeff:
I think the pchip interpolator is overkill for these use cases. For low dimensional interpolation, let's just use numpy's implementation of piecewise linear. For this use case, there really isn't a need for or benefit from the more complex hermite polynomial method.

celiotine:
Yeah, I was suggesting the Pchipinterpolator because it is not simple to rewrite some of the existing functions with np.interp(). But if that is what is preferred, it can be done, will just take more time.

astroJeff:
This looks fine now, to me. I agree it is a bit annoying that one cannot create an interpolation object, store it, then evaluate it later. The PChipInterpolator is fine where you have used it.

astroJeff (dismissed):

mkruckow:
After having a deeper look into this, I think we should have our own interpolation object/class. This won't be very complex to do, see the PchipInterpolator2 class, you only need an `init` and a `call` function. Here a suggestion how it could look like:
```
class interp1d:
    def __init__(self, x, y, kind='linear'):
        self.x = x
        self.y = y
        if kind not in ['linear']:
            raise NotImplementedError(f"kind = {kind} is not supported")
        self.kind = kind

    def __call__(self, x_new):
        if self.kind == 'linear':
            return np.interp(x_new, self.x, self.y)
        else:
            raise NotImplementedError(f"kind = {self.kind} is not supported")
```
This can be enhanced by
- allowing other kinds of interpolation
- having checks on the data to be unique
- having checks to ensure interpolation and/or treatments for extrapolation

posydon/binary_evol/DT/step_detached.py
```diff
@@ -27,23 +27,20 @@
...
+                                            check_state_of_star,
+                                            convert_metallicity_to_string,
+                                            set_binary_to_failed)
+from posydon.utils.interpolators import PchipInterpolator2
+from posydon.binary_evol.flow_chart import (STAR_STATES_CC, 
+                                            STAR_STATES_CO, 
+                                            STAR_STATES_H_RICH_EVOLVABLE,
+                                            STAR_STATES_HE_RICH_EVOLVABLE)
 import posydon.utils.constants as const
-from posydon.utils.posydonerror import NumericalError, MatchingError, POSYDONError, FlowError, ClassificationError
```
elizabethteng, line 46: ClassificationError gets called later in this file - did you mean to remove it?

posydon/utils/interpolators.py
elizabethteng, line 1: use in profile_interpolation.py requires the "fill_value" argument, which outside of the given "x" array either extrapolates if passed "extrapolate", or returns specific values if passed (below_x_val,above_x_val) tuple. We'll also have to remove "bounds_error" arg from instances of interp1d in profile_interpolation.

elizabethteng (dismissed):
new custom interp1d function needs argument "fill_value" replicating functionality from original scipy version

posydon/binary_evol/DT/step_detached.py
mkruckow: No, this was by mistake, good catch.

posydon/utils/interpolators.py
mkruckow: OK, I added the possibility to pass a tuple for `fill_value` (similarly `left` or `right` individually) and removed the `bounds_error=False` as this is the standard behavior for `np.interp`.
@elizabethteng , I'd not like to support extrapolation as numpy isn't doing that. Thus, could you please think of doing extrapolation yourself? The problem with extrapolation is, that there isn't a unique way of doing. I guess most common ways are to use the slop of the last two points or the overall slop. In any case connected at the boundary.

posydon/interpolation/interpolation.py
```diff
@@ -13,6 +13,7 @@
 from .data_scaling import DataScaler
 from posydon.grids.psygrid import PSyGrid
 from posydon.grids.MODELS import MODELS
+from posydon.utils.interpolators import interp1d
```
astroJeff, line 15: Can we remove the `from scipy import interpolate` statement above or is it used elsewhere in this file?

posydon/utils/interpolators.py
```diff
@@ -0,0 +1,121 @@
...
+        **kwargs : dict (optional)
+            Dictionary containing extra parameters.
+                - 'left' : float
+                    Y-value to return for evaluations below the data `x`.
+                - 'right' : float
+                    Y-value to return for evaluations above the data `x`.
+                - 'fill_value' : tuple of size 1 or 2
+                    Superseeded by 'left' or 'right'
+                    - size 1 : use y-value for above and below
+                    - size 2 : use first y-value for above and second for below
```
elizabethteng: I think this is backward: should be first value for below and second for above

elizabethteng (dismissed):

elizabethteng (approved):

posydon/utils/interpolators.py
maxbriel, line 1: Could we add more descriptive doc strings for the classes?
Especially the old `PchipInterpolator2` class is unclear in what it does/why it's necessary.
It could be something like this. Feel free to adapt.

"Wrapper around PchirpInterpolator from scipy (see documentation here[link]) 
Stores the interpolator object and allows for output to always be positive"

maxbriel (approved):

posydon/utils/interpolators.py
mkruckow: This can be covered in issue #187 .

PR #435, opened 2024-11-13
Add unit tests: grids

### LINKED ISSUE #518: PSyGrid documentation for join_grid
no example given in docs for how to run join_grids from posydon.grids.psygrid -- I'm getting an error when I try to use it using the syntax suggested by the source code. 

I ran:
`join_grids(('/projects/b1119/POSYDON_GRIDS_v1/CO-HMS/LITE/grid_m_CO_1.20.h5',
                  '/projects/b1119/POSYDON_GRIDS_v1/CO-HMS/LITE/grid_m_CO_1.43.h5'),
                 '/projects/b1119/eteng/new.h5')`

Output: 
Checking /projects/b1119/POSYDON_GRIDS_v1/CO-HMS/LITE/grid_m_CO_1.20.h5
    opening and getting configuration...
    copying configuration...
Checking /projects/b1119/POSYDON_GRIDS_v1/CO-HMS/LITE/grid_m_CO_1.43.h5
    opening and getting configuration...
    checking consistency of configuration and dtypes...
Finalizing grid properties...
Inferring which runs to include, from which grid...
---------------------------------------------------------------------------
AssertionError                            Traceback (most recent call last)
Cell In[24], line 1
----> 1 join_grids(('/projects/b1119/POSYDON_GRIDS_v1/CO-HMS/LITE/grid_m_CO_1.20.h5',
      2                   '/projects/b1119/POSYDON_GRIDS_v1/CO-HMS/LITE/grid_m_CO_1.43.h5'),
      3                  '/projects/b1119/eteng/')

File /projects/b1119/briel/software/POSYDON_versions/POSYDON_016a9b1/posydon/grids/psygrid.py:2209, in join_grids(input_paths, output_path, compression, description, verbose)
   2206 for grid_index, grid in enumerate(grids):
   2207     for run_index, dir_path in enumerate(grid.MESA_dirs):
   2208         # get the parameters part from the dir name
-> 2209         params_from_path = initial_values_from_dirname(dir_path)
   2210         if params_from_path in initial_params:
   2211             n_substitutions += 1

File /projects/b1119/briel/software/POSYDON_versions/POSYDON_016a9b1/posydon/grids/io.py:424, in initial_values_from_dirname(mesa_dir)
    422         variable_names = ["m1", "m2", "initial_period_in_days", "initial_z"]
    423     for variable_name in variable_names:
--> 424         assert variable_name in dirname
    426 values = [dirname.split(variable_name+"_")[1].split("_")[0]
    427           for variable_name in variable_names]
    428 return tuple(values)

AssertionError:

mkruckow:
This PR will get unit tests for the `grids` module.

- [x] `downsampling.py`
- [x] `io.py`
- [x] `MODELS.py`
- [x] `post_processing.py`
- [x] `psygrid.py`
- [x] `scrubbing.py`
- [x] `termination_flags.py`
- [x] update `continuous_integration.yml`
- [x] tested on small pop-syn

Some improvements on the code:

- adding/editing docstrings
- correcting typos (e.g. wrong variable names, wrong strings/values, wrong error/warning types)
- add missing parts (e.g. allow zipped files, make type conversions, missing `raise`, avoid division by zero)
- avoid impossible branches (e.g. directly after you added a key to a dictionary, no need to check for the key)
- add SN data to unit test directory and point the `PATH_TO_POSYDON_DATA` in the tests there.
- apply change on missing and new objects for utils tests as well
- add Runtimeerror in case a system tool, e.g. gzip, fails (although in tests of utils)


posydon/unit_tests/grids/test_downsampling.py
```diff
@@ -0,0 +1,191 @@
...
+from posydon.utils.posydonwarning import InappropriateValueWarning
+
+# define test classes collecting several test functions
+class TestElements:
+    # check for objects, which should be an element of the tested module
+    def test_dir(self):
+        elements = ['Pwarn', 'TrackDownsampler', '__authors__',\
+                    '__builtins__', '__cached__', '__doc__', '__file__',\
+                    '__loader__', '__name__', '__package__', '__spec__', 'np',\
+                    'rescale_from_0_to_1', 'sys']
```
elizabethteng: are the imported modules always supposed to be included in the elements here?

posydon/unit_tests/grids/test_downsampling.py
```diff
@@ -0,0 +1,191 @@
...
+    # test the TrackDownsampler class
+    def test_init(self, independent_sample, dependent_sample,\
+                  norm_dependent_sample, TrackDownsampler):
+        assert isroutine(TrackDownsampler.__init__)
+        # check that the instance is of correct type and all code in the
+        # __init__ got executed: the elements are created and initialized
+        assert TrackDownsampler.verbose == False
+        assert TrackDownsampler.keep == None
+        assert np.array_equal(TrackDownsampler.t, independent_sample)
+        assert np.array_equal(TrackDownsampler.X, norm_dependent_sample)
```
elizabethteng, line 101: shouldn't this be `assert np.array_equal(TrackDownsampler.X, dependent_sample)`?

elizabethteng: in downsampler.py the only assignment of self.X in the __init__ function is `self.X = np.array(dependent)`.

posydon/unit_tests/grids/test_downsampling.py
```diff
@@ -0,0 +1,191 @@
...
+        # examples: dependend is 1D
+        test_TrackDownsampler = totest.TrackDownsampler([0.0, 0.1], [1.0, 2.0])
+        assert np.array_equal(test_TrackDownsampler.t, np.array([0.0, 0.1]))
+        assert np.array_equal(test_TrackDownsampler.X, np.array([[0.0],\
```
elizabethteng, line 105: same concern as above with dependent_sample vs norm_dependent_sample

posydon/grids/downsampling.py
```diff
@@ -122,14 +173,23 @@ def extract_downsample(self, scale_back=True):
...
+              between maximum allowed dm and initial-final change.
+
         """
         t, X = self.t, self.X
         N = len(t)
 
         if max_err is None or N < 3:
-            return np.ones_like(t, dtype=bool)
+            self.keep = np.ones_like(t, dtype=bool)
+            return
```
elizabethteng, line 192: did you mean to leave the "return" empty? should it be "pass"?

posydon/unit_tests/grids/test_downsampling.py
```diff
@@ -0,0 +1,191 @@
...
+            t, X = TrackDownsampler.extract_downsample(scale_back=False)
+            assert np.array_equal(t, independent_sample[k])
+            assert np.array_equal(X, norm_dependent_sample[k, :])
+
+    def test_find_downsample(self, TrackDownsampler, capsys):
+        assert isroutine(TrackDownsampler.find_downsample)
+        # examples: keep all
+        TrackDownsampler.find_downsample()
+        for k in TrackDownsampler.keep:
+            assert k
```
elizabethteng, line 170: I'm a little confused what this line is doing -- does it run an assertion for each of the True/False assignments associated with each row?

elizabethteng:
I've finished reviewing changes to `downsampling.py` and `test_downsampling.py`. They generally look good, I just left a few comments on things I'd like some clarification on

elizabethteng:
just reviewed `MODELS.py` and `test_MODELS.py`. Looks good -- basically just checks that there's at least one core-collapse model and that each model has a mechanism and an engine.

elizabethteng (dismissed):

posydon/unit_tests/grids/test_downsampling.py
mkruckow: Yes, this list contains all the defined names, thus including all the names created by the imports.

posydon/unit_tests/grids/test_downsampling.py
mkruckow: Yes, it gets first assigned the values. But at the end of the init there is a call of `self.rescale()`, this normalizes the data.

posydon/grids/downsampling.py
mkruckow: No, it is a `return`, because the function is complete here. In this case all rows will be kept in the data (either having no downsampling: `max_err is None` or only the edge points: `N<3`), thus the keep array has the same shape as the data and all values are True.

posydon/unit_tests/grids/test_downsampling.py
mkruckow: It runs through the array to check each element to be True.
It should be equivalent to `np.array_equal(TrackDownsampler.keep, np.ones_like(TrackDownsampler.keep, dtype=bool))` just without the need of creating a new array before the comparison.

posydon/unit_tests/README.md
```diff
@@ -38,7 +38,7 @@ Like in code files we add an author list to those files to let others know whom
 
 We always need to import the module we'd like to test. *I recommend to import it with the local name `totest`.*
 
-There might be additional modules/functions needed to do the testing. *I recommend to not import stuff, which is already imported in the module, you like to test, instead use the reference in the tested module, e.g. if the module to test imports X, access it via `totest.X` in the unit test, to ensure using the same code, which is not subject to this test.*
+There might be additional modules/functions needed to do the testing. *I recommend to not import stuff, which is already imported in the module, you like to test, instead use the reference in the tested module, e.g. if the module to test imports X, access it via `totest.X` in the unit test, to ensure using the same code, which is not subject to this test. For convinience you might define some aliases for imported stuff, e.g. `np = totest.np`.*
```
sgossage:
```suggestion
There might be additional modules/functions needed to do the testing. *I recommend to not import stuff, which is already imported in the module, you like to test, instead use the reference in the tested module, e.g. if the module to test imports X, access it via `totest.X` in the unit test, to ensure using the same code, which is not subject to this test. For convenience you might define some aliases for imported stuff, e.g. `np = totest.np`.*
```

posydon/unit_tests/README.md
```diff
@@ -116,12 +116,18 @@ The context object can be used to check for more details of the raised error (th
...
     assert error_info.value.args[0] == "Error message"
 
-It should be noted that the error type is will match subclasses successfully.
+It should be noted that the error type will match subclasses successfully.
 
 #### Catching warnings
 
 Usually, pytest will catch all warnings and print them at the end of all tests. If your test will cause a warning which you don't like to have displayed, you can filter the warnings caught by pytest. To filter all warnings in a function or class you can decorate it with a filter, e.g. `@pytest.mark.filterwarnings("ignore:WARNINGTEXT")`. There are more things you can do on [warnings in pytest](https://docs.pytest.org/en/stable/how-to/capture-warnings.html), but you should use that only were needed. But you should be careful with the pytest warning catching, because it overwrites some parts of the python warnings, which even interferes badly with our POSYDON warnings (especially the filter changes). By using the `pytest.warns` context you can capture and check for warnings the same way as for [errors](#catching-raised-errors).
 
+### Do not test functions called inside or need for a function
```
sgossage:
```suggestion
### Do not test functions called inside or needed for a function
```

posydon/unit_tests/README.md
```diff
@@ -116,12 +116,18 @@ The context object can be used to check for more details of the raised error (th
...
+
+In more complex cases, the function the test acts on will call other functions. Here it can be useful to replace inner functions with `monkeypatch`.
+
+Some functions may require input generated from other functions, here it can be useful to call the other function within a try and return in case it fails to let the test fail on the function, which makes problems not a function, which needs its output. *I recommened therefore to first write a unit test for the functions which do not other functions.*
```
sgossage: Can you clarify/check this text a bit?

Currently ", which makes problems not a function, which needs its output." is unclear to me.

posydon/unit_tests/grids/test_MODELS.py
```diff
@@ -0,0 +1,39 @@
...
+class TestElements:
+    # check for objects, which should be an element of the tested module
+    def test_dir(self):
+        elements = ['MODELS', 'NEUTRINO_MASS_LOSS_UPPER_LIMIT',\
+                    'STATE_NS_STARMASS_UPPER_LIMIT', '__authors__',\
+                    '__builtins__', '__cached__', '__doc__', '__file__',\
+                    '__loader__', '__name__', '__package__', '__spec__']
+        assert dir(totest) == elements, "There might be added or removed "\
+                                        +"objects without an update on the "\
+                                        +"unit test."
```
sgossage:
```suggestion
        missing_in_module = list(set(dir(totest)) - set(elements))
        missing_in_test = list(set(elements) - set(dir(totest)))
        
        assert len(missing_in_test) == 0, \
            "The unit test could not find the following "\
            +f"objects in {totest.__name__}: {missing_in_test}. "\
            +f"\nThese objects may have been removed from {totest.__name__}. "\
            +"Consider updating this unit test by removing these objects from "\
            +"its elements list."
        
        assert len(missing_in_module) == 0, \
           f"The module {totest.__name__} has additional "\
           +"objects not considered by this unit test. They are: "\
           +f"{missing_in_module}.\nThese objects may have been "\
           +f"added to {totest.__name__}. Consider updating this unit "\
           +"test by adding these objects to its elements list."
```

I suggest something like this to help make debugging a little easier, here and similarly for the other unit tests.

sgossage:
Just a note for others working on this review, here's where I am in looking at files:

- [x] grids/downsampling.py
- [x] grids/io.py
- [x] grids/post_processing.py
- [x] grids/psygrid.py
- [x] grids/scrubbing.py
- [x] grids/termination_flags.py
- [x] unit_tests/README.md
- [x] POSYDON_data/* (I haven't checked all these, but looks like just adding data tables and .csv files)
- [x] grids/test_MODELS.py
- [x] grids/test_downsampling.py
- [x] grids/test_io.py
- [x] grids/test_post_processing.py
- [x] grids/test_psygrid.py
- [x] grids/test_scrubbing.py
- [x] grids/test_termination_flags.py
- [x] utils/test_gridutils.py
- [x] utils/test_posydonerror.py
- [x] utils/test_posydonwarning.py
- [x] utils/posydonerror.py

posydon/unit_tests/grids/test_downsampling.py
```diff
@@ -0,0 +1,191 @@
...
+    def test_rescale_from_0_to_1(self):
+        # missing argument
+        with raises(TypeError, match="missing 1 required positional "\
+                                     +"argument: 'array'"):
+            totest.rescale_from_0_to_1()
+        # bad input
+        with raises(ValueError, match="Scaler works for matrices only."):
+            totest.rescale_from_0_to_1(None)
+        # examples
+        test_values = [[0, 1, 3, -1], [0.5, 2, 3, -2], [0.4, 1.5, 3, -1]]
```
sgossage, line 60: I think `test_values` should be a numpy array to successfully pass to `rescale_from_0_to_1()`, rather than a list.

posydon/grids/post_processing.py
```diff
@@ -162,6 +200,8 @@ def post_process_grid(grid, index=None, star_2_CO=True, MODELS=MODELS,
             raise ValueError('Index range should have dim=2!')
         indicies = range(index[0], index[1])
         MESA_dirs = grid.MESA_dirs[index[0]:index[1]]
+    else:
+        raise TypeError('Index should be None, and integer or a list.')
```
sgossage:
```suggestion
        raise TypeError('The argument `index` should be None, an integer, or a list of two integers.')
```

posydon/unit_tests/grids/test_downsampling.py
mkruckow: The function converts it to a numpy array. As the docstring tells you only need an array like object, which includes lists.

posydon/unit_tests/grids/test_io.py
```diff
@@ -0,0 +1,438 @@
...
+        # a temporary grid.csv file for testing
+        path = os.path.join(tmp_path, "grid.csv")
+        with open(path, "w") as test_file:
+            test_file.write("unit,test\n")
+            test_file.write("1.2345678901234567890123456789012345678901234"\
+                            +"5678901234567890123456789,77777777.88888888\n")
+            for i in range(3,6):
+                test_file.write(f"{i}.0,{i}.0\n")
+        # additionally create a zipped file
+        os.system(f"gzip -1 -k {path}")
```
sgossage: This line throws an error for me

```
gzip: invalid option -- 'k'
Try `gzip --help' for more information.
```

Apparently the `-k` option was added in `gzip` v1.6 (we have 1.5 on Quest). An alternative that is supposed to work in any version is e.g., `f"gzip < {path} > {path}.gz"`, but in any case maybe we should consider not using `-k`.

The alternative that I mention above may not preserve permissions, it supposedly doesn't copy the file's metadata ([e.g., see some discussion here](https://unix.stackexchange.com/questions/46786/how-to-tell-gzip-to-keep-original-file)).

posydon/unit_tests/grids/test_io.py
```diff
@@ -0,0 +1,438 @@
...
+               "SINGLE_OUTPUT_FILE is of type: "\
+               + str(type(totest.SINGLE_OUTPUT_FILE))
+
+    def test_instance_EEP_FILE_EXTENSIONS(self):
+        assert isinstance(totest.EEP_FILE_EXTENSIONS, list),\
+               "EEP_FILE_EXTENSIONS is of type: "\
+               + str(type(totest.EEP_FILE_EXTENSIONS))
+
+    def test_instance_RunReader(self):
+        assert isclass(totest.RunReader)
```
sgossage, line 69: Should we have `AssertionError` messages for these tests (and others following)?

posydon/unit_tests/grids/test_io.py
mkruckow: gzip v1.6 was released in June 2013. Hence, do we like to add additional support for older versions? This feels a bit in contradiction to not support python versions below 3.11 (released October 2022).

posydon/unit_tests/grids/test_io.py
mkruckow: What additional information would we need in the message?

elizabethteng:
I have reviewed scrubbing.py and termination_flags.py and their associated tests. Here's what I've covered so far:

- [x]  downsampling.py and test
- [ ]  io.py and test
- [x]  MODELS.py and test
- [ ]  post_processing.py and test
- [ ]  psygrid.py and test
- [x]  scrubbing.py and test
- [x]  termination_flags.py and test

I think it would be good for someone who has more familiarity with MESA to review the termination flags test as well. The code looks good but I couldn't evaluate how comprehensive the example cases were.

posydon/unit_tests/grids/test_io.py
sgossage: For example here, maybe something simple like `f"RunReader is not a class, it has type {type(RunReader)}"` to help debug.

posydon/unit_tests/grids/test_io.py
sgossage: I'll look into whether we can force another `gzip` version or else upgrade our `gzip` to a higher version.

mkruckow: I have added a Runtimeerror in case a call of a system tool fails.

posydon/unit_tests/grids/test_io.py
mkruckow: The current pytest output on failing is:

> E       assert False
> E        +  where False = isclass(<function RunReader at 0x70a0c6311580>)
> E        +    where <function RunReader at 0x70a0c6311580> = totest.RunReader
> 
> posydon/unit_tests/grids/test_io.py:69: AssertionError

So, I'd say the information, that `RunReader` isn't a class is given. Do we need the information about the type? Hence, is it worth to add the type information when we are checking whether it is a class. (This is different for the checks on isinstance, which checks for the type.)

sgossage: I think something like that is good practice in general, I'm not sure if `type()` necessarily is the most helpful here and that one might be informative enough as is. 

However, there are a few places to check over and where additional `AssertionError` messages of this sort might be useful. If the test somehow fails, people should be able to tell from the error why, without having to write additional code to determine the issue.

For example, below in 

`90     assert isinstance(totest.POSYDON_FORMAT_OPTIONS[o], list)`

to output what `totest.POSYDON_FORMAT_OPTIONS[o]` is, if not a list so it can be debugged faster. We may also want to do this to be consistent with other cases where we output such information. 

Or here 

`156    assert "CONTENTS OF METADATA FILE:" in captured_output.out`

we might want to output what actually is in `captured_output.out` so someone diagnosing can readily compare without having to write code to check that. Or things like this

`93    assert totest.BINARY_OUTPUT_FILE == "out.txt"`

We could output something like

```
f"This test expected {totest.__name__}.BINARY_OUTPUT_FILE to have the format `out.txt`. It has format {totest.BINARY_OUTPUT_FILE}."
```

posydon/unit_tests/grids/test_io.py
sgossage: This will be resolved further by the upcoming Quest maintenance, where we'll upgrade to RHEL 8.

The new installation has `gzip` v 1.9

posydon/unit_tests/grids/test_post_processing.py
```diff
@@ -0,0 +1,659 @@
...
+        assert len(output) == 1+1
+        fmt = "{:1.2f} {:13.2f} {:12.2f} {:20.2f} {:20.2f} {:20.2f} {:20.2f} "\
+              +"{:20.2f} {:20.2f}"
+        for v in ["TESTMODEL", star.state, star.SN_type, fmt.format(star.f_fb,\
+                   star.mass, star.spin, star.m_disk_accreted,\
+                   star.m_disk_radiated, star.M4, star.mu4, star.h1_mass_ej,\
+                   np.nan)]:
+            assert v in output[0]
+
+    def test_post_process_grid(self, grid_path, capsys, monkeypatch):
```
sgossage: On my end, this yields a segmentation fault if `$PATH_TO_POSYDON_DATA` is an empty directory or doesn't contain the SN data. In this case, the code tries to download data, and I guess something about that process does not work well with `pytest`. 

This is triggered by the call `SN = StepSN(**MODEL, allow_spin_None=True)` in line `347` of `post_processing.py`, which then goes on to look for the SN datasets and downloads POSYDON data if it can't find them.

Maybe we want to run a check and raise an error if these directories do not exist before executing this test? Something like

```
from posydon.binary_evol.SN.step_SN import path_to_Sukhbold_datasets as sukhbold_path
from posydon.binary_evol.SN.step_SN import path_to_Patton_datasets as patton_path
from posydon.binary_evol.SN.step_SN import path_to_Couch_datasets as couch_path

...
    def test_post_process_grid(self, grid_path, capsys, monkeypatch):
        ....
        assert os.path.exists(sukhbold_path) and os.path.exists(patton_path) and os.path.exists(couch_path),\
            "This test requires POSYDON data. Please ensure that your `$PATH_TO_POSYDON_DATA`"\
            + "environment variable is properly set and leads to an existing POSYDON data set."
```

Or maybe there's another remedy to get the data download to work?

mkruckow: I have already added this data to this PR as a workaround, as you pointed out:

> POSYDON_data/* (I haven't checked all these, but looks like just adding data tables and .csv files)

Usually one should either have a full installed POSYDON version (the [instructions](https://posydon.org/POSYDON/getting-started/installation-guide.html) include to download the POSYDON data) or one should use a special PATH_TO_POSYDON_DATA as given in [continuous_integration.yml](https://github.com/POSYDON-code/POSYDON/pull/435/files#diff-11af9d5a6b06b599f7d4a3840b527743fb1f124615ef015b055bd4d036da2a5d):

`export PATH_TO_POSYDON_DATA=./posydon/unit_tests/_data/`

To make it easier for people doing none of this, I added a fixture to create temporary links to the data in the unit tests.

And yes, the better solution would be to have just this helper data being available to download. As long as this is not available, we need this workaround.

sgossage: Sounds good, thanks.

How about a warning message like this to tell users/testers to check their installation?

In the `link_SN_data()` fixture, something like:
```
        # create links if needed
        ....
            if not os.path.exists(engine_path):

                warnings.warn(f"$PATH_TO_POSYDON_DATA: {PATH_TO_POSYDON_DATA} is missing "\
                             +f"{SN_engine} data. Please check your POSYDON installation. "\
                             +"\n(Check step 5 of the installation instructions "\
                             +"(https://posydon.org/POSYDON/getting-started/installation-guide.html).", 
                             UserWarning, stacklevel=1)

              ...
```

which requires `import warnings` as well. And/or we could instruct to use the special path from [continuous_integration.yml](https://github.com/POSYDON-code/POSYDON/pull/435/files#diff-11af9d5a6b06b599f7d4a3840b527743fb1f124615ef015b055bd4d036da2a5d) for unit testing

posydon/unit_tests/grids/test_io.py
```diff
@@ -0,0 +1,446 @@
...
+            test_file.write("1.2345678901234567890123456789012345678901234"\
+                            +"5678901234567890123456789,77777777.88888888\n")
+            for i in range(3,6):
+                test_file.write(f"{i}.0,{i}.0\n")
+        # additionally create a zipped file
+        try:
+            os.system(f"gzip -1 -k {path}")
+        except:
+            raise RuntimeError("Please check that you have `gzip` installed "\
+                               +"and up to date.")
```
sgossage, line 119:
```suggestion
            raise RuntimeError("Please check that you have `gzip` installed "\
                               +"with version >= 1.6.")
```

and elsewhere where this error message occurs.

mkruckow: Here we capture all issues with gzip not only stuff solved with version 1.6 or higher. Thus I'd keep the general message as we don't know, why it failed. Yes, in your test case it was about having version >= 1.6, but this might not be the reason for others.
By the way, I added similar tries for other external tools used. Even we didn't had issues with them so far.

posydon/unit_tests/grids/test_post_processing.py
mkruckow: That's something for the test suite/validation tests to check the correct and complete installation incl. downloading the data. Thus I'd not put it here in the code testing/unit tests.

sgossage (approved):

.github/workflows/continuous_integration.yml
```diff
@@ -34,6 +34,6 @@ jobs:
         pip install pytest
         pip install pytest-cov
         export PATH_TO_POSYDON=./
-        export PATH_TO_POSYDON_DATA=./
+        export PATH_TO_POSYDON_DATA=./posydon/unit_tests/_data/
         export MESA_DIR=./
-        python -m pytest posydon/unit_tests/ --cov=posydon.utils --cov=posydon.config --cov-branch --cov-report term-missing --cov-fail-under=100
+        python -m pytest posydon/unit_tests/ --cov=posydon.config --cov=posydon.utils --cov=posydon.grids --cov-branch --cov-report term-missing --cov-fail-under=100
```
maxbriel, line 39: Would it be possible to start adding the coverage to the setup.cfg or pyproject.toml? Or use a .coveragerc file?
See:
https://pytest-cov.readthedocs.io/en/latest/config.html

maxbriel (approved):
The improvements to docstrings and code looks good. 

One small comment about maybe moving our coverage testing to setup.cfg, pyproject.toml, or a .coveragerc file. But not crucial for this PR.

I noticed when running the tests, those for `utils/configfile.py` have some missing lines, when run locally, but are covered on the Github runners.  Curious why that happens.

```
posydon/utils/configfile.py            116     38     40      0    64%   242-280, 287-312  
```

.github/workflows/continuous_integration.yml
maxbriel: In the `max_sampling`branch I've started moving these to a `.coveragerc` file :)

mkruckow: That's something to discuss in a separate PR. We should think whether we want to have the unit tests to be part of the installation (include into `setup.cfg`, I'd suggest to not add it to `pyproject.toml` separately, if needed putting there a reference to use `setup.cfg`) or whether that's only for development. Surely the other test class, the validation tests/test suite, should become part of the installation.

PR #436, opened 2024-11-14
Protect run_psycris_sequence, from misc imports

### LINKED ISSUE #134: TypeError in active_learning/psy_cris/run_params/run_psycris_sequence.py
When I do `import posydon.active_learning.psy_cris.run_params.run_psycris_sequence` I get the following error:
```
File ~/POSYDON/posydon/active_learning/psy_cris/run_params/run_psycris_sequence.py:45
     43 inifile_path = args.inifile_path
     44 save_data_dir = args.save_data_dir
---> 45 if not os.path.isfile(inifile_path):
     46     raise ValueError("Inifile not found. Please check that the path is correct.\nGiven: {}".format(inifile_path))
     47 if not os.path.isdir(save_data_dir):

File <frozen genericpath>:30, in isfile(path)

TypeError: stat: path should be string, bytes, os.PathLike or integer, not NoneType
```
It looks like the variables do not get initialized well and therefore create this TypeError.

ka-rocha:
Fixing #134 by adding a dunder main for command line use only.

### DIFF
--- a/posydon/active_learning/psy_cris/run_params/run_psycris_sequence.py
+++ b/posydon/active_learning/psy_cris/run_params/run_psycris_sequence.py
@@ -12,115 +12,119 @@ from posydon.active_learning.psy_cris.utils import parse_inifile
 from posydon.active_learning.psy_cris.utils import do_dynamic_sampling
 from posydon.active_learning.psy_cris.utils import calc_performance
 
-parser = argparse.ArgumentParser()
-parser.add_argument( "-ini", dest='inifile_path', type=str, help="Path to psy-cris inifile.")
-parser.add_argument( "-dir", dest='save_data_dir', type=str, help="Path to directory to save data.")
+def main():
+    parser = argparse.ArgumentParser()
+    parser.add_argument( "-ini", dest='inifile_path', type=str, help="Path to psy-cris inifile.")
+    parser.add_argument( "-dir", dest='save_data_dir', type=str, help="Path to directory to save data.")
 
-parser.add_argument( "-id", dest='run_id', type=str, help="Integer or str used to identify runs.",
-                    default = np.random.randint(low=10000, high=99999) )
-parser.add_argument( "-n", dest='n_sequences', type=int, help="N sequences to run.",
-                    default=1)
-parser.add_argument( "-num_start", dest='n_starting_points', type=int, help="N points to start with in even grid.",
-                    default=3**3)
-parser.add_argument( "-num_fin", dest='n_final_points', type=int, help="Total num points to converge to.",
-                    default=6**3)
-parser.add_argument( "-ppi", dest='points_per_iter', type=int, help="Num points between each iteration.",
-                    default=70)
-parser.add_argument( "-cls", dest='performance_cls_name', type=str, help="Classification algorithm used in performance calc.",
-                    default="linear")
-parser.add_argument( "-regr", dest='performance_regr_name', type=str, help="Regression algorithm used in performance calc.",
-                    default="rbf")
-parser.add_argument( "-res", dest='performance_res', type=int, help="Resolution to compute performance.",
-                    default=40)
-parser.add_argument( "-length_scale_mult", dest='length_scale_multiplier', type=float, help="Small class proposal length scale mult of normal.",
-                    default=0.3333)
-parser.add_argument( "-percent_increase", dest='percent_increase', type=float, help="Icrease points per iter by a percent of the training set.",
-                    default=-1)
-
-
-args = parser.parse_args()
-
-inifile_path = args.inifile_path
-save_data_dir = args.save_data_dir
-if not os.path.isfile(inifile_path):
-    raise ValueError("Inifile not found. Please check that the path is correct.\nGiven: {}".format(inifile_path))
-if not os.path.isdir(save_data_dir):
-    raise ValueError("Directory not found. Please check that save_data_dir exists.\nGiven: {}".format(save_data_dir))
-
-run_ID = args.run_id
-n_psycris_sequences = args.n_sequences
-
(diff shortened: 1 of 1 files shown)


ka-rocha:
Testing the code:
```python 
from  posydon.active_learning.psy_cris.run_params import run_psycris_sequence
```
now runs without error.

However, running the actual psycris sequence breaks half way through because of the latest syntax changes from e.g. `pandas`. I think we may move this to another issue if it is important to fix the import issue asap.

ka-rocha:
Also, since the diff is not clear: what I've done is wrapped the entire code into a main block so it runs only when called from the command line (which was the original intention of this py file for testing AL with synthetic data). I also added some type checks for the original error but this is not the main issue.

```python
def main():
    ...
    # do psycris sequence
    ...

if __name__ == '__main__':
    main()
```

mkruckow:
Thanks @ka-rocha ,

1. Wouldn't it be simpler to replace the function definition with the check for main directly?
2. Now it looks like this code it aimed to be run on command line directly. Code to be run from command line should go into `bin` to be although installed as a command on the environment.

ka-rocha:
@mkruckow Here are my initial thoughts:
 
> 1. Wouldn't it be simpler to replace the function definition with the check for main directly?

I think either way is fine but in my opinion the function definition is more pythonic. (e.g. https://docs.python.org/3/library/__main__.html#idiomatic-usage)

> 2. Now it looks like this code it aimed to be run on command line directly. Code to be run from command line should go into `bin` to be although installed as a command on the environment.

Yes the code is meant to run on command line but only in the case you wish to reproduce the analysis for Figures 2-5 from https://arxiv.org/abs/2203.16683. If the rule is to move everything into `bin` because it is executable then I suppose we could do this, although in my opinion this code is used very differently from the other scripts currently there so I would propose we keep it in the current location. 

Let me know what you think!

mkruckow:
> > 1. Wouldn't it be simpler to replace the function definition with the check for main directly?

> I think either way is fine but in my opinion the function definition is more pythonic. (e.g. https://docs.python.org/3/library/__main__.html#idiomatic-usage)

It's up to us, which way we prefer (I'm fine with both). I can only tell, that in all other parts of the code were we check for being in `__main__` we use the other way. I'm only against using different ways. Therefore from my side, please either adopt the way use in other code parts or change the other code parts depending which way we'd like to go. Maybe that's another thing to discuss and decide on in the next developer meeting.

> > 2. Now it looks like this code it aimed to be run on command line directly. Code to be run from command line should go into `bin` to be although installed as a command on the environment.

> Yes the code is meant to run on command line but only in the case you wish to reproduce the analysis for Figures 2-5 from https://arxiv.org/abs/2203.16683. If the rule is to move everything into `bin` because it is executable then I suppose we could do this, although in my opinion this code is used very differently from the other scripts currently there so I would propose we keep it in the current location.

OK, now I get your aim with the code. In the case of code to reproduce a figure, I'd suggest to make this not a command line executable, but a function called `create_figure_XY` with the arguments you'd like to parse through, which you don't run in command line, but in any python environment via
```
from posydon.active_learning.psy_cris.run_params.run_psycris_sequence import create_figure_XY
create_figure_XY()
```

ka-rocha:
> > > 1. Wouldn't it be simpler to replace the function definition with the check for main directly?
> 
> > I think either way is fine but in my opinion the function definition is more pythonic. (e.g. https://docs.python.org/3/library/__main__.html#idiomatic-usage)
> 
> It's up to us, which way we prefer (I'm fine with both). I can only tell, that in all other parts of the code were we check for being in `__main__` we use the other way. I'm only against using different ways. Therefore from my side, please either adopt the way use in other code parts or change the other code parts depending which way we'd like to go. Maybe that's another thing to discuss and decide on in the next developer meeting.

Yes I can appreciate consistency although your second suggestion is actually why the function definition is preferred over checking for dunder main in a random location in the file -- so that imports of desired methods are possible. IMO changing all dunder main usage across the code base is out of scope for a bug fix PR as this is directly addressing #134. If it is decided the other dunder main usage is preferred then I can update this to be consistent!

> > > 2. Now it looks like this code it aimed to be run on command line directly. Code to be run from command line should go into `bin` to be although installed as a command on the environment.
> 
> > Yes the code is meant to run on command line but only in the case you wish to reproduce the analysis for Figures 2-5 from https://arxiv.org/abs/2203.16683. If the rule is to move everything into `bin` because it is executable then I suppose we could do this, although in my opinion this code is used very differently from the other scripts currently there so I would propose we keep it in the current location.
> 
> OK, now I get your aim with the code. In the case of code to reproduce a figure, I'd suggest to make this not a command line executable, but a function called `create_figure_XY` with the arguments you'd like to parse through, which you don't run in command line, but in any python environment via
> 
> ```
> from posydon.active_learning.psy_cris.run_params.run_psycris_sequence import create_figure_XY
> create_figure_XY()
> ```

I could potentially do this, although the code we are looking at is actually the data generation step while the figures I reference require ML hold out cross validation. The whole pipeline for creating the figures is not in the source code currently and would require more code infrastructure to be added. In my opinion adding functionality to call code in ways it was not designed to be called is out of scope for a bug fix PR.

Please advise if these additions are considered required for addressing #134.

mkruckow:
Surely, you don't need to do all in one PR, just open new issues for remaining stuff, hence here my suggestion:

1. Change the code to use the check for main as in other code parts and open an issue to make a consistent change to the function call form. (Please don't miss to add labels to the issue and assign developers)
2. Leave like it is and open an issue, which is either assigned to you (and take care of it separately) or contains all the information another assigned person needs to know what code is missing.

mkruckow (approved):

astroJeff:
Looks good to me

astroJeff (approved):

PR #438, opened 2024-11-19
Change bbox to `standard`

maxbriel:
When saving figures, matplotlib was set to use `bbox : tight` in the `posydon.mplstyle`.
This changes the size of the resulting figure, despite setting your own or using the default POSYDON dimensions.

Setting it back to `standard` should respect the set figure size. 

The best way to keep the same figure size would be to use `constrained_layout`, but this is incompatible with custom white space between figures.

### DIFF
--- a/posydon/visualization/posydon.mplstyle
+++ b/posydon/visualization/posydon.mplstyle
@@ -17,7 +17,7 @@ figure.dpi 		: 300
 xtick.minor.visible     : True
 ytick.minor.visible     : True
 
-savefig.bbox		:  standard  # tight
+savefig.bbox		: standard  # tight
 savefig.pad_inches	: 0.1 # Padding when bbox is tight.
 savefig.dpi		: 300
 savefig.format		: pdf
--- a/posydon/visualization/posydon.mplstyle
+++ b/posydon/visualization/posydon.mplstyle
@@ -17,7 +17,7 @@ figure.dpi 		: 300
 xtick.minor.visible     : True
 ytick.minor.visible     : True
 
-savefig.bbox		: tight # standard
+savefig.bbox		:  standard  # tight
 savefig.pad_inches	: 0.1 # Padding when bbox is tight.
 savefig.dpi		: 300
 savefig.format		: pdf


mkruckow:
1. Have you checked, whether this causes changes on the figure for the v2 paper? I guess, when accepting this PR we should redo all the figures in the v2 paper.

2. Was there a reason, why Simone set the other option before?

maxbriel:
1. This causes changes for v2 plots. Most plots in the paper have slightly different sizes and therefore also slightly different font sizes etc. 

**full page width examples:**

- `HMS-HMS_MESA_grid_1e+00_Zsun_TF12_8_mass_ratio_panels.pdf` is 18.26 × 17.85 cm
- `CO-HeMS_TF12_9panel_BH.pdf` is 18.89 × 17.85 cm
- `violinplots.png is 2323 × 2336pixels (100 pixels/inch) -> 23.23x23.36inches -> 59x59.33cm
 
**column width examples:**
- ZAMS-max_radius-metal_lin_cut.pdf is 8.49 × 6.4 cm
- POSYDON_wind.pdf is 9.35 × 12.2 cm

The second number (height) is less important, but the first number is the width.

2. I can only guess, but in general it makes everything fit inside the figure nicely without too much whitespace.

sgossage:
Looks good, I don't think there's an easy way to get around this otherwise. I guess we may have to manually adjust figures to remove white space with this change.

As Max says, `bbox : 'tight'` will try to remove the white space around a figure in an automated way, but this then produces irregularly sized figures that don't respect the specified figure size.

mkruckow:
At least for me, the aim of the preset style is to avoid/minimize individual adjustments. Thus, if we need more adjustments with the change, that count as a big con for me.
On the other hand, I was thinking, whether this may causes use to have different styles. E.g. the automatically generated plots in the pipeline have to work without any manual adjustment.
Finally, if we use don't use tight, why do we keep `savefig.pad_inches`? According to the documentation, this only gets applied for `bbox='tight'`.

Maybe it would be best if you can provide some example plots to show in the dev meeting to discuss there, what we need/want.

maxbriel:
The fundamental issue, currently, is that you get different plot sizes. This requires more manual tweaking to get the right sizes for papers: fontsize, ticksize etc.
At the moment, the plots in the pipeline have different output sizes.

This is a larger problem to me than having additional whitespace. Why do you want to remove the additional whitespace?
If you really want to restrict the whitespace, we can change the subplot parameters of where the plots start.
This works even for a single parameter.

```
# The figure subplot parameters.  All dimensions are a fraction of the figure width and height.
figure.subplot.left:   0.125  # the left side of the subplots of the figure
figure.subplot.right:  0.9    # the right side of the subplots of the figure
figure.subplot.bottom: 0.11   # the bottom of the subplots of the figure
figure.subplot.top:    0.88   # the top of the subplots of the figure
```

We can remove the `pad_inches`.

mkruckow:
I think, the shrinkage is even not the main issue, why we have different sizes:
a full page width figure should be 7.1inch=18.0cm
a column width figure should be 3.38inch=8.59cm
Most of the examples for the v2 paper aren't smaller, but larger.

elizabethteng:
I think it's more important that the font sizes/line widths are consistent between plots and IMO having extra whitespace is a worthwhile price to pay to ensure that. So, I'm in favor of making this change.

astroJeff:
While I appreciate the attention to detail (thank you @maxbriel for bringing this up), I don't think this is where we should be spending most of our time. Let's make a decision and move on to more important things.

astroJeff:
Potentially causes more problems than it fixes. Closing.

PR #440, opened 2024-11-20
PSygrid force int input

maxbriel:
force the user to input an integer. This includes numpy versions of int.

If this is not forces, the `initial_values` and `final_values` can be loaded with a 1 value array/list, but the `history1`, etc are not loaded with that input. As a result, `None` is returned for those and it looks like the history is empty.

### DIFF
--- a/posydon/grids/psygrid.py
+++ b/posydon/grids/psygrid.py
@@ -1541,7 +1541,7 @@ class PSyGrid:
         return ret
 
     def __getitem__(self, index):
-        """Return a PSyRunView instance for the run with index `idx`."""
+        """Return a PSyRunView instance for the run with index `index`."""
         if not np.issubdtype(type(index), int):
             raise TypeError("Index {} is not of type int".format(index))
         if index not in self:
--- a/posydon/grids/psygrid.py
+++ b/posydon/grids/psygrid.py
@@ -1542,10 +1542,10 @@ class PSyGrid:
 
     def __getitem__(self, idx):
         """Return a PSyRunView instance for the run with index `idx`."""
-        if idx not in self:
-            raise KeyError("Index {} out of bounds.".format(idx))
         if not np.issubdtype(type(idx), int):
             raise TypeError("Index {} is not of type int".format(idx))
+        if idx not in self:
+            raise KeyError("Index {} out of bounds.".format(idx))
         return PSyRunView(self, idx)
 
     def get_pandas_initial_final(self):
@@ -1566,6 +1566,8 @@ class PSyGrid:
 
     def __contains__(self, index):
         """Return True if run with index `index` is in the grid."""
+        if not np.issubdtype(type(idx), int):
+            raise TypeError("Index {} is not of type int".format(idx))
         return 0 <= index < self.n_runs
 
     def __iter__(self):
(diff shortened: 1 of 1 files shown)


posydon/grids/psygrid.py
```diff
@@ -1544,6 +1544,8 @@ def __getitem__(self, idx):
         """Return a PSyRunView instance for the run with index `idx`."""
         if idx not in self:
             raise KeyError("Index {} out of bounds.".format(idx))
+        if not np.issubdtype(type(idx), int):
+            raise TypeError("Index {} is not of type int".format(idx))
```
mkruckow: Should we move this to the `__contains__` function or add a similar statement there (in the second case, this needs to be moved two lines up)?
It feels wrong to me to get `a in grid == True`, but `grid[a]` raises an error.

maxbriel: I see.
To verify: do you mean that the comparison can be made with a single-entry numpy array for `a in grid`, while we now force that to not work for `__getitem__`?

The same behaviour is preferable.

mkruckow: Yes, I think we should have the same restrictions to both functions.

By the way, I think we raise the wrong error type for `idx not in self`, it should be an IndexError, because we talk about an indexed and not about a mapping object.

posydon/grids/psygrid.py
```diff
@@ -1564,6 +1566,8 @@ def __len__(self):
 
     def __contains__(self, index):
         """Return True if run with index `index` is in the grid."""
+        if not np.issubdtype(type(idx), int):
```
sgossage:
```suggestion
        if not np.issubdtype(type(index), int):
```

posydon/grids/psygrid.py
```diff
@@ -1564,6 +1566,8 @@ def __len__(self):
...
+            raise TypeError("Index {} is not of type int".format(idx))
```
sgossage:
```suggestion
            raise TypeError("Index {} is not of type int".format(index))
```

posydon/grids/psygrid.py
sgossage: I suggest utilizing `index` rather than `idx` in this function to maintain familiarity between similar functionalities in the code (like `__contains__` which currently uses the former).

sgossage (dismissed):

mkruckow (dismissed):
OK for now. There is more to adjust to make clear, whether this object is aimed to be a sequence (indexed) or a mapping.

posydon/grids/psygrid.py
```diff
@@ -1540,11 +1540,13 @@ def one_if_ok(data):
 
         return ret
 
-    def __getitem__(self, idx):
+    def __getitem__(self, index):
         """Return a PSyRunView instance for the run with index `idx`."""
```
elizabethteng: plz update docstring to match change to arg name

elizabethteng (changes requested):

elizabethteng:
@maxbriel is there an ideological reason for replacing "idx" with "index"? I ask because there are many other instances where "idx" appears in this file. In any case, this should be good to go.

elizabethteng (approved):

mkruckow (approved):
For now, it's OK, but we should be aware, that there are more updates needed to ensure that the object behaves well as a sequence (indexed object).

maxbriel:
> @maxbriel is there an ideological reason for replacing "idx" with "index"? I ask because there are many other instances where "idx" appears in this file. In any case, this should be good to go.

@elizabethteng There was this conversation about it: https://github.com/POSYDON-code/POSYDON/pull/440#discussion_r1852812272 
No specific reason from my end.

sgossage (approved):

PR #460, opened 2024-12-01
Fix log vs log10 typo

maxbriel:
In the MD/MF metallicity fraction calculation, there was a typo for the outer bin in transforming the absolute metallicity into a log10 value. It was log, but should be log10!


mkruckow (approved):

elizabethteng (approved):
thanks for catching this!

Issue #458, opened 2024-11-28
Undefined binaries in flow: ('NS', 'H-rich_Core_H_burning', 'disrupted', 'CC2')


mkruckow:
When doing some tests for v2.0pre1, I have see several cases of

- ValueError: Undefined next step given stars/binary states ('NS', 'H-rich_Core_H_burning', 'disrupted', 'CC2').
- ValueError: Undefined next step given stars/binary states ('NS', 'H-rich_Core_H_burning', 'detached', 'CC2').

They happend both in IF and NN runs. It is questionable how a 'H-rich_Core_H_burning' is send to CC.

Here the full error of one example case:
```
Traceback (most recent call last):
  File "/home/users/k/kruckow/POSYDON/posydon/popsyn/binarypopulation.py", line 325, in _safe_evolve
    binary.evolve()
  File "/home/users/k/kruckow/POSYDON/posydon/binary_evol/binarystar.py", line 219, in evolve
    self.run_step()
  File "/home/users/k/kruckow/POSYDON/posydon/binary_evol/binarystar.py", line 238, in run_step
    raise ValueError("Undefined next step given stars/binary states {}.".format(total_state))
ValueError: Undefined next step given stars/binary states ('NS', 'H-rich_Core_H_burning', 'disrupted', 'CC2').

Failed Binary Initial Conditions:
S1 mass: 7.302102710931483
S2 mass: 0.7879377046773028
S1 state: H-rich_Core_H_burning
S2 state: H-rich_Core_H_burning
orbital period: 1719.5155572498547
eccentricity: 0.0
binary state: detached
binary event: detached
S1 natal kick array: [500.3825406264961 2.9365934718580298 1.2255496882454837 5.953462883339887]
S2 natal kick array: [None None None None]
```

maxbriel:
@mkruckow Is this related to our v2 issue with the single stars and the EEPs?  ([PR#243](https://github.com/POSYDON-code/POSYDON/issues/243))
There we had to evolve the v2 single stars past the age of the Universe to get all their EEPs to show up.

Till what point do we evolve the v1 version of single stars?

mkruckow:
Good point, this might in fact be the issue which we solved for v2 by evolving the stars further as it was done before (incl. v1).
As we can't solve it on the grid data level for v1, shell we try to catch the error and/or mention it in the v1 documentation?

celiotine:
Addressed in PR #468

PR #468, opened 2024-12-11
error check for undefined states in BinaryStar

celiotine:
This PR addresses Issue #458. It adds an error check for any binaries that enter an undefined state, and also creates a list of know undefined states in the flow that exist and don't need to be fixed, like those discussed in the issue.


celiotine:
I had to make additional changes to the POSYDON error class. The dependency on the BinaryStar and SingleStar classes was creating a circular import when trying to use POSYDON errors in those classes, so I removed the functionality related to these objects (no one has used it yet in the code anyway). I moved the "initial_condition_message" function to the BinaryStar class.

elizabethteng (approved):

astroJeff (approved):

PR #475, opened 2025-01-07
Grid slice plotting system selection fix

### LINKED ISSUE #474: Population.plot_popsyn_over_grid_slice() selecting wrong systems
When using the latest version of POSYDON `plot_popsyn_over_grid_slice()` selects the "wrong" systems. 

This is due to the following lines:
https://github.com/POSYDON-code/POSYDON/blob/13ec9547bd13f75d8e723287b64edf0f07614e4b/posydon/visualization/plot_pop.py#L246-L247

The system_index in the Dataframe is not required to be reset when a Transientpopulation is created, as such the `np.where` returns wrong results.

This can be solved by using the indices instead of `np.where`.

maxbriel:
Fixes Issue #474 and adds a docstring to `plot_popsyn_over_grid_slice`

### DIFF
--- a/posydon/visualization/plot_pop.py
+++ b/posydon/visualization/plot_pop.py
@@ -220,7 +220,7 @@ def plot_popsyn_over_grid_slice(pop,
     plot_dir : str
         Directory where to save the plots.
     prop : str
-        Property to plot on the grid. If None, no property is plotted.
+        Property to plot on the grid. If None, no property colour is plotted.
     prop_range : list of floats
         Range of the property to plot.
     log_prop : bool
@@ -236,7 +236,7 @@ def plot_popsyn_over_grid_slice(pop,
     close_fig : bool
         If True, the plot is closed.
     plot_extension : str
-        Extension of the plot.
+        File extension of the plot.
     verbose : bool
         If True, print information about the slices being plotted.
     '''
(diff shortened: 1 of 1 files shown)


posydon/visualization/plot_pop.py
```diff
@@ -187,27 +187,73 @@ def plot_hist_properties(df, ax=None, df_intrinsic=None, df_observable=None,
...
+    s : float
+        Size of the points.
+    show_fig : bool
+        If True, the plot is shown.
+    save_fig : bool
+        If True, the plot is saved.
+    close_fig : bool
+        If True, the plot is closed.
+    plot_extension : str
+        Extension of the plot.
```
mkruckow: Very minor: I'd be a bit more specific and call it file extension. As matplotlib uses it to determine the image file type.

posydon/visualization/plot_pop.py
```diff
@@ -187,27 +187,73 @@ def plot_hist_properties(df, ax=None, df_intrinsic=None, df_observable=None,
...
+    met_Zsun : float
+        Metallicity of the grid.
+    slices : list of floats
+        List of slices to plot. If None, all slices are plotted.
+    channel : str
+        Formation channel to plot. If None, all channels are plotted.    
+    plot_dir : str
+        Directory where to save the plots.
+    prop : str
+        Property to plot on the grid. If None, no property is plotted.
```
mkruckow: > If None, no property is plotted.
This might be misleading. In case of None, there is no property color coded, you'll still get a valid plot with the default properties of x and y.

mkruckow (approved):

elizabethteng (approved):

Issue #383, opened 2024-09-12
plot_pop_over_grid() change selection


maxbriel:
Currently, it's hardcoded what "event" from the binary is overplotted based on the grid choice selection.
That is HMS-HMS plots ZAMS, while CO-HMS plots RLO2 and CO-HeMS plots CE2.
These are very specific to BBH mergers.

This function should be altered to take the `step_name` which is closer to the grid that is being plotted.
This can either be a user input or a fixed value per grid.

PR #476, opened 2025-01-07
Split popsyn plotting into separate components for future modularity

maxbriel:
This PR does the following:

- refactor of the existing code into separate function to make the code clearer and allow for easier changes in the future.
- Fix an additional bug in overplotting, where the wrong mass was used in the CO grids (M1 instead of M2 of the population).
- Remove unused functions. I've left the plotting function of Perley+16 GRB rates since that could be useful.
     - `plot_merger_rate_density()`
     - `plot_grb_rate_density()`

### DIFF
--- a/posydon/visualization/plot_pop.py
+++ b/posydon/visualization/plot_pop.py
@@ -420,7 +420,27 @@ def plot_population_data(data,
                          log_prop=False,
                          alpha=0.3,
                          s=5):
-    '''Plot the population data based on the grid slice'''
+    '''Plot the population data based on the grid slice parameters
+    
+    Parameters
+    ----------
+    data : DataFrame
+        Population data
+    slice_3D_var_str : str
+        String of the 3D variable to slice the grid
+    slice_3D_var_range : list
+        Range of the 3D variable to slice the grid
+    prop : str (optional)
+        Property to plot on top of the grid slice
+    prop_range : list (optional)
+        Range of the property to plot
+    log_prop : bool (optional)
+        Logarithmic scale for the property
+    alpha : float (optional)
+        Transparency of the data points
+    s : float (optional)
+        Size of the data points
+    '''
 
     # get only slice data
     if slice_3D_var_str == 'mass_ratio':
--- a/posydon/visualization/plot_pop.py
+++ b/posydon/visualization/plot_pop.py
@@ -186,9 +186,9 @@ def plot_hist_properties(df, ax=None, df_intrinsic=None, df_observable=None,
         plt.show()
 
 
-
-
-def plot_popsyn_over_grid_slice(pop, grid_type, met_Zsun, slices=None, channel=None,
+def plot_popsyn_over_grid_slice(pop, grid_type, met_Zsun,
+                                termination_flag='combined_TF12',
+                                slices=None, channel=None,
                                 plot_dir='./', prop=None, prop_range=None,
                                 log_prop=False, alpha=0.3, s=5.,
                                 show_fig=True, save_fig=True, close_fig=True,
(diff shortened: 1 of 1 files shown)


mkruckow:
Could you be a bit more specific, which functions got removed? E.g. `plot_merger_rate_density` looks to me like a function which was used to produce plots for papers in the past, thus I'm not sure, whether we should remove them.

posydon/visualization/plot_pop.py
```diff
@@ -186,207 +167,326 @@ def plot_hist_properties(df, ax=None, df_intrinsic=None, df_observable=None,
...
+    if channel is not None and 'channel' not in pop.population.columns:
         raise ValueError('Formation channel information not available in popsynth data.')
 
-    # load grid
+    # get population data from the popsyn for the given metallicity
+    data = get_population_data(pop          = pop,
+                               metallicity  = met_Zsun,
+                               grid_type    = grid_type,
+                               channel      = channel,
+                               prop         = prop)
```
mkruckow: That's not the style we'd like to use.

posydon/visualization/plot_pop.py
```diff
@@ -186,207 +167,326 @@ def plot_hist_properties(df, ax=None, df_intrinsic=None, df_observable=None,
...
+
+def plot_population_data(data,
+                         slice_3D_var_str,
+                         slice_3D_var_range,
+                         prop=None,
+                         prop_range=None,
+                         log_prop=False,
+                         alpha=0.3,
+                         s=5):
+    '''Plot the population data based on the grid slice'''
```
mkruckow: This doc string is incomplete, please add the missing information.

posydon/visualization/plot_pop.py
```diff
@@ -186,207 +167,326 @@ def plot_hist_properties(df, ax=None, df_intrinsic=None, df_observable=None,
...
+
+    # get only slice data
+    if slice_3D_var_str == 'mass_ratio':
+        var = 'q'
+    elif slice_3D_var_str == 'star_2_mass':
+        var = 'S2_mass'
```
mkruckow, line 449: If `slice_3D_var_str` is neither of the two `var` will be undefined.

posydon/visualization/plot_pop.py
```diff
@@ -186,207 +167,326 @@ def plot_hist_properties(df, ax=None, df_intrinsic=None, df_observable=None,
...
+                label = DEFAULT_LABELS[prop][1]
             else:
-                plt.scatter(log10_m1, log10_p, s=s, marker='v', color='black', alpha=alpha, zorder=0.5)
-
-            if save_fig:
-                plt.savefig(os.path.join(plot_dir, fname%var), bbox_inches='tight')
-            if show_fig:
-                plt.show()
-            if close_fig:
-                plt.close()
```
mkruckow, line 390: It looks a bit strange to not allow to save, show or close the figure after the plotting a few lines later.

posydon/visualization/plot_pop.py
```diff
@@ -186,207 +167,326 @@ def plot_hist_properties(df, ax=None, df_intrinsic=None, df_observable=None,
...
+    
+    Returns
+    -------
+    bin_centers : list
+        List of bin centers
+    bin_edges : list
+        List of bin edges
+    PLOT_PROPERTIES : dict
+        Dictionary of plot properties
+    slice_3D_var_str : str
```
mkruckow, line 376: There is a missing return value `tmp_fname`, please add it.

posydon/visualization/plot_pop.py
```diff
@@ -186,207 +167,326 @@ def plot_hist_properties(df, ax=None, df_intrinsic=None, df_observable=None,
...
+    close_fig : bool (optional)
+        Close the figure
+    plot_extension : str (optional)
+        File extension for saving the plot
+    verbose : bool (optional)
+        Print information
+    
+    Returns
+    -------
+    None
```
mkruckow: I'd remove the returns here, because there is no `return` statement in this function, hence it is treated as a procedure. Yes, python always returns a None in such cases, but this value is meaningless.

posydon/visualization/plot_pop.py
```diff
@@ -23,57 +23,38 @@
...
+    if ax is None, it plotted on the latest axis, otherwise it is added to the given axis
+    
+    Parameters
+    ----------
+    ax : matplotlib.axes.Axes (optional)
+        Axes object to plot the data on
+    
+    Returns
+    -------
+    None
```
mkruckow: I'd remove the returns here, because there is no return statement in this function, hence it is treated as a procedure. Yes, python always returns a None in such cases, but this value is meaningless.

posydon/visualization/plot_pop.py
```diff
@@ -186,207 +167,326 @@ def plot_hist_properties(df, ax=None, df_intrinsic=None, df_observable=None,
...
-                else:
-                    label = prop
-                    if log_prop:
-                        label = 'log10_'+label
-                plt.colorbar(label=label, orientation='horizontal')
+    PLOT_PROPERTIES = {
+        'show_fig' : False,
+        'close_fig' : False,
+        'log10_x' : True,
+        'log10_y' : True,
```
mkruckow, line 409: Are you sure you like to hard code this? Shouldn't it depend on the grid (currently all our girds are log spaced, but this might not hold for smaller science case specific grids.)?

posydon/visualization/plot_pop.py
maxbriel: This code is still present in L283-288. It's the top-level function that has this option. Each sub-function doesn't allow this.

It looks weird on the git diff/github website

posydon/visualization/plot_pop.py
maxbriel: Changed.

posydon/visualization/plot_pop.py
maxbriel: Done

posydon/visualization/plot_pop.py
maxbriel: Done

posydon/visualization/plot_pop.py
maxbriel: Done.

posydon/visualization/plot_pop.py
maxbriel: This is how it was in the old code, so I will leave it like that here. It would be useful to add as an enhancement though. Feel free to open a new issue for it, if you think it's worth it :)

posydon/visualization/plot_pop.py
maxbriel: Value error raised

posydon/visualization/plot_pop.py
maxbriel: Added

posydon/visualization/plot_pop.py
mkruckow: Sure, that's how it was in the old code. But isn't the aim of this PR to make the code more general usable?

posydon/visualization/plot_pop.py
mkruckow: Yes, there should either be a ValueError raised or a fall back which defines `var`. Currently it will raise a NameError because of the non existing variable.

posydon/visualization/plot_pop.py
mkruckow: I have seen that it got moved. So we will get more top level functions in the future to e.g. only plot the population data, which allows to save, show or close the figure? So we don't want to have the low level function to be called? If so, should we start its name with an underscore?

maxbriel:
> Could you be a bit more specific, which functions got removed? E.g. plot_merger_rate_density looks to me like a function which was used to produce plots for papers in the past, thus I'm not sure, whether we should remove them.

For paper plots, the associated release can still be used to create the plots. A more general rate calculation and plotting are present to replace these functions.

posydon/visualization/plot_pop.py
maxbriel: It is to make it more modular and clean up the plotting routines. However, I have not added any new functionality (besides a bug fix).
If you would like plotting over other grid sizes/types to become a thing, you can add it to enhancement #383 as a sub-issue.

mkruckow: I opened a general issue #503 .

mkruckow (approved):

posydon/visualization/plot_pop.py
maxbriel: I don't think this is needed. The user can create a figure themselves call this sub-function and then store/save the figure. The sub-functions don't manipulate the figures/plot saving, but you can still use them for your own plotting.

philipp-rajah (approved):
Looks good to me!

PR #484, opened 2025-01-16
No check for WD formation in double CE case

### LINKED ISSUE #485: Non-interacting binaries in unstable MT grid regime
I'm looking at the evolution of systems through "grid" slices of the HMS-HMS grid. I'm noticing that some systems evolve without interactions but should interact through a common envelope.
I've attached an example slice plot around q=0.05 of M1 vs Period to show this behaviour at Z=10^-4Zsun. This plot shows the interaction during step-HMS-HMS for the simulated models. The shorter period systems  (<10^3 days) should definitely interact (and do so on the grid; see the other plot), but don't interact seems to happen for the "low" mass systems in the population M < 30Msun.

simulated systems:
Image

grid slice:
Image

According to the grid, these models should undergo a common envelope (unstable mass transfer).
Looking at a specific model in this region, the interpolation class of the HMS-HMS grid is correct as "Unstable RLOF during postMS", but the evolution goes to CC1 (see image below).
There's no indication of unstable mass transfer or an interaction during step HMS-HMS in the history of the binary.
It seems related to the unstable nature of the interaction since the stable interactions (green) are correctly present in this mass regime.

Example binary:
Image

mkruckow:
Currently, the check for WD formation may partly overwrites the double CE case, which it shouldn't do.

- [x] Move double CE into contact detection and return afterwards to skip checks on WD formation.

### DIFF
--- a/posydon/unit_tests/utils/test_common_functions.py
+++ b/posydon/unit_tests/utils/test_common_functions.py
@@ -1151,6 +1151,17 @@ class TestFunctions:
         binary.star_2.center_gamma_history=[]
         assert totest.get_binary_state_and_event_and_mt_case(binary, i=0) ==\
                ['detached', None, 'None']
+        # both stars overfill RL leading to double CE while WD condition is
+        # fulfilled as well
+        tests = [(2.0, 1.0, 'oDoubleCE1'), (1.0, 2.0, 'oDoubleCE2')]
+        for (ro1, ro2, e) in tests:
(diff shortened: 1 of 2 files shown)


maxbriel (dismissed):
This works for me and seems to fix the issue of there not being an interaction in the low mass regime.

maxbriel (approved):

astroJeff (approved):

PR #496, opened 2025-01-27
updating the data_download routines slightly

astroJeff:
Adding a check to make sure that PATH_TO_POSYDON_DATA is defined. This keeps a user from accidentally downloading multiple versions of the same data. This will be important if/when we move to a partial data download paradigm, where files are only downloaded when needed.

Also this fixes a few small typos in the print statements

### DIFF
--- a/posydon/utils/data_download.py
+++ b/posydon/utils/data_download.py
@@ -57,9 +57,7 @@ def data_download(file=file, MD5_check=True, verbose=False):
     """
     # First, make sure the path does not exist
     if os.path.exists(file):
-        if verbose:
-            print('POSYDON data already exists at', file)
-        return
+        raise FileExistsError(f'POSYDON data already exists at {file}')
 
     # Second, check to make sure PATH_TO_POSYDON_DATA is defined
     if "PATH_TO_POSYDON_DATA" not in os.environ:
--- a/posydon/utils/data_download.py
+++ b/posydon/utils/data_download.py
@@ -11,7 +11,6 @@ import hashlib
 import progressbar
 import tarfile
 from tqdm import tqdm
-from posydon.utils.posydonerror import DataError
 
 # get path to data, if not provided use the working directory
 PATH_TO_POSYDON_DATA = os.environ.get("PATH_TO_POSYDON_DATA",'./')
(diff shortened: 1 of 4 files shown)


posydon/utils/data_download.py
```diff
@@ -58,7 +58,14 @@ def data_download(file=file, MD5_check=True, verbose=False):
...
-            print('POSYDON data alraedy exists at', file)
+            print('POSYDON data already exists at', file)
+        return
+
+    # Second, check to make sure PATH_TO_POSYDON_DATA is defined
+    if "PATH_TO_POSYDON_DATA" not in os.environ:
+        if verbose:
+            print('You must define the PATH_TO_POSYDON_DATA environment '
+                  'variable before downloading POSYDON datasets')
         return
```
mkruckow: Is there a reason, why we seal the information, that a download isn't done behind a verbose check?
I might even think of raising an error, because the function didn't do the job it was called for. At least I'd expect a warning, but surely not nothing (without verbose).

mkruckow (dismissed):
Actually, the function doesn't rely on `PATH_TO_POSYDON_DATA`. You can download the data to any place you like, just the default is the location specified by `PATH_TO_POSYDON_DATA` if this variable isn't given it downloads to the current directory.
Thus this currently undermines the existing structure, which may should be replaced otherwise.
Additionally, there is currently no unit-test for the new case.

astroJeff:
Yes, it is failing because of the lack of a unit test. But the bigger question is the one you raise. I think we want to force the user to set a POSYDON data path. Do you think otherwise?

Agree about raising an exception. Will add now.

posydon/utils/data_download.py
```diff
@@ -63,10 +64,8 @@ def data_download(file=file, MD5_check=True, verbose=False):
 
     # Second, check to make sure PATH_TO_POSYDON_DATA is defined
     if "PATH_TO_POSYDON_DATA" not in os.environ:
-        if verbose:
-            print('You must define the PATH_TO_POSYDON_DATA environment '
-                  'variable before downloading POSYDON datasets')
-        return
+        raise DataError('You must define the PATH_TO_POSYDON_DATA environment'
+                        ' variable before downloading POSYDON datasets')
```
mkruckow: Do we need a new error type here? I'd suggest to use one of the following two:
From the perspective of `os.environ` it is a `KeyError`.
If you like to think of a not defined variable in the environment given by the OS, it would be a `NameError`.

Additionally, I'd suggest to replace the verbose in the first check as well and put there a `FileExistsError`.

mkruckow (approved):

maxbriel (approved):

PR #500, opened 2025-01-30
add more dump_rate info to docs

### LINKED ISSUE #478: missing binaries in oneline dataframe
I am finding that the number of binaries in the oneline dataframe is not equivalent to the total number of binaries in the history dataframe, meaning that some binaries are missing from the oneline dataframe.

One example is in this population here: `/projects/b1119/cliotine/pulsar_dev/pulsar_pops/runs/jan_2025/Fryer+12_delayed/1e+00_Zsun_population.h5`
I ran this population with the 10/28/24 grids and the most recent development branch.

Screen Shot 2025-01-13 at 9 58 20 AM

This is obviously a major issue because binaries cannot be reliably connected between the two dataframes in post processing.

celiotine:
I have added more explicit information to the docs and default `.ini` file about the `dump_rate` settings to help users avoid Issue #478.

### DIFF
--- a/docs/_source/components-overview/pop_syn/population_params.rst
+++ b/docs/_source/components-overview/pop_syn/population_params.rst
@@ -652,7 +652,7 @@ It also contains which sampling distributions to use for the initial conditions 
 
   * - ``dump_rate``
     - | Batch save after evolving N binaries.
-    - | To facilitate I/O performance, this should be higher than 500 for populations of 100.000 binaries or more.
+    - | To facilitate I/O performance, this should be at least 500 for populations of 100.000 binaries or more.
     - ``2000``
 
   * - ``temp_directory``
--- a/docs/_source/components-overview/pop_syn/population_params.rst
+++ b/docs/_source/components-overview/pop_syn/population_params.rst
@@ -652,7 +652,7 @@ It also contains which sampling distributions to use for the initial conditions 
 
   * - ``dump_rate``
     - | Batch save after evolving N binaries.
-    - | To facilitate I/O performance, this should not be set below 500 for populations of 100.000 binaries or more. 
+    - | To facilitate I/O performance, this should be higher than 500 for populations of 100.000 binaries or more.
     - ``2000``
 
   * - ``temp_directory``
--- a/docs/_source/components-overview/pop_syn/population_params.rst
+++ b/docs/_source/components-overview/pop_syn/population_params.rst
@@ -652,6 +652,7 @@ It also contains which sampling distributions to use for the initial conditions 
 
   * - ``dump_rate``
     - | Batch save after evolving N binaries.
+    - | To facilitate I/O performance, this should not be set below 500 for populations of 100.000 binaries or more. 
     - ``2000``
 
   * - ``temp_directory``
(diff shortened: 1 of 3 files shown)


mkruckow:
I'm just a bit wondering about the dependency we should put up:
It you should increase the dump rate if you have more jobs running in parallel writing to the same file system.
Additionally, there could be the case to increase the dump rate if the same jobs writes too often, that the file system can't empty the buffer on time.

By the way, why do we call it dump rate, wouldn't the actual rate be the inverse of the value we put and the value in fact be the dump size?

celiotine:
I did not run tests to see how this is impacted by changing the number of job arrays. Maybe @maxbriel has an idea?

maxbriel:
If the dump_rate is high enough, it shouldn't matter for the number of job arrays. Technically, there will be some dependence, but it's not a one-to-one. Since evolving each binary takes a different duration, each job array will write at a different moment. This probably means it's easier for the system to handle compared to many consecutive writes from a single job. 

These descriptions look good to me for warning the user.
I would use a positive phrasing, but that's a personal preference.  Feel free to accept or reject them :) 


> By the way, why do we call it dump rate, wouldn't the actual rate be the inverse of the value we put and the value in fact be the dump size?

You could check "git blame" on Github to see when it was named, but someone probably gave it that name a long time ago.

docs/_source/components-overview/pop_syn/population_params.rst
```diff
@@ -652,6 +652,7 @@ It also contains which sampling distributions to use for the initial conditions
 
   * - ``dump_rate``
     - | Batch save after evolving N binaries.
+    - | To facilitate I/O performance, this should not be set below 500 for populations of 100.000 binaries or more.
```
maxbriel: Maybe change it to a positive phrasing?

```suggestion
    - | To facilitate I/O performance, this should be higher than 500 for populations of 100.000 binaries or more.
```

docs/_source/troubleshooting-faqs/code-questions.rst
```diff
@@ -46,6 +46,9 @@ Frequently Asked Questions
     
     The memory usage of the 8000 :code:`dump_rate` run is stable at around 6GB, while the 10.000 :code:`dump_rate` run is stable at around 6.8GB.
 
+    In general, the :code:`dump_rate` should not be set below 500 for populations of 100.000 binaries or more.
```
maxbriel: Same as the other comment
```suggestion
    In general, the :code:`dump_rate` should be set above 500 for populations of 100.000 binaries or more.
```

posydon/popsyn/population_params_default.ini
```diff
@@ -281,6 +281,7 @@
   # set maximum ram per cpu before batch saving (GB)
   dump_rate = 2000
   # batch save after evolving N binaries
+  # this should not be set below 500 for populations of 100,000 binaries or more
```
maxbriel:
```suggestion
  # this should be set above 500 for populations of 100,000 binaries or more
```

maxbriel (dismissed):

docs/_source/troubleshooting-faqs/code-questions.rst
astroJeff: I agree. However, the phrase proposed suggests the dump rate ought to be greater than 500, when we actually want the dump rate to be 500 or more. Fixing now...

posydon/popsyn/population_params_default.ini
astroJeff: Same as above. We want the phrasing to be such that the dump rate is *at least* 500

astroJeff (approved):

maxbriel (approved):

PR #501, opened 2025-02-04
chi_eff NaN replace by 0

### LINKED ISSUE #319: S1/S2 spins NaN
From @mpgalleg comment on the Population Analysis Framework:

> I tried to follow the tutorial and ran into an error when trying to calculate the observable population of BBH mergers using a population ran with NN. I did not run into this problem when using a population ran with IF.
> 
 ```
ValueError: Input X contains NaN.
KNeighborsRegressor does not accept missing values encoded as NaN natively. For supervised learning, you might want to consider sklearn.ensemble.HistGradientBoostingClassifier and Regressor which accept missing values encoded as NaNs natively. Alternatively, it is possible to preprocess the data, for instance by using an imputer transformer in a pipeline or drop samples with missing values. See https://scikit-learn.org/stable/modules/impute.html You can find a list of all estimators that handle NaN values at the following page: https://scikit-learn.org/stable/modules/impute.html#estimators-that-handle-nan-values
```
> 
> Seems like it is breaking when trying to predict points between the grid of sensitivity? @ZepeiX also had a similar error, but I do not know if it was with NN or IF population.
> 
> update:
> the problem is happening because some S1_spin and S2_spin are nan, which causes chi_eff values to be nan. Nan is provided to the function predict_pdet in selection_effects.py
> 
> It is happening to 142/41882 merging binary black holes. These BBH mergers are not specific to a particular formation channel or metallicity. They are coming from formation channels
> ‘ZAMS_oRLO1_CC1_oRLO2_CC2_END’, ‘ZAMS_oRLO1-contact_CC1_oRLO2_CC2_END’, ‘ZAMS_oRLO1_CC1_oRLO2_oCE2_CC2_END’, ‘ZAMS_CC1_oRLO2_oCE2_CC2_END’ with metallicities 0.1 , 0.01 , 0.001, 0.2.

Because the spins are NaN, this is not an issue with the population part, but with the output of the population run. This requires more triaging to figure out what's going on.

@mpgalleg, would you be able to provide an example binary that has this issue from a NN population? Feel free to provide a file with the population (on slack) that has this issue.

### PR DESCRIPTION
**Pull request by maxbriel**
Might close #319. But it has two parts:

1. NaN's in a nearest neighbour run, which might be cause by a negative spin. Unclear how we should proceed with this.
2. NaN's in a initial-final run. These are most likely caused be a mismatch between the interpolation class and CO type. We can't really fix these without fixing the classifier. This allows the `chi_eff` calculation to still work while there are NaN values in the spins.

When values are replaced, a warning is raised.

### DIFF
--- a/posydon/popsyn/transient_select_funcs.py
+++ b/posydon/popsyn/transient_select_funcs.py
@@ -147,17 +147,20 @@ def chi_eff(m_1, m_2, a_1, a_2, tilt_1, tilt_2):
         
     '''
     if pd.isna(a_1).any():
-        Pwarn("a_1 contains NaN, replacing with 0.0", 'ReplaceValueWarning')
+        Pwarn("a_1 contains undefined values, replacing them with 0.0",
+              'ReplaceValueWarning')
         a_1[pd.isna(a_1)] = 0.0
     if np.isnan(a_2).any():
-        Pwarn("a_2 contains NaN, replacing with 0.0", 'ReplaceValueWarning')
+        Pwarn("a_2 contains undefined values, replacing them with 0.0",
+              'ReplaceValueWarning')
         a_2[pd.isna(a_2)] = 0.0
     if np.isnan(tilt_1).any():
-        Pwarn("tilt_1 contains NaN, replacing with 0.0", 'ReplaceValueWarning')
-        
+        Pwarn("tilt_1 contains undefined values, replacing them with 0.0",
+              'ReplaceValueWarning')
         tilt_1[pd.isna(tilt_1)] = 0.0
     if np.isnan(tilt_2).any():
-        Pwarn("tilt_2 contains NaN, replacing with 0.0", 'ReplaceValueWarning')
+        Pwarn("tilt_2 contains undefined values, replacing them with 0.0",
+              'ReplaceValueWarning')
         tilt_2[pd.isna(tilt_2)] = 0.
     return (m_1*a_1*np.cos(tilt_1)+m_2*a_2*np.cos(tilt_2))/(m_1+m_2)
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: posydon/popsyn/transient_select_funcs.py [hunk 1]
```diff
@@ -123,7 +123,41 @@ def GRB_selection(history_chunk, oneline_chunk, formation_channels_chunk=None, S
...
+    
+    Returns
+    -------
+    np.ndarray
+        The effective spin of the BHs.
+        
+    '''
+    if np.isnan(a_1).any():
+        Pwarn("a_1 contains NaN, replacing with 0.0", 'ReplaceValueWarning')
+        a_1 = np.nan_to_num(a_1, nan=0.0)
```
mkruckow: `nan_to_num` does not work if any of the values in the ndarray is `None`, e.g. `np.nan_to_num(np.array([1.0, np.nan, None]), nan=0.0)` returns `array([1.0, nan, None], dtype=object)`
Additionally, we wanted to move away from `np.isnan`? Shell we use something like `a_1[pd.isna(a_1)]=0.0` instead?

astroJeff: I like this idea. The code you suggest also catches `None` values, so we should update the warning message accordingly

astroJeff:
@maxbriel have you checked this with a new population? I understand it's a rare error.

posydon/popsyn/transient_select_funcs.py
maxbriel: Implemented.

maxbriel:
@astroJeff I ran a NN run ~two weeks ago and I didn't see any problematic values there. I did also run more 2 populations with the [max_newPISN](https://github.com/POSYDON-code/POSYDON/pull/417/commits/f475744fc481dc523495165187cd2c6cd4a9afdc) branch. In the highest PISN shift, I found 2 binaries in a total run of 8.000.000 binaries with a NaN spin as described in #319. 

I haven't run an IF run recently, but my guess is that it will happen more often there due to a mismatch between the interpolation class and CO classifiers.

astroJeff (dismissed):

mkruckow (dismissed):

astroJeff (dismissed):

astroJeff (approved):

mkruckow (approved):

PR #504, opened 2025-02-05
Decouple ECSN treatment from SN_MODELS

### LINKED ISSUE #312: Use of Tauris+15 for ECSN
Podsiadlowski+04 seems to overestimate the range under which ECSNe exist (preliminary results from @dimsour94 ) even for single stars. In POSYDON this would affect the kick of various NS binaries and thus will overestimate the bound systems after CC1

Tauris+15 seems to have a more "reasonable" rate of ECSNe. Two suggestions to move forward would be:

- Use Tauris+15 as default , but this will affect various default results in NS XRBs, compact object mergers, etc

- At least allow Tauris+2015 as an option in the interpolated post-processed SN MODELS in grids/MODELS. Right now all 10 of them allow only for Podsiadlowski+04

mkruckow:
Currently, ECSN prescriptions are limited to be resolved in the grids, otherwise there won't be interpolators trained on them.

To circumvent this issue, it would be better to check for the interpolated core masses always and determine the NS being formed via ECSN not from the SN_MODELS.

- [x] escape the matching of ECSN when checking the SN_MODEL
- [x] check for ECSN to skip the usage of the SN interpolator values
- [x] check stuff just below or above the ECSN regime (At the moment it only throws a warning and tries to recalculate the SN properties)
- [x] tested on small pop-syn

### DIFF
--- a/posydon/binary_evol/SN/step_SN.py
+++ b/posydon/binary_evol/SN/step_SN.py
@@ -546,9 +546,9 @@ class StepSN(object):
                 else:
                     MODEL_properties = getattr(star, MODEL_NAME_SEL)
 
-                    SN_type = check_SN_type(m_core=star.co_core_mass,
-                                            m_He_core=star.he_core_mass,
-                                            m_star=star.mass)[3]
+                    SN_type = self.check_SN_type(m_core=star.co_core_mass,
+                                                 m_He_core=star.he_core_mass,
+                                                 m_star=star.mass)[3]
                     if self.use_profiles and profile is not None:
                         alternative = "Instead use profiles."
                     elif self.use_core_masses:
--- a/posydon/binary_evol/SN/step_SN.py
+++ b/posydon/binary_evol/SN/step_SN.py
@@ -714,6 +714,10 @@ class StepSN(object):
                     # is used to get the compact object spin
                     star.mass = m_grav
                     if m_grav >= self.max_NS_mass:
+                        if SN_type == "ECSN":
+                            Pwarn("An ECSN should not form a black hole: "
+                                  f"m_grav={m_grav}.",
+                                  "InappropriateValueWarning")
                         # see Eq. 14, Fryer, C. L., Belczynski, K., Wiktorowicz,
                         # G., Dominik, M., Kalogera, V., & Holz, D. E. (2012), ApJ, 749(1), 91.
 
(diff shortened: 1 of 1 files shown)


posydon/binary_evol/SN/step_SN.py
```diff
@@ -679,7 +709,7 @@ def collapse_star(self, star):
                     star.h1_mass_ej, star.he4_mass_ej = \
                         get_ejecta_element_mass_at_collapse(star,star.mass,verbose=self.verbose)
 
-                elif self.use_core_masses:
+                elif self.use_core_masses or SN_type == "ECSN":
                     # If the profile is not available the star spin
```
astroJeff, line 717: We should not allow a system with an SN_type of ECSN to evolve into a BH. Rather, in such cases, m_grav ought to be adjusted downwards. For what it's worth, I don't expect this to ever happen.

mkruckow: I added a warning for this case which should never happen.

astroJeff (dismissed):

astro-abhishek:
@mkruckow I ran test populations of 100k binaries using the Tauris+15 ECSN prescription, and with the fixes in this PR, NSs are now being produced through electron captures. I confirmed this for DNS systems where none were being formed before.

@astroJeff @tassos25 With this fix, the # of DNS produced with the Tauris+15 ECSN (with pre-SN C/O core masses: [1.37, 1.43]) is lower than what I have been getting till now with the Podsiadlowski+04 prescripiton (ECSN for pre-SN He core masses: [1.4, 2.5]). This means that the DNS rates need to be updated and possibly other results.

astro-abhishek (dismissed):

astroJeff:
@astro-abhishek, can you perform a couple consistency checks to make sure the PR is behaving as expected: 1) all the ECS systems are producing NSs, 2) the systems that have SN_type of ECS indeed have CO core masses in the right range, 3) there are no NSs (or BHs) being produced with CO core masses less than this mass range, 4) there are no WDs with larger CO core masses, and 5) there are no non-ECS systems with CO core masses in the Taurus range.

astro-abhishek:
Consistency checks for this PR (population size 100k binaries):

- [x] All systems that went through ECSN formed a NS.
- [x] All systems that went through ECSN had CO core masses in Tauris range (min is around 1.37004 and max is 1.42997).
- [x] No system with pre-SN CO core mass < 1.37 Msun forms a NS or BH.
- [x] No system with pre-SN CO core mass > 1.43 Msun forms a WD.
- [x] All systems with pre-SN CO core mass in the range [1.37, 1.43] Msun goes through ECSN.

mkruckow:
@astro-abhishek asked to **keep the branch after merging** to allow him to take over the changes in his branch/fork for DNS.

astro-abhishek:
Here's a quick update on this PR. @mkruckow asked me to check how often the new warnings in this PR are getting triggered:

- `ECSN in SN_MODEL replaced by:` 2572 occurrences 

- `The SN_type does not match the star`: 280 occurrences

- `An ECSN should not form a black hole`: 0 occurrences

The first warning is supposed to reflect the # of reductions in ECSN cases from Podsiadlowski+04 to Tauris+15 (out of 100k binaries).
The second warning gives an estimate for the occurrence of misclassifications from the interpolator.
The third warning should never happen, when all works as designed.

sgossage (dismissed):

sgossage (approved):

astroJeff (approved):

PR #505, opened 2025-02-07
Use of `PATH_*` variables

### LINKED ISSUE #165: Remove degenerate call to `PATH_TO_POSYDON` and  `PATH_TO_POSYDON_DATA`
After PR#144, we decided to move the imports of `PATH_TO_POSYDON` and  `PATH_TO_POSYDON_DATA` into `posydon/config.py`, this is cleaner for the end user that does not need to remember where to grab the paths from the `posydon/utils/` files. The code needs to be made consistent to have all the imports to these paths coming form the `config.py` file.

mkruckow:
We need a common way to deal with our path variables defined in the environment.

- [x] update `config.py`
- [x] add unit test for `config.py`
- [x] import PATHS always from `config.py`
   - include "POSYDON_data" in the python variable `PATH_TO_POSYDON_DATA` on load instead of adding it later
- [x] tested on small pop-syn

### DIFF
--- a/posydon/config.py
+++ b/posydon/config.py
@@ -4,7 +4,7 @@ from dotenv import load_dotenv
 load_dotenv()
 
 def ensure_path(name):
-    """Get the PATH set in the environment
+    """Get the 'PATH_*' variable set in the environment
     
     Parameters
     ----------
--- a/posydon/config.py
+++ b/posydon/config.py
@@ -30,4 +30,5 @@ def ensure_path(name):
 
 # POSYDON environment variables
 PATH_TO_POSYDON = ensure_path("PATH_TO_POSYDON")
-PATH_TO_POSYDON_DATA = ensure_path("PATH_TO_POSYDON_DATA")
+PATH_TO_POSYDON_DATA = os.path.join(ensure_path("PATH_TO_POSYDON_DATA"),
+                                    "POSYDON_data")
--- a/posydon/config.py
+++ b/posydon/config.py
@@ -3,6 +3,31 @@ from dotenv import load_dotenv
 
 load_dotenv()
 
+def ensure_path(name):
+    """Get the PATH set in the environment
+    
+    Parameters
+    ----------
+    name : str
+        Name of the environment variable.
+    
+    Returns
+    -------
+    str
+        Defined path or raises an error if the variable is not defined and
+        pointing to a valid path.
+    
+    """
+    if not isinstance(name, str):
+        raise TypeError("'name' has to be a string.")
+    value = os.getenv(name)
+    if value is None:
+        raise NameError(f"{name} is not defined in the environment.")
+    if not os.path.isdir(value):
+        raise NotADirectoryError(f"{value} given in {name} is an invalid "
+                                 "path.")
+    return value
+
 # POSYDON environment variables
-PATH_TO_POSYDON = os.getenv("PATH_TO_POSYDON")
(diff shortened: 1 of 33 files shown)


posydon/config.py
```diff
@@ -3,6 +3,32 @@
...
+        raise NotADirectoryError(f"{value} given in {name} is an invalid "
+                                 "path.")
+    return value
+
 # POSYDON environment variables
-PATH_TO_POSYDON = os.getenv("PATH_TO_POSYDON")
-PATH_TO_POSYDON_DATA = os.getenv("PATH_TO_POSYDON_DATA")
\ No newline at end of file
+PATH_TO_POSYDON = ensure_path("PATH_TO_POSYDON")
+PATH_TO_POSYDON_DATA = os.path.join(ensure_path("PATH_TO_POSYDON_DATA"),
```
elizabethteng, line 33: doesn't this enforce that `PATH_TO_POSYDON_DATA` is inside `PATH_TO_POSYDON`? What if someone is using data that are stored separately to their posydon repo?

posydon/popsyn/transient_select_funcs.py
```diff
@@ -12,7 +12,7 @@
 # We should alter the code to remove these warnings, but for now we suppress them
 warnings.simplefilter(action='ignore', category=pd.errors.PerformanceWarning)
 
-PATH_TO_PDET_GRID = os.path.join(PATH_TO_POSYDON_DATA, 'POSYDON_data/selection_effects/pdet_grid.hdf5')
```
elizabethteng, line 15: did we change the way we structure the posydon data? or should `POSYDON_data` be part of `PATH_TO_POSYDON_DATA` when it wasn't before? (I may just be behind on this)

elizabethteng (dismissed):

posydon/config.py
mkruckow: No it doesn't, there are two independent variables `PATH_TO_POSYDON` and `PATH_TO_POSYDON_DATA`, hence they can be at different places. In our installation guide we recommended the following relation: `PATH_TO_POSYDON_DATA = os.path.join(PATH_TO_POSYDON, "data")`, while it isn't mandatory. I guess, this is an artifact from the time data was stored in a submodule on git-lfs.
For some reason, I don't know as it was before my time, someone used os.path.join(PATH_TO_POSYDON_DATA,  "POSYDON_data") to load the data from there. I guess it was a consequence of how the data archive was packed: with the directory "POSYDON_data" to be unpacked in `PATH_TO_POSYDON_DATA`. If we'd like to change that we'd need to change the published data sets.

posydon/popsyn/transient_select_funcs.py
mkruckow: Previously, the "POSYDON_data" was added at all places where `PATH_TO_POSYDON_DATA` was used, now we add that part already after loading the variable from the environment. Thus, we don't need to add it at all places in the code individually.

posydon/config.py
```diff
@@ -3,6 +3,32 @@
 
 load_dotenv()
 
+def ensure_path(name):
+    """Get the PATH set in the environment
```
maxbriel:
```suggestion
    """Get the path name set in the environment
```
I would like to suggest not using PATH in the doc string, since it's an environment variable in Linux and Mac (maybe Windows). It's confused me.
Same for the PR name ;)

maxbriel (dismissed):

maxbriel (approved):

elizabethteng (approved):

PR #507, opened 2025-02-13
Modular data download

### LINKED ISSUE #87: rethink POSYDON data v2 automatic download method
Eventually, we will be releasing v2 data on Zenodo; given the huge size of the new datasets 30+ GB we need to implement methods that only automatically pull parts of the datasets so that it will avoid overwhelming the user with unnecessary datasets. E.g., locally, one might only need SFR, p_det, or core-collapse table to do some computations and does not need the parts of the datasets containing grids/interpolates. We might also want to have the user confirm before downloading any data.

NOTE: It was decided we do not want to support dataset releases/versioning controls on git LFS for v2 as in v1.

mkruckow:
With getting more datasets we need a better scalable and user frindly way to get the datasets.

IMPORTANT: As this PR makes changes in the executables, you may need to reinstall POSYDON via `pip install -e .` in your POSYDON directory and environment.

- [x] create data sets
- [x] update `data_download` functions
- [x] add functionality for command line interaction with `get-posydon-data`
- [x] update files calling `data_download`: for now only todos added to download only needed parts as soon as we have more specialized datasets
- [x] update docs: installation steps 4 and 5 switched and adding text how to get the help for the `get-posydon-data` which should than allow the user to use the command for their needs.
- [x] testing:
  - help
  - list all datasets
  - try verbose and md5check options
  - download:
    - single set
    - multiple sets
- [x] tested on small pop-syn

### DIFF
--- a/docs/_source/getting-started/installation-guide.rst
+++ b/docs/_source/getting-started/installation-guide.rst
@@ -46,19 +46,24 @@ Using Anaconda (Recommended)
 
         conda install -c posydon posydon
 
+    .. note:: 
+        Please remember the current path (get it via :code:`pwd`) or the one you specified for the next step.
+
 .. _posydon-env:
 
 4. **Set Environment Variables**
 
-    Export the required paths (please change the location names accordingly to your installation):
+    Export the required paths (please change the location names accordingly to your installation from the previous step):
 
     .. code-block:: bash
 
         export PATH_TO_POSYDON=/path/to/your/posydon/installation
         export PATH_TO_POSYDON_DATA=/path/where/you/want/to/store/data
 
+    The path for the data location is up to your choice but we recommend :code:`PATH_TO_POSYDON_DATA=$PATH_TO_POSYDON/data`.
+
     .. note:: 
-        You can add these lines to your ``~/.bashrc`` or ``~/.bash_profile`` or your shell equivalent to ensure the environment variables are set every time you open a new terminal.
+        You can add these lines to your :code:`~/.bashrc` or :code:`~/.bash_profile` or your shell equivalent to ensure the environment variables are set every time you open a new terminal.
 
 5. **Download the Dataset**
 
--- a/docs/_source/getting-started/installation-guide.rst
+++ b/docs/_source/getting-started/installation-guide.rst
@@ -48,20 +48,7 @@ Using Anaconda (Recommended)
 
 .. _posydon-env:
 
-4. **Download the Dataset**
-
-    .. warning::
-        The POSYDON v2.0.0 dataset is not yet available on Zenodo. The above instructions currently point to the POSYDON v1.0.0 dataset release. Please refer to the development version of the dataset available on Northwestern and UNIGE HPC facilities for now. To have access to latest pre-release dataset (230914) you must be a POSYDON core developer, please refer to the #developers Slack channel.
-
-    You can use POSYDON's built-in API command (by default the downloaded data will be saved in the current directory, hence please navigate to your desired location first):
-
-    .. code-block:: bash
-
-        get-posydon-data
-
-    Alternatively, you can manually download the dataset from Zenodo using the provided `link <https://zenodo.org/communities/posydon/>`_. (TODO: update link to v2)
-
-5. **Set Environment Variables**
+4. **Set Environment Variables**
 
     Export the required paths (please change the location names accordingly to your installation):
 
@@ -72,6 +59,21 @@ Using Anaconda (Recommended)
 
     .. note:: 
         You can add these lines to your ``~/.bashrc`` or ``~/.bash_profile`` or your shell equivalent to ensure the environment variables are set every time you open a new terminal.
+
+5. **Download the Dataset**
+
+    .. warning::
+        The POSYDON v2.0.0 dataset is not yet available on Zenodo. The instructions currently point to the POSYDON v1.0.0 dataset release. Please refer to the development version of the dataset available on Northwestern and UNIGE HPC facilities for now. To have access to latest pre-release dataset (241028) you must be a POSYDON core developer, please refer to the #developers Slack channel.
+
+    You can use POSYDON's built-in API command (the downloaded data will be saved in the directory specified by :code:`PATH_TO_POSYDON_DATA`):
+
+    .. code-block:: bash
+
+        get-posydon-data
+
+    May use :code:`get-posydon-data -h` to see all the options for this command, which allows to list all the datasets and download the one of your choice.
+
+    Alternatively, you can manually download the datasets from Zenodo. You can find the POSYDON datasets on the `POSYDON community <https://zenodo.org/communities/posydon/>`_ on Zenodo.
 
 .. _dev-version:
 
--- a/posydon/utils/data_download.py
+++ b/posydon/utils/data_download.py
@@ -37,7 +37,10 @@ def _parse_commandline():
                         nargs='?',
                         default='v1')
     parser.add_argument('-l', '--listedsets',
-                        help="list the datasets: 'complete' shows the full dataset able to run POSYDON, 'individual' lists the datasets on zenodo, which might need others to run population synthesis (default: complete)",
+                        help="list the datasets: 'complete' shows the full "
+                             "dataset able to run POSYDON, 'individual' lists "
+                             "the datasets on zenodo, which might need others "
+                             "to run population synthesis (default: complete)",
                         nargs='?',
                         const='complete',
                         choices=['complete', 'individual'])
--- a/posydon/utils/data_download.py
+++ b/posydon/utils/data_download.py
@@ -37,7 +37,7 @@ def _parse_commandline():
                         nargs='?',
                         default='v1')
     parser.add_argument('-l', '--listedsets',
-                        help="list the datasets: (default: complete)",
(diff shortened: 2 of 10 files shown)


docs/_source/getting-started/installation-guide.rst
```diff
@@ -48,30 +48,32 @@ Using Anaconda (Recommended)
...
-    .. warning::
-        The POSYDON v2.0.0 dataset is not yet available on Zenodo. The above instructions currently point to the POSYDON v1.0.0 dataset release. Please refer to the development version of the dataset available on Northwestern and UNIGE HPC facilities for now. To have access to latest pre-release dataset (230914) you must be a POSYDON core developer, please refer to the #developers Slack channel.
-
-    You can use POSYDON's built-in API command (by default the downloaded data will be saved in the current directory, hence please navigate to your desired location first):
+    Export the required paths (please change the location names accordingly to your installation):
 
     .. code-block:: bash
 
-        get-posydon-data
+        export PATH_TO_POSYDON=/path/to/your/posydon/installation
```
sgossage, line 60: We should consider adding some brief advice on where this path should be in the case of a conda installation, as it may be less obvious to users. Perhaps here or in the FAQ?

We can provide a blurb advising to use e.g., 
```
import posydon
posydon.__file__
```
which outputs `/path/to/your/posydon/installation/__init__.py` and taking the base path.

sgossage (dismissed):

docs/_source/getting-started/installation-guide.rst
sgossage: The code looks good. I plan to perform a test run.

posydon/utils/data_download.py
```diff
@@ -40,40 +76,92 @@ def __call__(self, block_num, block_size, total_size):
...
+        MD5_check = False
+        Pwarn("MD5 undefined, skip MD5 check.", "ReplaceValueWarning")
+    filename = os.path.basename(data_url)
+    directory = os.path.dirname(PATH_TO_POSYDON_DATA)
+    filepath = os.path.join(directory, filename)
+    if not os.path.isdir(os.path.dirname(filepath)):
+        raise NotADirectoryError("PATH_TO_POSYDON_DATA does not refer to a "
+                                 "valid directory.")
+    if os.path.exists(filepath):
+        raise FileExistsError(f"POSYDON data already exists at {filename}.")
```
sgossage: may be good to use `filepath` (rather than `filename`) here in the `FileExistsError` message so users more readily see where they have pre-existing data.

docs/_source/getting-started/installation-guide.rst
sgossage: The `get-posydon-data` command seems to be working as expected on my end.

posydon/utils/data_download.py
mkruckow: Sure!

docs/_source/getting-started/installation-guide.rst
mkruckow: I have adopted the text, while strictly speaking the `PATH_TO_POSYDON` is not part of this PR.
By the way, your suggested way gives a wrong path, because it would point to `$PATH_TO_POSYDON/posydon`.
Hence, instead I connected it to the previous step of the installation.

posydon/utils/datasets.py
```diff
@@ -0,0 +1,167 @@
...
+             + "super-Eddington accretion, compatible with code release "\
+             + "v2.0.0-pre1",
+    'url': "https://zenodo.org/records/14216817"
+}
+COMPLETE_SETS['v1'] = ['v1_for_v2.0.0-pre1']
+COMPLETE_SETS['super-Eddington_v1'] = ['v1_for_v2.0.0-pre1',
+                                       'super-Eddington_v1']
+
+# v2
+ZENODO_COLLECTION['v2_girds_2Zsun'] = {
```
sgossage:
```suggestion
ZENODO_COLLECTION['v2_grids_2Zsun'] = {
```

and likewise for other entries here.

sgossage:
Something to consider...if it does not exist yet, I think some documentation for the terminology "individual" and "complete" set could be good for the user.

I've tested the `get-posydon-data -l` call and it works well. The main concern I have is that the information it outputs would be hard to parse for a first time user. Either in our docs, or maybe there's a way to incorporate this information in the console output.

It currently outputs this:

```
Defined complete sets are:
  - 'v1' consisting of:
    - 'v1_for_v2.0.0-pre1': Re-postprocessed POSYDON v1.0 dataset compatible
                            with code release v2.0.0-pre1
  - 'super-Eddington_v1' consisting of:
    - 'v1_for_v2.0.0-pre1': Re-postprocessed POSYDON v1.0 dataset compatible
                            with code release v2.0.0-pre1
    - 'super-Eddington_v1': Re-postprocessed POSYDON v1.0 dataset, assuming
                            super-Eddington accretion, compatible with code
                            release v2.0.0-pre1
  - 'v2' consisting of:
    - 'auxiliary': Auxiliary POSYDON data
    - 'v2_girds_2Zsun': POSYDON v2.0 dataset at 2 Zsun
    - 'v2_girds_1Zsun': POSYDON v2.0 dataset at Zsun
    - 'v2_girds_0.45Zsun': POSYDON v2.0 dataset at 0.45 Zsun
    - 'v2_girds_0.2Zsun': POSYDON v2.0 dataset at 0.2 Zsun
    - 'v2_girds_0.1Zsun': POSYDON v2.0 dataset at 0.1 Zsun
    - 'v2_girds_0.01Zsun': POSYDON v2.0 dataset at 0.01 Zsun
    - 'v2_girds_1e-3Zsun': POSYDON v2.0 dataset at 10^{-3} Zsun
    - 'v2_girds_1e-4Zsun': POSYDON v2.0 dataset at 10^{-4} Zsun
  - 'v2_2Zsun' consisting of:
    - 'auxiliary': Auxiliary POSYDON data
    - 'v2_girds_2Zsun': POSYDON v2.0 dataset at 2 Zsun
  - 'v2_1Zsun' consisting of:
    - 'auxiliary': Auxiliary POSYDON data
    - 'v2_girds_1Zsun': POSYDON v2.0 dataset at Zsun
  - 'v2_0.45Zsun' consisting of:
    - 'auxiliary': Auxiliary POSYDON data
    - 'v2_girds_0.45Zsun': POSYDON v2.0 dataset at 0.45 Zsun
  - 'v2_0.2Zsun' consisting of:
    - 'auxiliary': Auxiliary POSYDON data
    - 'v2_girds_0.2Zsun': POSYDON v2.0 dataset at 0.2 Zsun
  - 'v2_0.1Zsun' consisting of:
    - 'auxiliary': Auxiliary POSYDON data
    - 'v2_girds_0.1Zsun': POSYDON v2.0 dataset at 0.1 Zsun
  - 'v2_0.01Zsun' consisting of:
    - 'auxiliary': Auxiliary POSYDON data
    - 'v2_girds_0.01Zsun': POSYDON v2.0 dataset at 0.01 Zsun
  - 'v2_1e-3Zsun' consisting of:
    - 'auxiliary': Auxiliary POSYDON data
    - 'v2_girds_1e-3Zsun': POSYDON v2.0 dataset at 10^{-3} Zsun
  - 'v2_1e-4Zsun' consisting of:
    - 'auxiliary': Auxiliary POSYDON data
    - 'v2_girds_1e-4Zsun': POSYDON v2.0 dataset at 10^{-4} Zsun
```

and when giving a bad positional argument to the command, it returns

```
get-posydon-data: error: argument -l/--listedsets: invalid choice: 'badinput' (choose from 'complete', 'individual')
```

where there isn't much info on what the difference between these is.

I was looking into whether we might organize this information (e.g., grid name, compatibility, other notes) into a table. The Python package `rich` looked like it could be nice for terminal output. However, that would be yet another required package, so maybe we'd want something more lightweight.

posydon/utils/datasets.py
mkruckow: corrected.

mkruckow:
I have added information in the help texts.
I was considering to use a table in the first place, but took a tree structure to allow us more flexibility in the length of names. If we go for a table, we'd need to restrict the size of each column, hence the length of the keys in the dictionaries in `datasets.py`.
@sgossage, could you be more specific which information you are missing (please have in mind, that there is the `-v` option) or is there too much when getting the list?

sgossage:
I think the verbosity is helpful, but I don't see much info for new users on what the difference between `complete` and `individual` sets are. Maybe I've overlooked that info though. 

I'm not sure if it's better in the FAQ maybe, but I think some guidance on what these are and maybe simple use case examples for why a user might want to download an `individual` set vs. a `complete` set.

If something simple can be added to the `-h` message, maybe that could work.

mkruckow:
There is some information in the help text. I'm not sure, whether and if so where we'd like to put more. Usually users should only think about the complete sets unless they are knowing how to combine different sets and/or want to create some own grids.

sgossage:
> There is some information in the help text. I'm not sure, whether and if so where we'd like to put more. Usually users should only think about the complete sets unless they are knowing how to combine different sets and/or want to create some own grids.

Sounds good -- maybe something to consider moving forward.

sgossage (approved):

astroJeff (approved):

PR #493, opened 2025-01-24
Rework Population weighting

maxbriel:
This reworks the population resampling to allow for more variables:

- [x] Split out the IMFs
- [x] Add unit testing to IMFs
- [x] Add support for multiple mass ratio distributions
- [x] Add support for additional period distributions
- [x] Add support for more binary fractions (currently fb=0 is not possible even if single stars are evolved)
- [x] Add support for additional IMFs?

PR #509, opened 2025-02-18
Fix logic for single star fraction population normalisation

maxbriel:
This fixes a bug in the logic.

For `f_bin_simulated != 1`, the logic would result in the second if/else statement always being run, despite wanting to run the `f_bin_nature == 0`.


mkruckow (approved):
Won't that become obsolete with PR #493 ?

sgossage (approved):

elizabethteng (approved):

PR #511, opened 2025-02-19
Isolate matching code

### LINKED ISSUE #117: Spin matching out on its own step
Right now, matching and detached evolution occur in the same step. These should be spun off into independent steps, so we can separately identify the effect of the matching algorithm and the effect of the binary evolution in the detached step.

sgossage:
This substantially restructures `step_detached.py`, isolating the code that performs track matching into its own set of files: `track_match.py` and some auxiliaries (`key_library.py`).

1. Matching can be performed in a step after setting up the `TrackMatcher` class with its `do_matching()` function. 
2. The result of the matching can be stored in the binary history using an optional boolean `record_matching`, which is enabled as a keyword for `step_detached` and can be set in `population_params.ini`.

Here's an example binary with the new matching output turned on (the binary `event` is displayed as either `Match1`, `Match2`, or `Match12` when star 1, 2, or stars 1 and 2 are matched), with the matching output in the highlighted row:

<details><summary>image</summary>

Screenshot 2025-06-12 154932

</details> 

The verbose output was also updated, which now looks like this for the same system:

<details><summary>image</summary>

Screenshot 2025-06-12 155224

</details>

___

> [!NOTE]
><details><summary> Some of the major changes:</summary>
><ul>
><li> A new TrackMatcher class that handles functionalities related to matching binary star models to single star models in a new   `track_matcher.py`.
><li> Algorithmically defined star states for our 3 matching scenarios (hopefully to ensure consistent definitions): HMS, post-MS, stripped He. This is done in `posydon/binary_evol/flow_chart.py`. ⚠️ These are a bit different from how they were defined before, but consistent with our papers.
><li> A new `key_library.py` in `posydon.binary_evol.DT` that holds common lists of data column names needed to access and translate column names between POSYDON h5 files and MESA. This is to avoid repetitive code throughout `step_detached.py` and `track_matcher.py`
><li> Adds percent difference between previous step and matched quantities to verbose output during matching
><li> Adds ability to record matching step in binary history, as a substep of the current step using a population synthesis `.ini` file option.
><li> A new method for `binarystar.py`, `check_who_exists()`, which can be used to set new class attributes `self.non_existent_companion` and `self.companion_1_exists`/`self.companion_2_exists`. These were previously variables of the `detached_step` that may more appropriately fit as BinaryStar class attributes.
><li>BUG FIX: Fixes an error in assigning calculated values from detached evolution to the mass loss rates of star 1. Previously, star 1 copied the values of star 2. ⚠️ Unclear how this may alter pop synth results, but it is now working as intended. Systems affected would probably have been rare survivors of a common envelope.
></ul>
></details> 

<details><summary>Tasks</summary>

- [x] Isolate code relevant to matching
- [x] Implement Camille's update to the matching algorithm from PR#521
- [x] Incorporate changes from PR#532
- [x] Update doc strings, cleaning
- [x] Test everything works
- [x] Add matching step to flow
</details>


mkruckow:
The matching should not go into `girds`, it belongs to `interpolation`.

sgossage:
I will update this PR with Camille's changes in [PR#521](https://github.com/POSYDON-code/POSYDON/pull/521)

posydon/grids/track_match.py
```diff
@@ -829,8 +829,20 @@ def get_MESA_labels(list_for_matching):
...
+                percent_diff = [
+                                "% difference",
+                                f'{100.0*abs(star.mass - self.get_track_val("mass", htrack, *best_sol.x))/star.mass :.3f}', 
+                                f'{100.0*abs(star.log_R - self.get_track_val("log_R", htrack, *best_sol.x))/star.log_R:.3f}',
+                                f'{100.0*abs(star.center_h1 - self.get_track_val("center_h1", htrack, *best_sol.x))/star.center_h1:.3f}',
+                                f'{100.0*abs(star.surface_h1 - self.get_track_val("surface_h1", htrack, *best_sol.x))/star.surface_h1:.4f}',
+                                f'{100.0*abs(star.he_core_mass - self.get_track_val("he_core_mass", htrack, *best_sol.x))/star.he_core_mass:.3f}',
+                                f'{100.0*abs(star.center_he4 - self.get_track_val("center_he4", htrack, *best_sol.x))/star.center_he4:.4f}',
+                                f'{100.0*abs(star.surface_he4 - self.get_track_val("surface_he4", htrack, *best_sol.x))/star.surface_he4:.4f}',    
+                                f'{100.0*abs(star.center_c12 - self.get_track_val("center_c12", htrack, *best_sol.x))/star.center_c12:.4f}'
```
mkruckow: minor: please check the 80 characters limit

mkruckow: I'd remove the `abs` as the sign tell you the direction it is off. E.g. for the mass an increase (gain) is worth then a decrease (loss).

astroJeff:
@sgossage can you check to make sure the text in the v2 paper is consistent with the updates you have made here?

philipp-rajah (dismissed):
Reviewed for stylistic changes, looks good to me.

astroJeff:
@sgossage I reviewed your branch and made some pretty significant changes to your matching class. Please pull the latest and take a look at the updates I made.

sgossage:
Thanks, @astroJeff , I reviewed your changes and made some revisions to them to fix scope errors and also further functionalize things. Please have a look and see what you think.

I plan to update the docstrings for the new code accordingly.

posydon/binary_evol/DT/step_detached.py
```diff
@@ -11,230 +11,42 @@
...
                                             set_binary_to_failed)
-from posydon.utils.interpolators import PchipInterpolator2
 from posydon.binary_evol.flow_chart import (STAR_STATES_CC, 
-                                            STAR_STATES_CO, 
+                                            STAR_STATES_CO,
+                                            STAR_STATES_H_RICH,
                                             STAR_STATES_H_RICH_EVOLVABLE,
-                                            STAR_STATES_HE_RICH_EVOLVABLE)
+                                            STAR_STATES_HE_RICH_EVOLVABLE,
+                                            STAR_STATES_FOR_Hestar_MATCHING)
```
maxbriel: `STAR_STATES_FOR_Hestar_MATCHING` is not used in the end. Why?
Same for `STAR_STATES_CO` and `STAR_STATES_H_RICH`. 

Should they be used or can they be removed?

posydon/binary_evol/DT/step_detached.py
```diff
@@ -1956,9 +957,10 @@ def diffeq(
         do_stellar_evolution_and_spin_from_winds=True,
         verbose=False,
 ):
-    """Diff. equation describing the orbital evolution of a detached binary.
+    """
+        Diff. equation describing the orbital evolution of a detached binary.
 
-    The equation handles wind mass-loss [1]_, tidal [2]_, gravational [3]_
+        The equation handles wind mass-loss [1]_, tidal [2]_, gravational [3]_
```
maxbriel, line 914:
```suggestion
    The equation handles wind mass-loss [1]_, tidal [2]_, gravational [3]_
```

posydon/binary_evol/DT/track_match.py
```diff
@@ -0,0 +1,1951 @@
...
+        if grid_name_strippedHe is None:
+            grid_name_strippedHe = os.path.join('single_HeMS', self.metallicity+'_Zsun.h5')
+        self.grid_strippedHe = GRIDInterpolator(os.path.join(path, grid_name_strippedHe))
+
+        # ==================================================================================
+
+        # Initialize the matching lists:
+
+        # min/max ranges of initial masses for each grid
+        m_min_H = np.min(self.grid_Hrich.grid_mass)
```
maxbriel, line 343: Wasn't a way to do this added by @kkovlakas ? Or was that only for binary grids?
It doesn't matter too much.

posydon/binary_evol/binarystar.py
```diff
@@ -364,10 +406,49 @@ def to_df(self, **kwargs):
         pandas DataFrame
 
         """
+
+        def equalize_columns(data_to_save, all_keys, properties_dtypes,
+                             max_col_length):
+            """Iterate through data_to_save to append null values."""
```
maxbriel:
```suggestion
            """Iterate through data_to_save to append null values.
            
            Parameters
            -----------
            data_to_save : list[str]
                Columns to save to the final dataframe
            all_keys : list[str]
                All available keys to store.
           properties_dtypes : dict
               Types of all available keys
           max_col_length: int
               The maximum length of columns.  
            """
```

maxbriel (dismissed):

maxbriel:
Nicely done! This looks really good and is nicely modular.

1. I know this was in the code before, but in `__call__` of `step_detached`, there are 7 functions defined. Is it easy to move these out of the other function?
2. Same with `match_to_single_star` in `track_match.py`, where there are quite a few functions defined inside that one function.
For both, it would be great to separate the functions out.

posydon/binary_evol/DT/step_detached.py
sgossage: Thanks, Max -- removed those now. They're all used in track_matcher.py now.

sgossage:
> Nicely done! This looks really good and is nicely modular.
> 
> 1. I know this was in the code before, but in `__call__` of `step_detached`, there are 7 functions defined. Is it easy to move these out of the other function?
> 2. Same with `match_to_single_star` in `track_match.py`, where there are quite a few functions defined inside that one function.
>    For both, it would be great to separate the functions out.

Thanks, Max.

I've now cleaned out the `__call__` function in `step_detached` and isolated most of the functions. 

I've also cleaned up `match_to_single_star`, which is now its own function. Most of the functions that were nested there are now nested inside of the separate `match_through_minimization` function (b/c most of those subfunctions should only be used by the minimization routine); `match_through_root` is also its own function now.

posydon/binary_evol/DT/track_match.py
sgossage: [PR#271](https://github.com/POSYDON-code/POSYDON/pull/271) added that functionality to `MesaGridStep`. We should consider generalizing that functionality.

maxbriel: Not necessary for now then.

maxbriel (approved):
Thank you for the improvements! This looks good to me now. Eventually, we should move all functions outside of function definitions, because they're impossible to test. But for this PR it is fine to keep it as is. We will probably revisit this code in the future.

Is this a PR to be included in v2? Not yet right?

sgossage:
> Thank you for the improvements! This looks good to me now. Eventually, we should move all functions outside of function definitions, because they're impossible to test. But for this PR it is fine to keep it as is. We will probably revisit this code in the future.
> 
> Is this a PR to be included in v2? Not yet right?

thanks, Max -- that's right, I believe we planned this for v2.1

astroJeff (approved):

PR #512, opened 2025-02-21
Add WDs to infer_star_state

### LINKED ISSUE #366: White Dwarf Termination Flag Label
When plotting a 2D grid of CO-HeMS, we found that the compact object flag wasn't labeling WD type stars. In the file `POSYDON_DIR/posydon/utils/common_functions.py`, the function `infer_star_state` does not have a return case for white dwarfs. 

 We have tested that adding a return case for white dwarfs in this function solves the issue. For example: 
```
def infer_star_state(star_mass=None, surface_h1=None,
                     center_h1=None, center_he4=None, center_c12=None,
                     log_LH=None, log_LHe=None, log_Lnuc=None, star_CO=False):
    """Infer the star state (corresponding to termination flags 2 and 3)."""
    if star_CO:
        if (star_mass <= STATE_NS_STARMASS_UPPER_LIMIT) & (star_mass >= STATE_NS_STARMASS_LOWER_LIMIT):
            return "NS"
        elif star_mass < STATE_NS_STARMASS_LOWER_LIMIT:
            return "WD"
        else:
            return "BH"
 ```
where `STATE_NS_STARMASS_LOWER_LIMIT` is defined in `POSYDON_DIR/posydon/utils/limits_thresholds.py` as `STATE_NS_STARMASS_LOWER_LIMIT=1.37` as in (Fragos et. al 2023). 

We also made a corresponding white dwarf label in `POSYDON_DIR/posydon/visualization/plot_defaults.py`. For example:

```
        'WD':
            ['*', 1, 'tab:blue', 'WD'] 
```
Should this be implemented?

mkruckow:
- [x] add WDs to `infer_star_state`
- [x] add new limits `STATE_NS_STARMASS_LOWER_LIMIT` and `STATE_WD_STARMASS_UPPER_LIMIT`
- [x] adopt unit tests
- [x] tested on small pop-syn

### DIFF
--- a/posydon/unit_tests/utils/test_common_functions.py
+++ b/posydon/unit_tests/utils/test_common_functions.py
@@ -73,7 +73,9 @@ class TestElements:
                     'MT_CASE_UNDETERMINED', 'MT_STR_TO_CASE',\
                     'Pwarn', 'REL_LOG10_BURNING_THRESHOLD',\
                     'RICHNESS_STATES', 'RL_RELATIVE_OVERFLOW_THRESHOLD',\
+                    'STATE_NS_STARMASS_LOWER_LIMIT',\
                     'STATE_NS_STARMASS_UPPER_LIMIT', 'STATE_UNDETERMINED',\
+                    'STATE_WD_STARMASS_UPPER_LIMIT',\
                     'Schwarzschild_Radius', 'THRESHOLD_CENTRAL_ABUNDANCE',\
                     'THRESHOLD_HE_NAKED_ABUNDANCE', '__authors__',\
                     '__builtins__', '__cached__', '__doc__', '__file__',\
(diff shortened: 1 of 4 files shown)


celiotine (dismissed):
population runs ok and resulting NS/WD mass distributions look reasonable, maybe someone else wants to confirm that these are the lower/upper limits we want to use

sgossage:
The thresholds seem reasonable to me -- is there a reference for our choice of 1.48 Msol for the WD upper mass limit? It seems this can vary depending on the analysis

sgossage (dismissed):

mkruckow:
OK, I have added a bunch of references.

sgossage (approved):

astroJeff (approved):

PR #521, opened 2025-02-27
Fix logic flow in matching

### LINKED ISSUE #447: Fix and improve matching logic flow
The logic flow for determining the best matching solution for binaries in the detached step needs improved. Currently, if solution 1 does not pass the strict matching tolerance, the matching tries solution 2 and checks if it passes the strict tolerance, etc. At the end of this process, there is a "relaxed" matching tolerance check that is applied to the _final_ matching solution. This is causing some binaries to fail in the matching that actually have a valid matching solution from earlier solution attempts. Specifically, they have a matching solution that occurred before the final matching attempt that passed the relaxed tolerance check but did not pass the strict tolerance check.

Here is an example binary that could be successfully matched but is failing because of this issue:
```
star_1 = SingleStar(**{'mass':   7.55285829829485, 'state': 'H-rich_Core_H_burning',
                'natal_kick_array': [40.91509926587841, 2.6295454150818256, 1.6718337470964977, 6.0408769315244895]})
star_2 = SingleStar(**{'mass': 6.742063481560266, 'state': 'H-rich_Core_H_burning',
binary_IF = BinaryStar(star_1, star_2, **{'time': 0.0,  'state': 'detached',  'event': 'ZAMS', 
                        'orbital_period':  17.957531550841225, 'eccentricity': 0.0}, properties = sim_prop_IF)
```
run with the .ini file `/projects/b1119/cliotine/dev_general/debugging/params.ini`.

This is the matching output for this binary:
Screen Shot 2024-11-22 at 1 45 39 PM

As you can see, the solution from the 1st alternative matching attempt just exceeds the strict tolerance and would be a very reasonable "best" matching solution for the binary, but because the last matching attempt is checked against the relaxed limit, the matching fails overall.

NOTE: I already tried to implement a solution to this in PR #430, keeping the "best" matching solution in a separate variable and then checking the best solution against the relaxed tolerance at the end. While this fixed the matching issue, I found that it was causing new problems for some binaries related to the ODE solver in the detached step. For this reason, I believe the issue should be more thoroughly investigated in a separate PR. 
In addition, I think we should discuss the purpose of the relaxed tolerance limit. From discussion with @ezapartas , this limit was originally introduced to successfully match the majority of stripped_He stars. If having an increased tolerance limit creates further computational problems for non stripped_He stars (again, this requires further investigation), then perhaps it should not be applied to all binaries.

celiotine:
This PR fixes the algorithmic bug discussed in Issue #447 and also restructures the matching output so that it is easier to read and interpret.
Here is what the binary described in Issue #447 looks like when evolved now with this PR. Previously, it was not being matched due to the bug, but now it has a reasonable match:
Screen Shot 2025-02-27 at 10 23 50 AM

### DIFF
--- a/posydon/binary_evol/DT/step_detached.py
+++ b/posydon/binary_evol/DT/step_detached.py
@@ -755,7 +755,7 @@ class detached_step:
                 sol = minimize(sq_diff_function, x0, method="Powell")
 
                 ## if alternative matching has a better solution, make it the new best solution
-                if np.abs(sol.fun) < np.abs(best_sol.fun):
+                if (np.abs(sol.fun) < np.abs(best_sol.fun) and sol.success):
                     best_sol = sol
 
             # 2nd attempt: use alternative matching parameters
@@ -781,7 +781,7 @@ class detached_step:
                 x0 = get_root0(MESA_labels, posydon_attributes, htrack, rs=rs)
                 sol = minimize(sq_diff_function, x0, method="TNC", bounds=bnds)
 
-                if np.abs(sol.fun) < np.abs(best_sol.fun):
+                if (np.abs(sol.fun) < np.abs(best_sol.fun) and sol.success):
                     best_sol = sol
 
             # 3rd attempt: match an He-star with an H-rich grid, or vice versa (not applicable for HMS stars)
@@ -820,7 +820,7 @@ class detached_step:
 
                     try:
                         sol = minimize(sq_diff_function, x0, method="TNC", bounds=bnds)
-                        if np.abs(sol.fun) < np.abs(best_sol.fun):
+                        if (np.abs(sol.fun) < np.abs(best_sol.fun) and sol.success):
                             best_sol = sol
 
                         if self.verbose:
@@ -838,7 +838,7 @@ class detached_step:
                           np.abs(best_sol.fun), ">", tolerance_matching_integration_hard)
                 initials = (np.nan, np.nan)
                 
-            elif np.abs(best_sol.fun) < tolerance_matching_integration_hard:
+            else:
                 if self.verbose:
                     print("\nFinal matching result is considered successful with best tolerance "
                         f'{np.abs(best_sol.fun):.8f}', "<", tolerance_matching_integration_hard)
--- a/posydon/binary_evol/DT/step_detached.py
+++ b/posydon/binary_evol/DT/step_detached.py
@@ -846,17 +846,18 @@ class detached_step:
 
         if self.verbose:
             if not np.isnan(initials[0]):
-                val_names = ["  ", "mass", "log_R", "he_core_mass", "center_he4", "surface_he4", 
-                             "surface_h1", "center_c12"]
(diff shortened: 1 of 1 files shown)


posydon/binary_evol/DT/step_detached.py
```diff
@@ -738,28 +738,34 @@ def get_MESA_labels(list_for_matching):
...
+                    print (f"Initial matching attempt was unsuccessful:"
+                           f"\n tolerance {np.abs(sol.fun)} > {tolerance_matching_integration}, sol.success = {sol.success}")
+                    
+                    print("\nAlternative matching started (1st attempt)")
                     print("(Now trying an alternative minimization method)")
                     
                 sol = minimize(sq_diff_function, x0, method="Powell")
 
+                ## if alternative matching has a better solution, make it the new best solution
+                if np.abs(sol.fun) < np.abs(best_sol.fun):
```
mkruckow: For the next try we although check on the success, hence should we compare the success evolution here as well? (same in later tries)

posydon/binary_evol/DT/step_detached.py
```diff
@@ -810,48 +820,65 @@ def get_MESA_labels(list_for_matching):
...
+            if (np.abs(best_sol.fun) > tolerance_matching_integration_hard or not best_sol.success):
                 if self.verbose:
-                    print("\nMatching result is NOT successful, with tolerance ",
-                          np.abs(sol.fun), ">", tolerance_matching_integration_hard)
+                    print("\nFinal matching result is NOT successful with best tolerance ",
+                          np.abs(best_sol.fun), ">", tolerance_matching_integration_hard)
                 initials = (np.nan, np.nan)
                 
-            elif np.abs(sol.fun) < tolerance_matching_integration_hard:
+            elif np.abs(best_sol.fun) < tolerance_matching_integration_hard:
```
mkruckow: Why do we have this `elif` here? Hence, do we want to allow a case, where initials isn't set here? I'd suggest to simply turn it into an `else`, otherwise if we need to have three cases, please add the third `else` case.

mkruckow (approved):

astroJeff:
@celiotine this looks good, but there I have a question. It looks like a helium star was matched to a hydrogen star. I am concerned that the "correct" match ought to have been with the last matching step. @mkruckow, can you look into this?

mkruckow:
> @celiotine this looks good, but there I have a question. It looks like a helium star was matched to a hydrogen star. I am concerned that the "correct" match ought to have been with the last matching step. @mkruckow, can you look into this?

Yes, that's the last attempt to find a match in a different grid in case there is no match found in the desired grid. This might be a consequence of complicated evolution. In the case of a He star going to H-grid: this will happen for stars which accreted enough He to change the surface value to classify it as He star, but being so little that there is still a thick H layer below. When the accretor had time to let the chemical composition settle down this star would become a H-rich star nevertheless, hence should match to the H-grid.

sgossage (approved):

PR #529, opened 2025-03-06
Add total_mass_h1 and total_mass_he4

### LINKED ISSUE #497: Include total_mass_h1 in the next post-process round for MS+MS mergers
Summary: Adding the total_mass_h1 property in the stars would be helpful for a better estimate of the MS+MS merger products, following [Schneider+2016](https://arxiv.org/pdf/1601.05084) and [Wang et al. 2022](https://arxiv.org/pdf/2202.05552).

We discussed this here in slack: https://posydon.slack.com/archives/CDCS3MS31/p1738145357918819

- We need to add the parameter in the new post-processed grids (at least in the **final values of binary grids** and in the **total history/EEPs of single stars grids**). This is the first, more urgent step, and it would be great to be queued for the v2 data release : We confirmed that total_mass_h1  is recorded in the histories. We concluded that it is straightforward to implement this but needs post-processing of at least the final values of the binary_grids for total_mass_h1 (so that we have this info in step_CE and step_merged) and for the history of single star grids (so that they are matched with their EEPs, as MS merger products).

- For what is suggested for the MS+MS mergers total_mass_h1 is enough as a property. total_mass_he seems another very reasonable one, and I can already think of a few cases that it will be handful (so thinking ahead, if we post-process the grids, keeping this too would ideally be helpful). But Matthias has already pointed the discussion on grid sizes, so it is a technical decision on whether it is afforded. But it is not important for Dimitris specific question.

- Implement the merging prescription of Schneider et al. would be done by @dimsour94 and @ezapartas and is probably aimed for a version of POSYDON after v2.0. "There they take into account the rejuvanation process. The implementation appears straightforward, as it only requires the average hydrogen mass fraction of both stars before merging and their initial hydrogen mass fraction. Using this information, we can estimate the average hydrogen mass fraction of the merger product and then align it with our single-star model that matches both the mass and average hydrogen mass fraction of the merged product"

mkruckow:
- [x] adding `total_mass_h1` and `total_mass_he4` to the code to be saved from the MESA histories
- [x] make them accessible in the pop-syn
- [x] update docs
- [x] recreate grids (HMS-HMS not yet done)

Other changes:

- Bugfix: exclude `accreted_He_Core_H_burning` from single star HRD (this state was introduced to the plotting in PR #404 )


ezapartas (approved):

sgossage:
I think this looks good, but do we need to complete a pop-syn test before accepting?

sgossage (approved):

mkruckow:
> I think this looks good, but do we need to complete a pop-syn test before accepting?

We don't need, it would have been nice to have a check, that nothing breaks in pop-syn with this PR.

PR #530, opened 2025-03-07
Add Moe & Di Stefano distributions of initial binary properties

mkruckow:
- [x] add code to generate binaries from Moe & Di Stefano distributions
- [x] incorporate into binary initialization
- [x] update `population_params_default.ini`
- [x] update doc
- [x] test on small pop-syn


maxbriel:
@mkruckow I've finally gotten around to this, and I will run a larger popsyn with it. Will report back once finished.

posydon/popsyn/independent_sample.py
```diff
@@ -9,12 +9,16 @@
...
+    "Matthias Kruckow <Matthias.Kruckow@unige.ch>",
 ]
 
 
 import numpy as np
 from scipy.stats import truncnorm
 from posydon.utils.common_functions import rejection_sampler
+from posydon.popsyn.Moes_distributions import Moe2017PsandQs
+
+_gen_Moe2017 = None
```
maxbriel: is it really necessary to have this global parameter?

The `BinaryPopulation` should do its sampling in one call of `generate_independent_samples`. The generator should thus be fixed for each sample taken. Of course, rerunning the `generate_independent_samples` will require it to be rebuild, but is it necessary to store it like this?

posydon/popsyn/population_params_default.ini
```diff
@@ -285,254 +304,266 @@
...
   entropy = None
-  # `None` uses system entropy (recommended)
+    # if None uses system entropy (recommended): None or int
   number_of_binaries = 10
-    # int
-  binary_fraction_scheme =  'const'
-    #'const' 'Moe_17'
+    # int (0, inf)
+  binary_fraction_scheme = 'const'
+    # 'const', 'Moe_17'
```
maxbriel: Can we make the usage of the Moe and Di Stefano consistent?

There's an old version still there. The new `Moe2017` is cleaner. This should be changed in the relevant code too (should be in `independent_sample.py`).

It would be nice to include both authors in the name if possible.

posydon/popsyn/population_params_default.ini
```diff
@@ -285,254 +304,266 @@
...
 
   read_samples_from_file = ''
-    # path to file to read initial parameters from (if empty string get random samples)
+    # path to file to read initial parameters from (if empty string get random
+    #                                               samples)
   primary_mass_scheme = 'Kroupa2001'
     # 'Salpeter', 'Kroupa1993', 'Kroupa2001'
   primary_mass_min = 7.0
-    # float (0,130)
+    # float (0, inf)
```
maxbriel, line 365: We should not allow or suggest that one can set the primary mass to any high value. Our grids have a limit of ~300 Msun. There's no actual check, but we should give a number based on our grids.

Additionally, the current implementation of the underlying mass does not allow our sampling to go above 200 Msun, since that's the limit for the mass limit correction.

posydon/popsyn/independent_sample.py
mkruckow: That's exactly the point, the `BinaryPopulation` does not generate all initial parameters of the full population in one call at the moment. Thus, I save the object to avoid the rebuild, which would currently happen for each simulated binary.

posydon/popsyn/population_params_default.ini
mkruckow: First, the binary fraction scheme doesn't use the new class, instead it uses other code to get just the binary fraction. The old version is needed in case you'd like to not use it for everything but the binary fraction only.

Second, why should we get the second author in here we don't do that for any other distribution, they are all called just by first author.

posydon/popsyn/population_params_default.ini
mkruckow: With this argument we are not allowed to use 0 either. The values stated here are the limits on having a positive value for the mass, that's it. All other limitation have to be captured by the code. I don't think that it is good practice to put limits of the current data hard coded in the example file. Even more the values will differ for v1 vs. v2 grid, hence we'd need to give a value per gird set.
Yes, the sampling and weighting should be able to handle all given mass ranges, but that's outside the scope of this PR. It might be good to take care of it in the PR #493 .

posydon/popsyn/population_params_default.ini
maxbriel: I'm requesting the scheme names to be the same, since they come from the same work. It isn't very clear to have two different notations for the same work.

posydon/popsyn/population_params_default.ini
maxbriel: I would agree that 0 is not allowed either. As such, we should put these limits in there.

posydon/popsyn/Moes_distributions.py
```diff
@@ -0,0 +1,470 @@
...
+        Parameters
+        ----------
+        M1 : float or ndarray of float
+            Mass of the primary in Msun.
+        M_min : float (default: 0.08)
+            Minimum mass of a star in Msun.
+        M_max : float (default: 150.0)
+            Maximum mass of a star in Msun.
+        all_binaries : bool (default: True)
+            If true set binary fration to 1.
```
maxbriel:
```suggestion
            If true, set binary fraction to 1.
```

maxbriel:
The population synthesis ran correctly.
I need some clarification, though:
It looks like currently, only binaries are sampled, because `all_binaries=True`. Is the binary fraction not considered in this sampling? It seems like the mechanics are there to do the calculation to get the binary fraction.

Shouldn't it be essential to have the binary vs single fraction included in the sampling? I know that the code setup is a bit strange for the sampling of single vs binaries.

posydon/popsyn/population_params_default.ini
mkruckow: But it isn't the same even it comes from the same work. Yes, there might be a better way to name them. As an equivalent I'd orient on how we do for the SN prescription from Fryer+12 as we have there although two prescriptions coming from the same work.
Thus I now named them `Moe+17-massdependent` and `Moe+17-PsandQs`

posydon/popsyn/population_params_default.ini
mkruckow: I left the values but added a statement, that the values should be chosen according to the used grids.

mkruckow:
> The population synthesis ran correctly. I need some clarification, though: It looks like currently, only binaries are sampled, because `all_binaries=True`. Is the binary fraction not considered in this sampling? It seems like the mechanics are there to do the calculation to get the binary fraction.
> 
> Shouldn't it be essential to have the binary vs single fraction included in the sampling? I know that the code setup is a bit strange for the sampling of single vs binaries.

I have created issue #586 .

maxbriel (dismissed):

maxbriel (dismissed):

astroJeff:
@kasdaglie can you take a look at this PR and review it? The only thing it doesn't do is include a variable binary fraction, see issue #586 .

maxbriel (approved):

astroJeff:
This PR is superseded by PR #611

Issue #463, opened 2024-12-05
Consistent dunder main usage


ka-rocha:
As discussed in #436, the use of dunder main for command line calls to python files is not consistent across the codebase. The changes added in #436 uses the function protocol while other instances may put the check for dunder main at some unspecified location within the file above the code intended to be executed.

Since the functional use is recommended in the [python documentation](https://docs.python.org/3/library/__main__.html#idiomatic-usage) with rational described there, this issue is to the transition all dunder main calls to match for logical consistency (e.g. `bin/posydon-run-grid`).

astroJeff:
@ka-rocha can you highlight a couple places in the code where you see issues? I would like a better understanding of this issue.

mkruckow:
I think, that's mainly an issue for our code in the `bin` directory. Here, I'd even go a step further and try to outsource all the functions to the `posydon/utils` (this makes it although easier to write unit tests for those parts.
@ka-rocha , please let us know if you have other files in mind which are not located in `bin`.

maxbriel:
Based on PR #436, I think @ka-rocha is referring to the usage of `if __name__ == "__main__"` inside sub-modules instead of having them separated into separate files, such as the `bin/` scripts.

Moving the code out of the `bin` to `utils` is still a good idea, but not the point of this issue. We can keep this open or move them to a separate new issue :)

I searched the codebase and found some additional files. All contained to the `active_learning` module:
- [posydon.active_learning.psy_cris.synth_data_3D](https://github.com/POSYDON-code/POSYDON/blob/d63207dd87acf636681c20db2d4f2785799c0ed0/posydon/active_learning/psy_cris/synthetic_data/synth_data_3D.py#L131-L132)
- [posydon.active_learning.psy_cris.run_params.run_psycris_sequence](https://github.com/POSYDON-code/POSYDON/blob/d63207dd87acf636681c20db2d4f2785799c0ed0/posydon/active_learning/psy_cris/run_params/run_psycris_sequence.py#L140-L141)
- [posydon.active_learning.psy_cris.synth_data_2D](https://github.com/POSYDON-code/POSYDON/blob/d63207dd87acf636681c20db2d4f2785799c0ed0/posydon/active_learning/psy_cris/synthetic_data/synth_data_2D.py#L46-L47)
- old tests from the `active_learning` sub-module.

There hasn't been any active development on this sub-module, so I'm not sure how we should proceed with these sections. @astroJeff and @mkruckow What would you recommend?

PR #533, opened 2025-03-10
Add rescue script

asb5468:
This script is designed to rescue popsyn job arrays generated with `posydon-setup-popsyn` where select jobs in the array fail (it's only useful if the failures are due to transient filesystem issues rather than actual issues with the code). It omits entire redshifts if they have completed successfully and can be run multiple times until the full job array is complete for each redshift. The input argument is the directory where the original run scripts for the job arrays were created (assumed to be `./` if no `--rundir` is passed). It's set up to run as a command line tool, and I've tested that it executes the expected functionality in all possible combinations of arguments passed and multiple iterations of rescue.


sgossage:
Sylvia had population synthesis jobs run on Quest that experienced this issue:

```
srun: error: load_plugins: unable to find plugin: serializer/json
srun: error: init: unable to load JSON serializer: Plugin has invalid format or unable to be loaded
srun: error: Couldn't load specified plugin name for cli_filter/lua: Plugin init() callback failed
srun: error: cannot create cli_filter context for cli_filter/lua
srun: fatal: failed to initialize cli_filter plugin
```

Which appears to be a transient file system error as mentioned in the PR description.

bin/posydon-rescue-popsyn
```diff
@@ -0,0 +1,54 @@
+#!/usr/bin/env python
+# script to rescue failed popsyn job arrays
+# #Author: Sylvia Biscoveanu
+import argparse 
+import glob
+import os
+
+def main(rundir):
```
sgossage:
```suggestion
def rescue_runs(rundir):
```

bin/posydon-rescue-popsyn
```diff
@@ -0,0 +1,54 @@
...
+        else:
+            new_line = line.replace('array.slurm', f'array.slurm.rescue{rescue_number}')
+            master_rescue.write(new_line)
+    print(f'Now run:\n\tcd {rundir}; bash {master_rescue_filename}')
+                
+if __name__=='__main__':
+    parser = argparse.ArgumentParser()
+    parser.add_argument('--rundir', default='./', help='directory where the jobs originally ran')
+    args = parser.parse_args()
+    main(**vars(args))
```
sgossage:
```suggestion
    rescue_runs(**vars(args))
```

sgossage:
Thanks Sylvia, please use `os.path.join()` to construct paths where possible and add a docstring for users. See examples in the other scripts located in `POSYDON/bin/`.

mkruckow:
Thanks for putting something together @asb5468 .
First, do we know, what caused to issue on Quest, it looks like @ZepeiX gets similar issues with his runs. Thus, it would be better to check how we can avoid the error to happen in the first place.
Please orient yourself on [bin/get-posydon-data](https://github.com/POSYDON-code/POSYDON/pull/507/files#diff-f1cc0374fedaafabf8760de43197dc1fdebbafc7a79a924a8ab832b0274dd8e0) for the style of the bin files, this will become the new standard as soon as I get time to work on issue #463 .

asb5468:
@mkruckow I agree it would be good to understand the underlying slurm issue, but I can imagine that something like this is useful in multiple generic cases (i.e., some jobs in the array time out or hit OOM issues while others do not).

To make sure I understand your comment about [bin/get-posydon-data](https://github.com/POSYDON-code/POSYDON/pull/507/files#diff-f1cc0374fedaafabf8760de43197dc1fdebbafc7a79a924a8ab832b0274dd8e0), is the request that I move the function to posydon.utils and then just import the function and run it in the script that goes under `bin/`?

mkruckow:
@asb5468 , yes we might want to have it more generic, there we'd need to see, whether we need to be more flexible, e.g. for issues running out of memory, you'd need to be able to change those things as well in the rescuing script.
Additionally, we where discussing, whether we should include it in the posydon-setup-popsyn as a sub command there.

P.S.: Yes, you got my comment about outsourcing the code to a module file correctly. But this might depend on the answer, whether we will keep it separate or include it into posydon-setup-popsyn.

asb5468:
@sgossage For a docstring, do you mean within the function itself or at the beginning of the script? The various files in `/bin/` seem to do this differently.

asb5468:
It turns out that some jobs ran out of time in my last rescue attempt, so I've edited this script to allow the time and mem-per-cpu to be edited in the slurm submit files too, passed as optional arguments to the script.

sgossage:
Just an update after following up with the IT team. They identified an issue on a particular node that they're working to resolve now and recommended excluding the particular node in the meantime via putting this in the SLURM submission script:

`#SBATCH --exclude=NODE_NAME`

Using the name of the offending node in place of `NODE_NAME`. Would it be possible to adapt the script to identify failed runs, get the node they ran on, and add this line to a resubmission script with the identified node?

sgossage:
> @sgossage For a docstring, do you mean within the function itself or at the beginning of the script? The various files in `/bin/` seem to do this differently.

Here I was thinking a docstring within the function itself.

bin/posydon-rescue-popsyn
```diff
@@ -32,6 +33,26 @@ def rescue_runs(rundir):
                             missing.append(run)
                     missing_str = ','.join(map(str, missing))
                     new_sub_file.write(f'#SBATCH --array={missing_str}\n')
+                elif ('--time' in line) and (time is not None):
+                    split_times = time.split(':')
+                    bad_time_format = False
+                    if len(split_times) != 3:
```
mkruckow, line 39: We should allow all SLURM time formats, see https://slurm.schedmd.com/sbatch.html#OPT_time

bin/posydon-rescue-popsyn
```diff
@@ -32,6 +33,26 @@ def rescue_runs(rundir):
...
+                    elif not all([st.isdigit() for st in split_times]):
+                        bad_time_format = True
+                    if bad_time_format:
+                        os.remove(new_file_name)
+                        raise ValueError('--time must be in the format HH:MM:SS, ' 
+                                         f'instead {time} was passed!')
+                    else:
+                        new_sub_file.write(f'#SBATCH --time={time}\n')
+                elif ('--mem-per-cpu' in line) and (mem_per_cpu is not None):
+                    if not mem_per_cpu.split('G')[0].isdigit():
```
mkruckow, line 50: Again, we shouldn't be too restrictive here.

mkruckow:
> Using the name of the offending node in place of `NODE_NAME`. Would it be possible to adapt the script to identify failed runs, get the node they ran on, and add this line to a resubmission script with the identified node?

You can get the nodes a job run on with `sacct --format NodeList --jobs={SLURMID}`, where you replace the SLURMID with the ID you can get from the log files.

asb5468:
@maxbriel pointed out on slack that this script does not correctly accomodate the way that posydon splits up the jobs between the elements of the array, which I'm not familiar with, so this MR should probably be closed. But it would still be really helpful to have something like this available that comes packaged with posydon, because my understanding is that currently the only way to recover from a partial run failure is to redo the full population run.

asb5468:
This may actually be salvageable using the SLURM_ARRAY_TASK_COUNT and  SLURM_ARRAY_TASK_MIN environment variables, as suggested by Max. I've taken a stab at editing the script to add those in, but someone with more experience should definitely take a look. I don't have the bandwidth to broaden the logic to accept other formats for the time and memory, but the branch settings are such that any of the folks that can review the MR can push to the branch.

bin/posydon-rescue-popsyn
```diff
@@ -53,6 +53,10 @@ def rescue_runs(rundir, time, mem_per_cpu):
                                          f'instead {mem_per_cpu} was passed!')
                     else:
                         new_sub_file.write(f'#SBATCH --mem-per-cpu={mem_per_cpu}\n')
+                elif 'PATH_TO_POSYDON_DATA' in line:
+                    new_sub_file.write(line)
+                    new_sub_file.write(f'export SLURM_ARRAY_TASK_COUNT={len(inds)}\n')
+                    new_sub_file.write(f'export SLURM_ARRAY_TASK_MIN={min(inds)}\n')
```
mkruckow, line 59: This will only work as long as the original submission script didn't get modified. So as long as you resubmit failed jobs only through this script.

mkruckow: I think we'd need to test whether SLURM has any problems with modifying these variables, e.g. in case it uses them to dynamically redetermine the number of jobs. This might depend on the SLURM installation settings.

maxbriel: Back when I encountered this issue, I did some testing on it and found that it was set when the job was launched. Altering them in here is fine, albeit not elegant.

maxbriel:
@sgossage @asb5468 I'm doing a lot of popsynth runs at the moment, and I need something similar to this with some extra functionality. So I will pick this up and work it into the other bin script.

maxbriel:
@asb5468 I've created PR #550 to integrate this within the `posydon-popsyn` command.
This does not keep track of the rescue attempt at the moment, is that crucial for your use case?

mkruckow:
This PR is replaced by PR #550 , please check it out! If there are important things missing there, please comment on PR #550 .

PR #539, opened 2025-03-13
Add files to installation for conda installation/packaging

maxbriel:
New release for v2.0.0-pre2 to fix missing files in the conda installation

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,3 +1,18 @@
+# add the main README and this file
 include README.md MANIFEST.in
+
+# add the versioneer
 include versioneer.py
 include posydon/_version.py
+
+# add additional README files
+recursive-include posydon README.md
+
+# add default ini files
+recursive-include posydon *.ini
+
+# add matplotlib style file
+include posydon/visualization/posydon.mplstyle
+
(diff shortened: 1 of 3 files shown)


posydon/unit_tests/__init__.py
mkruckow, line 1: There are no `unit_tests` on `main`, hence I'd not include the creation of this directory and its subdirectories in this PR.

astroJeff: I agree with @mkruckow on this. Can we leave these out for now, or is there some other reason to keep them in?

sgossage: @maxbriel everything looks good to me outside of making the decision here

maxbriel: removed :)

mkruckow (approved):
Looks good for v2.0.0-pre2. What is the status of updating version 1.0.4, as this would need to be done first in case we do it?

maxbriel:
I've published a release of v1.0.5, which I verified before publishing that the conda package will contain the `.ini` files in the tarball. The new v1.0.5 on conda contains the `.ini` files too.

astroJeff (approved):

PR #542, opened 2025-03-17
Upgrade setuptools

### LINKED ISSUE #373: Deprecation warning on POSYDON editable install
While creating a new install of POSYDON using `pip install -e .`. I received the following deprecation warning for a future pip version about this type of editable installation. Since this is the way we recommend developers to install POSYDON, we should check what the new method is.

This should not affect normal installations that are not editable.

```
DEPRECATION: Legacy editable install of posydon==1.0.0+255.g2093203 from file:///srv/astro/projects/posydon/software/POSYDON_versions/POSYDON_2093203 (setup.py develop) is deprecated.
pip 25.0 will enforce this behaviour change.
A possible replacement is to add a pyproject.toml or enable --use-pep517, and use setuptools >= 64.
If the resulting installation is not behaving as expected, try using --config-settings editable_mode=compat.
Please consult the setuptools documentation for more information.
Discussion can be found at https://github.com/pypa/pip/issues/11457
```

mkruckow:
The new standard [PEP 621](https://peps.python.org/pep-0621) requires a different file than `setup.py`.

- [x] create `pyproject.toml`
  - fill info
  - try to load as much as possible dynamically from `setup.py` to be backward compatible
- [x] some small cleanup in `setup.py`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,3 +1,6 @@
+# following PEP 621: https://peps.python.org/pep-0621
+# details: https://packaging.python.org/en/latest/guides/writing-pyproject-toml
+
 [build-system]
 requires = ["setuptools >= 76.0.0", "versioneer"]
 build-backend = "setuptools.build_meta"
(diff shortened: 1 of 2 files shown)


sgossage (approved):

maxbriel (approved):
Thanks for doing this one!

Issue #314, opened 2024-05-29
Checks when steps get initialized


mkruckow:
I recently experienced a case that step SN always failed, in the example it was caused by having no SN model which matches.
But instead of getting told about this at the beginning of the pop-syn, it run all through letting all systems entering the SN step fail there. Even more striking all systems which didn't get to this step evolved correctly, hence a quick check on a system, whether the run went well gave the wrong impression.

Hence, I'd suggest that we should add checks in the initialization of all steps to tell the user at this point if a step will always fail. I guess, this will be like the adding of the error and warning classes, that different people will need to do this for different steps.

mkruckow:
Maybe it would even be better to check that already in our [posydon-setup-popsyn](https://github.com/POSYDON-code/POSYDON/blob/development/bin/posydon-setup-popsyn).
One more thing to check early: is the POSYDON data exists or need to be downloaded. Especially, when running pop-syn in an array, it is bad if each job tries to download the data separately.

mkruckow:
For record:
- In PR #417 , is no check included to ensure valid values of `PPI_extra_mass_loss`, which should be taken care of here.
- Additionally, we should think about whether we like to allow the following case, may causing PPISN cases not labeled as such:
```
conserve_hydrogen_envelope = False
conserve_hydrogen_PPI  = True
```

astroJeff:
I don't think it's reasonable to have a check for every possible combination of parameters, but we can have some basic checks to ensure parameters are reasonably set. I am thinking this ought to be run by `PopulationRunner` in `synthetic_population.py`

mkruckow:
Sure, it's not about getting checks for all possible combinations, but it would be good to do checks for all combinations which will otherwise raise errors later down the line.

PR #545, opened 2025-03-18
Add check for valid MODELS in posydon-setup-popsyn

maxbriel:
Adds a basic check to make sure the parameters defined in the ini file for step_SN is a valid MODEL.

It only checks against the MODELS if `use_interp_values=True`.

### DIFF
--- a/bin/posydon-setup-popsyn
+++ b/bin/posydon-setup-popsyn
@@ -10,6 +10,7 @@
 import os
 import argparse
 from posydon.popsyn.io import (binarypop_kwargs_from_ini, 
+                               simprop_kwargs_from_ini,
                                create_run_script_text,
                                create_merge_script_text)
 from posydon.config import PATH_TO_POSYDON, PATH_TO_POSYDON_DATA
@@ -17,6 +18,7 @@ from posydon.popsyn.binarypopulation import BinaryPopulation
 from posydon.popsyn.io import binarypop_kwargs_from_ini
 from posydon.utils.common_functions import convert_metallicity_to_string
 from posydon.utils.posydonwarning import Pwarn
+from posydon.grids.MODELS import MODELS
 
 
 def create_run_script(ini_file):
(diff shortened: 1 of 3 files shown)


bin/posydon-setup-popsyn
```diff
@@ -113,7 +115,45 @@ srun python ./merge_metallicity.py {metallicity}
...
+        tmp = MODEL_NAME
+        for key, val in MODEL.items():
+            if "use_" in key or key=="ECSN":
+                # escape values, which are allowed to differ
+                continue
+            if step_SN_MODEL[key] != val:
+                tmp = None
+                break
+        if tmp is not None:
+            MODEL_NAME_SEL = tmp
```
mkruckow: We shouldn't duplicate code, it would be better to put this into a function in the utils and call the same function here and in step_SN, where it was copied from.

mkruckow:
It would be good to get those kind of checks for all the parameters in the ini file see issue #314 .

bin/posydon-setup-popsyn
```diff
@@ -113,7 +115,45 @@ srun python ./merge_metallicity.py {metallicity}
...
+    step_SN_MODEL = simprop_kwargs['step_SN'][1]
+    # always allow the use of non-interpolation values
+    if step_SN_MODEL['use_interp_values'] == False:
+        return True
+    # step_SN MODEL check
+    MODEL_NAME_SEL = None
+    for MODEL_NAME, MODEL in MODELS.items():
+        tmp = MODEL_NAME
+        for key, val in MODEL.items():
+            if "use_" in key or key=="ECSN":
```
sgossage: do we also need to escape `conserve_hydrogen_envelope`, `max_neutrino_mass_loss`, `max_NS_mass`, `approx_at_he_depletion`?

mkruckow: We can't exclude stuff, which goes in the calculation of the CO properties, e.g. mass or spin, which need the full profile to get for BHs. Thus, we can't exclude without having a full integrated and working profile interpolation:
- `mechanism`
- `engine`
- `PISN`
- `PISN_CO_shift`
- `PPI_extra_mass_loss`
- `conserve_hydrogen_envelope`
- `conserve_hydrogen_PPI`
- `max_neutrino_mass_loss`
- `max_NS_mass`
- `allow_spin_None`
- `approx_at_he_depletion`

The only reason, why we can exclude `ECSN` is that it only influences WDs and NSs, and for both formations we don't use the profile, see PR #504 .

bin/posydon-setup-popsyn
maxbriel: I've removed the duplicate code and created a function in MODELS.py, since that's the only place it uses information from.

bin/posydon-setup-popsyn
maxbriel: Technically at the moment, `PPI_extra_mass_loss` and `PISN_CO_shift` can be excluded, because they're not set in the MODELS. We only check against the parameters inside each MODEL against those set in step_SN.

posydon/grids/MODELS.py
```diff
@@ -176,3 +176,36 @@
...
 }
+
+def get_MODEL_NAME(input_MODEL, verbose=False):
+    '''Find MODEL_NAME that matches input_MODEL.
+    
+    Parameters
+    ----------
+    input_MODEL : dict
+        Dictionary with the properties of the model.
+
```
mkruckow:
```suggestion
    verbose : bool (default: False)
        Enables additional output.
          
```

bin/posydon-setup-popsyn
mkruckow: Sure, but shouldn't we include them in the MODELS?

mkruckow (approved):
Looks as a good first step of checking the ini file on setup.

bin/posydon-setup-popsyn
maxbriel: If we have trained models that require them, yes. But at the moment we don't have any.

sgossage (approved):

PR #546, opened 2025-03-19
Enforce `step_SN` inputs (e.g., `engine`) are strings

sgossage:
A user experienced long binary population synthesis run times due to an improper input in their `.ini` file. This modifies the supplied default `population_params_default.ini` file to specify that the `engine` property of `step_SN` should be an empty string rather than None, as setting it as None can cause binary evolution to hang.

This also places a check in `step_SN.py` to set `engine` to an empty string in case it is set as None. 

A number of other properties in the `.ini` file allow None as input. We might consider ensuring the inputs accept consistent formats. Should we include additional changes in this PR along those lines?

### DIFF
--- a/posydon/binary_evol/SN/step_SN.py
+++ b/posydon/binary_evol/SN/step_SN.py
@@ -262,6 +262,12 @@ class StepSN(object):
                     raise ValueError(key + " is not a valid parameter name!")
             for varname in MODEL:
                 default_value = MODEL[varname]
+
+                # Enforce that `engine` is an empty string if not properly set
+                if varname == "engine":
+                    if not isinstance(kwargs[varname], str):
+                        kwargs[varname] = ""
+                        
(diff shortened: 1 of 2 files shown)


posydon/binary_evol/SN/step_SN.py
```diff
@@ -262,6 +262,12 @@ def __init__(self, **kwargs):
                     raise ValueError(key + " is not a valid parameter name!")
             for varname in MODEL:
                 default_value = MODEL[varname]
+
+                # Enforce that `engine` is an empty string if not properly set
+                if varname == "engine":
+                    if not isinstance(kwargs[varname], str):
+                        kwargs[varname] = ""
+
```
mkruckow, line 270: With PR #545 , this replacement is happening to late and would need to happen on the loading of the parameters from the ini file.
Beside that, I'd suggest to not code one case, but do it for all the parameters to check their types.

PR #554, opened 2025-03-27
POSYDON docs build multiple version + link them

maxbriel:
Requires PR #579 and Website PR [#57](https://github.com/POSYDON-code/POSYDON-code.github.io/pull/57)

This reshuffles the documentation building/creation:
1. It recreates the documentation **on a release** to automatically include the new release as links in the older documentation, i.e. the version selection in the current v1.0.4 documentation. The version selector is now automatically added to any version of the documentation.
2. It adds the different v2 pre-release documentation versions.
3. It adds a specific documentation version for the `development` version.
4. It removes an old unused "Deploy Github Page unstable"

To achieve this, I have to move some stuff around:
1. A new workflow that creates v1 and v2 documentation on release.
2. Move the ghpages actions to conserve the history of the `gh_pages` branch instead of overwriting it each time.

I've only partially been able to test the workflow. I can run the commands locally and was able to run parts as a local github action, but was unable to test the push to the website. 
This will require some changes to the main POSYDON website to link to the correct versions again (for the "latest/development" specifically). 

**This version will push to the gh-pages branch once accepted.**


sgossage:
Tested and all steps appear to complete successfully when deploying the v1 and v2 pages.

Is this ready to be reviewed or still things to be done, Max?

maxbriel:
@sgossage It's now ready to be reviewed. I made sure to remove the new v2.0.0-dev tag from v2 versions creation. 
Additionally, I've removed the dry-run statement.

The main website links will need to be adjusted and I'm not sure how the `history:true` will work with the file currently already present on the `gh-pages` branch. 
We might need to do a manual build to create the right folders & structure, and then these CI's will update it accordingly. Maybe we can discuss this during the dev meeting on Thursday?

.github/workflows/deploy-github-pages-development.yml
```diff
@@ -0,0 +1,82 @@
...
+
+        # install the dependencies in python
+        python -m pip install pandoc
+        python -m pip install coverage cpp-coveralls flake8 pytest
+
+    - name: Make Documentation
+      env:
+        PATH_TO_POSYDON: ./
+        PATH_TO_POSYDON_DATA: ./
+        MESA_DIR: ./
```
mkruckow: Should we remove those?

- `PATH_TO_POSYDON` is reset at the beginning of the run
- `PATH_TO_POSYDON_DATA` and `MESA_DIR` shouldn't be needed for the documentation

(same in the release deployment)

.github/workflows/deploy-github-pages-release.yml
```diff
@@ -0,0 +1,287 @@
...
+        # REQUIRED FOR OLDER VERSIONS OF POSYDON BEFORE THE INCLUSION OF 
+        # THe _templates folder and versioning code.
+        # get the _templates folder and move it out of the POSYDON folder
+        cp -r $PATH_TO_POSYDON/docs/_source/_templates $GITHUB_WORKSPACE
+
+        # get the injection code from the release
+        injection_code=$(awk '/# INJECTION GRAB START/,/# INJECTION GRAB END/' "${PATH_TO_POSYDON}/docs/_source/conf.py")
+
+        # get the available tags
+        TAGS=$(git tag -l | sort -V | grep "v2" | grep -v "v2.0.0-dev")
```
mkruckow: I'd suggest to be more general to exclude all development tags (may apply the same to all versions even there isn't a development tag):
```suggestion
        TAGS=$(git tag -l | sort -V | grep "v2" | grep -v "-dev")
```

.github/workflows/deploy-github-pages-release.yml
```diff
@@ -0,0 +1,287 @@
...
+          mv _build/html/ $output_folder/$tag_without_v
+
+          # cleanup for the next build
+          make clean
+          cd ../
+        done
+        
+    - name: Deploy to GitHub Pages
+      if: success()
+      uses: crazy-max/ghaction-github-pages@v2
```
mkruckow: Should we update this version to v4 as we did for the other scripts?
(I know that not the intent of this PR, but might be a good addition)

.github/workflows/deploy-github-pages-release.yml
```diff
@@ -0,0 +1,287 @@
...
+        
+        if [ -z "$VERSIONS" ]; then
+          echo "No version directories found!"
+          exit 1
+        fi
+        
+        echo "Found versions: $VERSIONS"
+        
+        # Sort versions and get the latest
+        LATEST=$(echo "$VERSIONS" | sort -V | tail -n 1)
```
mkruckow, line 261: Unfortunately, this will sort `v2.0.0` before `v2.0.0-pre1`.

.github/workflows/deploy-github-pages-development.yml
```diff
@@ -0,0 +1,82 @@
...
+        git clean -fdx
+        python -m pip install .[doc,vis]
+      
+        # build the documentation
+        cd docs && make html
+
+        # move and copy docs out of the POSYDON folder
+        touch _build/html/.nojekyll
+        cd ../
+        mv docs/_build/html/ _build/development
```
sgossage:
```suggestion
        touch _build/html/.nojekyll
        cd ../
        mkdir -p _build/development
        mv docs/_build/html/ _build/development
```

I previously tested `deply-github-pages-release.yml` (which works fine), but hadn't tested `deploy-github-pages-development.yml`. It looks like there's an error here, with a fix as suggested above. The workflow fails because the directory _build/development that it tries to copy to doesn't exist, so the suggestion is to create it prior to `cp`.

mkruckow: But that's a move not a copy statement, thus you don't need it to exist.
I guess, what you mean is that we should ensure, the that the parent directory exists:
`mkdir -p _build`

.github/workflows/deploy-github-pages-development.yml
maxbriel: The PATH_TO_POSYDON_DATA and MESA_DIR are required to be able to load in modules that rely on it. POSYDON needs to be installed and autodoc loads each file to load in the docstring. If these are missing, it fails to import it, that's after the `posydon.config` changes :) 

PATH_TO_POSYDON doesn't need to be set here.

.github/workflows/deploy-github-pages-release.yml
maxbriel: I did it for the other one. Forgot for this file :)

.github/workflows/deploy-github-pages-release.yml
maxbriel: I've added a `grep -v 'dev' | grep -v 'pre'` to exclude v2.0.0-pre1 and any -dev version that might be added. (the `-dev`) is just in case.
However, this will sort only full releases and not pre-releases as the latest...

.github/workflows/deploy-github-pages-release.yml
```diff
@@ -0,0 +1,292 @@
...
+        
+        echo "Found versions: $VERSIONS"
+        
+        # First try to find the latest non-pre-release version
+        LATEST=$(echo "$VERSIONS" | grep -v "-pre" | sort -V | tail -n 1)
+        
+        # If no non-pre-release version is found, use the latest including pre-releases
+        if [ -z "$LATEST" ]; then
+          LATEST=$(echo "$VERSIONS" | sort -V | tail -n 1)
+        fi
```
mkruckow: This won't work as:
- grep doesn't like the dash (either leave out the dash or use the `-e` option)
- you will get v1.0.5 now as latest as you only allow the pre if no other version exists.

```suggestion
        # First try to find the overall latest
        LATEST=$(echo "$VERSIONS" | sort -V | tail -n 1)
        
        # Second try to find the latest non-pre-release version
        LATESTNOPRE=$(echo "$VERSIONS" | grep -v "pre" | sort -V | tail -n 1)
        
        # Third get the main version part of $LATEST
        LATESTWITHOUTPRE=$(echo "$LATEST" | awk -F "-pre" '{print $1}')
        
        # If there is a version release after the last pre-release version, use the non pre-release
        if [[ "$LATESTNOPRE" == "$LATESTWITHOUTPRE" ]]; then
          LATEST=$LATESTNOPRE
        fi
```

maxbriel: Thank you for catching these and the commit suggestion!
This was a mess to get working and test. So this is great!

.github/workflows/deploy-github-pages-release.yml
```diff
@@ -0,0 +1,298 @@
...
+        # REQUIRED FOR OLDER VERSIONS OF POSYDON BEFORE THE INCLUSION OF 
+        # THe _templates folder and versioning code.
+        # get the _templates folder and move it out of the POSYDON folder
+        cp -r $PATH_TO_POSYDON/docs/_source/_templates $GITHUB_WORKSPACE
+
+        # get the injection code from the release
+        injection_code=$(awk '/# INJECTION GRAB START/,/# INJECTION GRAB END/' "${PATH_TO_POSYDON}/docs/_source/conf.py")
+
+        # get the available tags
+        TAGS=$(git tag -l | sort -V | grep "v2" | grep -v "-dev")
```
sgossage:
```suggestion
        TAGS=$(git tag -l | sort -V | grep "v2" | grep -v "dev")
```
Using `"-dev"` throws an error for me when testing with `act`:

```
grep: invalid argument ‘ev’ for ‘--directories’
```
where grep interprets `"-dev"` as `"--directories ev"`. Using just `"dev"` works, as long as you think that's specific enough for the code's needs.

mkruckow: Yes, the same as I pointed out at the other place either you need the `-e` option of `grep` or leave out/mask the dash.

sgossage (dismissed):

mkruckow (dismissed):

maxbriel:
I've removed the dry-run argument to make this version work live once accepted.

mkruckow (approved):

elizabethteng (approved):

sgossage (approved):

PR #513, opened 2025-02-25
update docs: installation, processing pipeline, ML components

elizabethteng:
- [x] processing pipeline
    - [x] psygrid object
        - [x] review language
        - [x] test code
    - [x] post-processing pipeline
        - [x] pipeline steps
            - [x] review language
            - [x] test code
        - [x] pipeline additions
            - [x] review language
            - [x] test code 
    - [x] documentation of ini file
        - [x] review language
        - [x] test code
- [x] machine learning components
    - [x] initial final classification and interpolation
        - [x] review language
        - [x] test code
    - [x] profile interpolation
        - [x] review language
        - [x] test code

PR #558, opened 2025-03-31
Rename step9 to stepF

mkruckow:
To overcome the question, why there is a jump from step 4/5 to step 9, I renamed step9 to stepF as suggested by @elizabethteng see PR #513 .


elizabethteng (approved):
looks good, thanks for taking care of this!

sgossage (approved):

PR #570, opened 2025-04-09
Core radius redefine 

ZepeiX:
The code changes the core radius used to evaluate whether CE is successful by comparing it to the post-CE Roche-lobe radius (only for core_not_replaced_stableMT). 

The core radius is assigned as he_core_radius, not r_core_CE_1/10/30cent.

### DIFF
--- a/posydon/binary_evol/CE/step_CEE.py
+++ b/posydon/binary_evol/CE/step_CEE.py
@@ -490,6 +490,7 @@ class StepCEE(object):
         state1_i = donor.state
         m2_i = comp_star.mass
         radius2 = 10**comp_star.log_R
+        he_core_radius1 = donor.he_core_radius
 
         if verbose:
             print("Pre CE")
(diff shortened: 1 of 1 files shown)


mkruckow:
I think we should move the full determination of the final masses and cores before the merging check. Even before the RL calculation to cover possible corrections on mass (I don't know, why there are no corrections for the radius).

astro-abhishek:
A request to keep this branch rather than deleting it after the PR is merged.

astroJeff:
@mkruckow I hear your concern, but I'm looking at the code and to fix it as you suggest requires rewriting CEE_simple_alpha_prescription. This is something I desperately want to see, but we have an immediate need for this fix in the short term.

astroJeff (approved):

mkruckow:
This PR will get superseded by PR #574 .

astroJeff:
As @mkruckow pointed out above, this PR is superseded by PR #574. Closing.

PR #580, opened 2025-04-29
Check for array-like data in `BinaryStar.to_df()` in case of failed binaries

sgossage:
This adds functionality to `binarystar.py`'s `BinaryStar` class function `to_df` to check for array like data types in data columns and pad them appropriately in the case of a failed binary.

A user reported the following error previously, which the fix above is meant to address:

```
Error during breakdown of BinaryStar(ERR, FAILED, p=0.74, S1=(H-rich_Shell_H_burning,M=0.19), S2=(WD,M=1.50)):
'float' object is not subscriptable
  File "/projects/b1095/sbisco/posydon_runs/bbh_test2/./run_metallicity.py", line 14, in <module>
    synpop.evolve()
  File "/home/asb9716/.conda/envs/posydon_dev/lib/python3.11/site-packages/posydon/popsyn/binarypopulation.py", line 258, in evolve
    self._safe_evolve(**self.kwargs)
  File "/home/asb9716/.conda/envs/posydon_dev/lib/python3.11/site-packages/posydon/popsyn/binarypopulation.py", line 418, in _safe_evolve
    self.manager.save(path, mode="w", **kw)
  File "/home/asb9716/.conda/envs/posydon_dev/lib/python3.11/site-packages/posydon/popsyn/binarypopulation.py", line 830, in save
    history_df = self.to_df(**kwargs)
                 ^^^^^^^^^^^^^^^^^^^^
  File "/home/asb9716/.conda/envs/posydon_dev/lib/python3.11/site-packages/posydon/popsyn/binarypopulation.py", line 687, in to_df
    holder.append(binary.to_df(**kwargs))
                  ^^^^^^^^^^^^^^^^^^^^^^
  File "/home/asb9716/.conda/envs/posydon_dev/lib/python3.11/site-packages/posydon/binary_evol/binarystar.py", line 421, in to_df
    V_sys_x[i] = bin_df.iloc[i]['V_sys'][0]
                 ~~~~~~~~~~~~~~~~~~~~~~~^^^
TypeError: 'float' object is not subscriptable
```

The error occurs when the code attempts to subscript `V_sys` in this case, which apparently exists as a `float`. Prior to this PR, data columns were padded with `np.nan` only, which is inappropriate for variables like `V_sys`, which are arrays. Attempting to index `np.nan` can lead to the above error. This pads array-like data with arrays comprised of `np.nan` (or their appropriate data type as determined from `popsyn/io.py`), rather than just `np.nan`.

### DIFF
--- a/posydon/binary_evol/binarystar.py
+++ b/posydon/binary_evol/binarystar.py
@@ -23,6 +23,7 @@ __authors__ = [
     "Philipp Moura Srivastava <philipp.msrivastava@gmail.com>",
     "Devina Misra <devina.misra@unige.ch>",
     "Scott Coughlin <scottcoughlin2014@u.northwestern.edu>",
+    "Seth Gossage <seth.gossage@northwestern.edu>"
 ]
 
 
@@ -391,7 +392,12 @@ class BinaryStar:
             all_equal_length_cols = len(set(col_lengths)) == 1
             if not all_equal_length_cols:
                 for col in data_to_save:
-                    col.extend([np.nan] * abs(max_col_length - len(col)))
+                    # check if the data column type is array-like
+                    if hasattr(col, '__len__') and hasattr(col, '__getitem__') and not isinstance(col, str):
+                        data_len = len(col)
+                        col.extend( [[np.nan] * data_len] * abs(max_col_length - len(col)))
+                    else:
+                        col.extend([np.nan] * abs(max_col_length - len(col)))
 
             where_none = np.array([[True if var is None else False
                                     for var in column]
--- a/posydon/binary_evol/binarystar.py
+++ b/posydon/binary_evol/binarystar.py
@@ -409,8 +409,6 @@ class BinaryStar:
                         # never happen.
                         dtype = 'float64'
 
-                    max_ele_length = 0
-
                     # check type of data and determine what to fill data column with
                     if dtype == 'string':
                         filler_value = 'None'
--- a/posydon/binary_evol/binarystar.py
+++ b/posydon/binary_evol/binarystar.py
@@ -148,7 +148,7 @@ class BinaryStar:
         # Set the initial binary properties
         for item in BINARYPROPERTIES:
             if item == 'V_sys':
-                setattr(self, item, binary_kwargs.pop(item, [0,0,0]))
+                setattr(self, item, binary_kwargs.pop(item, np.array([0,0,0])))
             elif item == 'mass_transfer_case':
                 setattr(self, item, binary_kwargs.pop(item, 'None'))
             elif item == 'nearest_neighbour_distance':
@@ -390,6 +390,7 @@ class BinaryStar:
             # If a binary fails, usually history cols have diff lengths.
             # This should append NAN to create even columns.
             all_equal_length_cols = len(set(col_lengths)) == 1
+
             if not all_equal_length_cols:
                 for col in data_to_save:
                     # check if the data column type is array-like
--- a/posydon/binary_evol/binarystar.py
+++ b/posydon/binary_evol/binarystar.py
@@ -413,12 +413,7 @@ class BinaryStar:
                         # get the max length that data elements have
                         ele_lengths = [len(x) for x in col]
                         max_ele_length = np.max(ele_lengths)
-
-                        try:
-                            sub_dtype = OBJECT_FIXED_SUB_DTYPES[dkey]
-                        except KeyError:
-                            sub_type = ''
-                            pass
+                        sub_dtype = OBJECT_FIXED_SUB_DTYPES.get(dkey, '')
 
                         # array of strings
                         if sub_dtype == 'string':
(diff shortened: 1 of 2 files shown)


elizabethteng (dismissed):

sgossage:
The user who experienced the error that this PR addresses confirms that it appears to solve the issue for their case.

posydon/binary_evol/binarystar.py
```diff
@@ -391,7 +392,13 @@ def to_df(self, **kwargs):
             all_equal_length_cols = len(set(col_lengths)) == 1
             if not all_equal_length_cols:
                 for col in data_to_save:
-                    col.extend([np.nan] * abs(max_col_length - len(col)))
+                    # check if the data column type is array-like
+                    element = col[0]
+                    if hasattr(element, '__len__') and hasattr(element, '__getitem__') and not isinstance(element, str):
```
mkruckow: I'd suggest to define which types of elements need this treatment, there might be more types, where it shouldn't be applied than just strings.
Additionally, we should only replace stuff with nans for floats and integers (may add complex numbers). Thus I'd suggest to change this completely to check the type of the element and add stuff according to the dynamically determined type.
P.S.: As usual, it is dangerous to use the first element only as this might not be representative/well defined.

sgossage: I've added some code to dynamically check which filler values should be used. I checked the data types defined in `popsyn/io.py` and we have `string`, `float64`, and `object`. I added two new data types that are more specific versions of `object`: `object.string` and `object.float64` to specify what the  subtypes within the object should be.

At the moment, I check for the dtype should be, according to how we've defined it in `popsyn/io.py` and set the filler value accordingly.

We could write something fully more generalized, but if so, perhaps we should consider then generalizing the dtypes defined in `popsyn/io.py`.

Also, as a filler for `string` type data, I'm using `'None'`, which is what these data are initialized with. Should we use something else for `string` types?

mkruckow: I'd recommend using an empty string instead of `'None'`.

Your new types won't be understood by numpy or pandas. Thus, we can't come up with new ones unless we reduce them before passing them to any numpy or pandas function.
I think, we need to keep them as `'object'`. But, we could define an additional dictionary, e.g.
```
OBJECT_FIXED_SUB_DTYPES = {
    'V_sys': 'float64',
    ...
}
```
In the case an object dtype is hit, you can check whether there is a given sub dtype for this object by checking for the same key in the new directory. If it is in there you can make a numpy array with this subtype otherwise, simply fill the object with a None type object (or a `np.nan` like the fallback to a float).

By the way, the `nearest_neighbour_distance` shouldn't be a string.

posydon/binary_evol/binarystar.py
```diff
@@ -389,9 +396,50 @@ def to_df(self, **kwargs):
...
-                for col in data_to_save:
-                    col.extend([np.nan] * abs(max_col_length - len(col)))
+                for i, col in enumerate(data_to_save):
+
+                    dkey = all_keys[i]
+                    try:
+                        dtype = properties_dtypes[dkey]
+                    except KeyError:
+                        dtype = ''
+                        pass
```
mkruckow: Just a bit of style (one-liner and without raising errors):
```suggestion
                    dtype = properties_dtypes.get(dkey, '')
```

posydon/binary_evol/binarystar.py
```diff
@@ -389,9 +396,50 @@ def to_df(self, **kwargs):
...
+                    elif dtype == 'object':
+                        # get the max length that data elements have
+                        ele_lengths = [len(x) for x in col]
+                        max_ele_length = np.max(ele_lengths)
+
+                        try:
+                            sub_dtype = OBJECT_FIXED_SUB_DTYPES[dkey]
+                        except KeyError:
+                            sub_type = ''
+                            pass
```
mkruckow: Same as above:
```suggestion
                        sub_dtype = OBJECT_FIXED_SUB_DTYPES.get(dkey, '')
```

mkruckow (approved):
OK, was there a test pop-syn run done with this PR?

sgossage:
> OK, was there a test pop-syn run done with this PR?

Not that I know of. It's been tested that this fixes the previously failing binaries, but I don't think a full population synthesis has been tested with this.

sgossage:
To test, I ran an 8 metallicity pop with 1k binaries each. This completed without error.

maxbriel (approved):
The code looks fine.
@sgossage It doesn't matter that 'nearest_neighbour_distance` is a list, while `V_sys` is a numpy array?

Since populations were run successfully, I'm okay accepting this.

There was a discussion/issue from 2 years ago mentioning that this `V_sys` should be rewritten to not be an array, but separate variables instead: #155 
That would only solve the `V_sys` part, not the 'nearest_neighbour_distance` part though.

philipp-rajah (approved):
I've reviewed this for stylistic comments as I am not familiar with this part of the code, and it looks good to me.

PR #585, opened 2025-05-02
Hotfixes new documentation build

maxbriel:
Currently, not all the tags from Github are correctly pulled in this workflow, as such the code fails.
This should enforce that the Github tags are correctly pulled.

Additionally, this should consolidate the usage of "v" in the linking and folders


mkruckow (approved):
Yes, I was already realizing, that the links on the docs to other versions missed the "v" in the link.

sgossage (approved):

PR #588, opened 2025-05-02
Hotfix documentation-build again

maxbriel:
Another fix to the building of the development version. This time, entering the git folder

I don't understand why this worked locally with Local Github Actions (act) and not on the actual Github Actions. 

@sgossage, you didn't experience issues either with the local workflow, right?


mkruckow (approved):
Yeah, I think this became necessary because of the change in the checkout of the tags:
```
      with:
        fetch-tags: true
        path: 'POSYDON'
```

elizabethteng (approved):

PR #593, opened 2025-05-08
Fix fetching all tags of the repository

maxbriel:
The documentation workflows are failing because they cannot find all the tags in the repository.
It turns out that `fetch-tags: true` doesn't fetch any tags without altering the default `fetch-depth` from 1. See this issue on the checkout repository: https://github.com/actions/checkout/issues/1471

Adding `fetch-depth:0` to the checkout action is the recommended solution.


mkruckow (approved):
OK, let's try it.

elizabethteng (approved):

PR #594, opened 2025-05-08
Bump miniconda version to v3 and add proper bash shell initiation

### LINKED ISSUE #538: [BUG]: Failing installation of extras in the Github Actions
The Github action for the installation of the extras is failing.
The error looks like it comes from mpi4py. We should resolve this to make sure these tests are succeeding

It gets run every week, but seems to have failed for a long time.

maxbriel:
Main points:
- setup-miniconda v2 -> v3
- add bash -el {0} let conda be seen.
- add conda install `qt`
- removed sourcing `conda`; already done.

The latest setup-miniconda version is v3 with the bash shell initiation (set as default for the steps).

This loads `conda` properly. 

The installation still fails due to what seems like a pyQT issue (with QT5 not being installed), although this is not super clear.
I added the latest version to the conda environment.

I was unable to verify if this addition of installing qt allows POSYDON to be installed, since it hangs due to the virtualisation on my machine. However, it no longer crashes as it did without qt installed.

### DIFF
--- a/.github/workflows/install_extras.yml
+++ b/.github/workflows/install_extras.yml
@@ -15,7 +15,7 @@ jobs:
       matrix:
         os: [macos-13, macos-latest, ubuntu-latest]
         python-version: ['3.11']  # can add more if we want to support
-      
+    
     steps:
     - name: Checkout repository
       uses: actions/checkout@v4
--- a/.github/workflows/install_extras.yml
+++ b/.github/workflows/install_extras.yml
@@ -33,6 +33,5 @@ jobs:
     - name: Install POSYDON with extras
       run: |
         conda create -n test-env python=3.11 mpi4py conda-forge::qt -y
-        source $(conda info --base)/etc/profile.d/conda.sh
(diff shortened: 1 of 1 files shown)


maxbriel:
**UPDATE:** I can confirm that the `qt` installation worked on `ubuntu-latest`.

mkruckow:
I just had an idea that we can try it on github without "make a change -> merge -> wait till next week":
Could you change change/add the trigger to run on this PR (taken from the continuous integration):
```
on:
  pull_request:
    branches: [development]
```
Thus, we could see whether it runs through on github. If it does, you can remove this part again before merging this PR.

elizabethteng (dismissed):

maxbriel:
It successfully passed the installation check: https://github.com/POSYDON-code/POSYDON/actions/runs/14906380659/job/41869629733

mkruckow (approved):

astroJeff (approved):

Issue #259, opened 2024-02-26
Support for WIndows


mkruckow:
I do have a new master student working with POSYDON. Here in Geneva the students often use Windows instead of a Unix system. Currently, it looks like the only (bad) working solution is to get a linux environment on WIndows and run all in there.

I'd like to give here a record of issues coming up on Windows, please feel free to add stuff, if you have people using POSYDON on Windows:

- Always use `os.path.join` to gets paths and use it in the way it is meant to work on all platforms.
  - e.g. `os.path.join(PATH_TO_POSYDON, "posydon/popsyn/population_params_default.ini")` from the [tutorial](https://posydon.org/POSYDON/tutorials-examples/population-synthesis/10_binaries_pop_syn.html) won't work and should read `os.path.join(PATH_TO_POSYDON, "posydon", "popsyn", "population_params_default.ini")` There might more of those issues in the tutorials and/or the code itself.
- The signal handlers won't work on Windows. Hence we'd need an alternative there or at least an option to switch it off. The following commands from [posydon/binary_evol/binarystar.py](https://github.com/POSYDON-code/POSYDON/blob/development/posydon/binary_evol/binarystar.py) `signal.signal(signal.SIGALRM, signal_handler)`, `signal.alarm(MAXIMUM_STEP_TIME)`, `signal.alarm(0)` only work on Unix.
- There is a bigger issue with the pkls which can't easily be loaded on Windows because the module we use has missing, but needed functions in the Windows version.

mkruckow:
We probably won't support this.

PR #604, opened 2025-05-21
Update dataset definitions for v2 data release

mkruckow:
- [x] update documentation: adjust size requirements(, remove Windows from supported OS, see issue #259 )
- [x] update default download: switch from v1 data to v2 data
- adjust data sets:
  - [x] adjust descriptions
  - [x] adjust urls and checksums (needs publication on zenodo first)
- may update v1 and super-Eddington datasets (@ZepeiX) as the zenodo version don't contain the new SN models
- [ ] testing it

### DIFF
--- a/docs/_source/getting-started/prerequisites.rst
+++ b/docs/_source/getting-started/prerequisites.rst
@@ -28,9 +28,9 @@ Hardware Requirements
 
 In order to run a population synthesis model with POSYDON, you must ensure that you have enough free storage space. It depends on the dataset how much free space you need:
 
-- v1: approximately **40GB**
-- v2: approximately **140GB** (the default in the :ref:`installation <installation-guide>`)
-- one of the eight metallicities in v2: approximately **20GB**
+- DR1: approximately **40GB**
+- DR2: approximately **140GB** (the default in the :ref:`installation <installation-guide>`)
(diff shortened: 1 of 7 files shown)


astroJeff (dismissed):

astroJeff:
These look good so far - it's just the URL and MD5 checksums for each data that need to be adjusted, right?

mkruckow:
Now, it's only the urls and the check sums. I updated the automatic downloads if files are missing and the documentation under the assumptions there won't be any changes on the tar archives anymore.

mkruckow:
I have renamed stuff to data release (DR) so that we use version (v) only for the code. For more details on the zenodo structure, see https://github.com/POSYDON-code/POSYDON-Internal-Documentation/blob/matthias_zenodo/MESA_grids/processing.md#add-the-set-to-the-zenodo-collection

maxbriel:
I've updated the checksums + the URLs. I tested some basic downloads of the auxiliary data and one of the metallicities.

I'm unsure why so many changed files are showing up. When comparing this branch against development, the changes are less: https://github.com/POSYDON-code/POSYDON/compare/development...matthias_update_datasets_for_v2_release

astroJeff (approved):

elizabethteng (approved):

sgossage (approved):

PR #607, opened 2025-05-30
Use analytic integrals in `profile_collapse.py`

### LINKED ISSUE #596: [BUG]: numerical integration producing high-spin BH in PISN regime
# Describe the bug

While post-processing the grids, @sgossage found PISN models failing during the `do_core_collapse_BH`.

```
Error during MODEL06 direct core collapse prescrition!
We got a=1.1659e+13 from shell 2591 containing 1.7937e-07 M_sun
TF1: Primary enters pair-instability regime
interpolation class: stable_MT
run directory: b'/projects/b1119/POSYDON_GRIDS_v2/HMS-HMS/1e-03_Zsun/rerun_LBV_wind+dedt_energy_eqn_grid_random_combined/Zbase_0.0000_m1_257.9654_m2_155.6811_initial_z_1.4200e-05_initial_period_in_days_1.6561e+01_grid_index_119'
```

This model undergoes PISN and experiences instabilities in the outer layers of the star. As such, MESA is placing many cells in this region with short radius differences. As a result, the density per cell blows up, see the figure below. **This should only be an issue in a small region of parameter space**. 
image

Within the `do_core_collapse_BH` calculation, we are numerically integrating to get the `J_direct` value. Since we are working with very small radius differences and large densities, the numerical accuracy of the integration becomes important and breaks down in the model above. As a result, `J_direct` is not accurate enough and `a_BH` increases above 1.

https://github.com/POSYDON-code/POSYDON/blob/48ca21bfaaea4428242d2c796c1a73ce719f9769/posydon/binary_evol/SN/profile_collapse.py#L584-L612

# Solution

The first solution is to change the numerical integration to an analytical one. The equations we're integrating over are basic, and their integral can be done analytically. This would also increase the accuracy and speed of this calculation.

Additional thoughts/solutions: the mass in the cell does not change to large values. We should also be able to use it instead of the density to reduce the impact of the large density values.
image
image

cgarling:
Replaces `integrate.quad` calls with analytic integrals which @maxbriel suggested may fix https://github.com/POSYDON-code/POSYDON/issues/596.

### DIFF
--- a/posydon/binary_evol/SN/profile_collapse.py
+++ b/posydon/binary_evol/SN/profile_collapse.py
@@ -583,8 +583,8 @@ def do_core_collapse_BH(star,
             # eq. 9 of Batta & Ramirez-Ruiz (2019) for theta<theta_disk
             # J_shell = J_direct + J_disk
 
-            # Analytic integral of int_0^theta sin^3(t) dt is 4 / 3 * (2 + cos(theta)) * sin(theta/2)**4
-            temp1 = 4 /3 * (2 + np.cos(theta_disk)) * np.sin(theta_disk/2)**4
+            # Analytic integral of int_0^theta sin^3(t) dt is 4 / 3 * (2 + cos(theta)) * sin(theta/2)^4
+            temp1 = 4 / 3 * (2 + np.cos(theta_disk)) * np.sin(theta_disk/2)**4
 
             # Analytic integral of int_r-dr^r r^4 dr is r^5/5 - (r-dr)^5/5
             temp2 = r_shell**5 / 5 - (r_shell - dr_shell)**5 / 5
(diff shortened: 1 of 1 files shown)


maxbriel:
@cgarling Thank you for this PR! I will get around to a review later this week

maxbriel (dismissed):
LGTM. 

@POSYDON-code/posydon-core-developers Accepting this PR will require a re-post-processing of the grids to include this change in the grid's core-collapse values

sgossage (dismissed):

astroJeff:
We have rebased your suggestions which required a separate PR, now integrated into PR #653. We have added you in the list of credits. Thank you, @cgarling for the suggestion.

PR #611, opened 2025-06-03
Jeff moe di stefano

astroJeff:
This iterates from Matthias's branch to add the Moe & Di Stefano sampler


posydon/popsyn/Moes_distributions.py
```diff
@@ -0,0 +1,470 @@
...
+                # section 9.1 and Eqn. 5
+                if(mylogP <= 1.0):
+                    Ftwin = FtwinlogPle1
+                else:
+                    Ftwin = FtwinlogPle1 * (1.0-(mylogP-1)/(logPtwin-1.0))
+                if(mylogP >= logPtwin):
+                    Ftwin = 0.0
+                # Power-law slope gamma_largeq for M1 < 1.2 Msun and various P;
+                # Eqn. 9
+                if(mylogP <= 5.0):
```
sgossage:
```suggestion
                if(mylogP < 5.0):
```

posydon/popsyn/Moes_distributions.py
```diff
@@ -0,0 +1,470 @@
...
+                    gl_1p2 = -0.5
+                if(mylogP > 5.0):
```
sgossage:
```suggestion
                if(mylogP >= 5.0):
```

This suggestion is along the lines of the above comment. I haven't made suggestions for the other mentioned equations here to keep things tidy.

posydon/popsyn/Moes_distributions.py
```diff
@@ -0,0 +1,470 @@
...
+                    gl_1p2 = -0.5 - 0.3 * (mylogP-5.0)
+                # Power-law slope gamma_largeq for M1 = 3.5 Msun and various P;
+                # Eqn. 10
+                if(mylogP <= 1.0):
```
sgossage, line 188:
```suggestion
                if(mylogP < 1.0):
```

posydon/popsyn/Moes_distributions.py
```diff
@@ -0,0 +1,470 @@
...
+                    gl_3p5 = -0.5
+                if((mylogP > 1.0) and (mylogP <= 4.5)):
```
sgossage, line 190:
```suggestion
                if((mylogP >= 1.0) and (mylogP <= 4.5)):
```

posydon/popsyn/Moes_distributions.py
```diff
@@ -0,0 +1,470 @@
...
+                # normalize cumfq(q=1.0) = 1
+                cumfq = cumfq / np.max(cumfq)
+                # save to grid
+                self.cumqdist[:,j,i] = cumfq
+                # Given M1 and P, q_factor is the ratio of all binaries
+                # 0.1 < q < 1.0 to those with 0.3 < q < 1.0
+                q_factor = self._idl_tabulate(self.qv, fq)
+                # Given M1 & P, calculate power-law slope eta of eccentricity
+                # distribution
+                if(mylogP >= 0.7):
```
sgossage, line 271: why logP >= 0.7 and not logP > 0.5? (as in eqns. (18) and (19))

posydon/popsyn/Moes_distributions.py
```diff
@@ -0,0 +1,470 @@
...
+
+    def __init__(self, n_M1=101, n_logP=158, n_q=91, n_e=100):
+        """Initializing the class.
+
+        Parameters
+        ----------
+        n_M1 : int (Default: 101)
+            Number of supporting points in the primary's mass.
+        n_logP : int (Default: 158)
+            Number of supporting points in the logarythmic period.
```
sgossage:
```suggestion
            Number of supporting points in the logarithmic period.
```

posydon/popsyn/Moes_distributions.py
```diff
@@ -0,0 +1,470 @@
...
+            Integral value.
+        """
+
+        def integrate_newton_cotes(x, f):
+            """Integrate f from x[0] to x[-1] via Newton cotes.
+
+            Parameters
+            ----------
+            x : ndarray of floats
+                Values of integration variable.
```
sgossage:
```suggestion
                Values of integration variable.
```
```suggestion
                1D array of values of the integration variable.
```

posydon/popsyn/Moes_distributions.py
```diff
@@ -0,0 +1,470 @@
...
+            f : ndarray of floats
+                Function values at x, therefore need to have the same shape as
```
sgossage:
```suggestion
                1D array of function values at given x values, therefore this needs to have the same shape as
```

posydon/popsyn/Moes_distributions.py
```diff
@@ -0,0 +1,470 @@
...
+        This function is similar to the IDL function int_tabulated except, this
+        function seems to be slightly more exact in its solution. Therefore,
+        relative to the IDL code, there are small numerical differences.
+
+        Parameters
+        ----------
+        x : ndarray of floats
+            Values of integration variable.
+        f : ndarray of floats
+            Function values at x, therefore need to have the same shape as x.
```
sgossage:
```suggestion
            1D array of function values at given x, therefore this needs to have the same shape as x.
```

posydon/popsyn/Moes_distributions.py
```diff
@@ -0,0 +1,470 @@
...
+        """Discrete integration in chunks to avoid instabilities for
+        Newton-Cotes rule at high degree.
+        This function is similar to the IDL function int_tabulated except, this
+        function seems to be slightly more exact in its solution. Therefore,
+        relative to the IDL code, there are small numerical differences.
+
+        Parameters
+        ----------
+        x : ndarray of floats
+            Values of integration variable.
```
sgossage:
```suggestion
           1D array of values of the integration variable.
```

sgossage: We may want to specify that these effectively need to be 1D arrays (because of the `newton_cotes` function that is used).

posydon/popsyn/Moes_distributions.py
```diff
@@ -0,0 +1,470 @@
...
+        """
+        self.numM1 = n_M1
+        self.numlogP = n_logP
+        self.numq = n_q
+        self.nume = n_e
+        # ranges where M+D17 has statistics corrected for selection effects:
+        # 0.8 < M1/Msun < 40 with self.numM1 steps in log space
+        self.M1v = 10**(np.linspace(np.log10(0.8), np.log10(40.0), self.numM1))
+        # 0.15 < log10(P/day) < 8.0 with self.numlogP steps
+        self.logPv = np.linspace(0.15, 8.0, self.numlogP)
```
sgossage, line 119: Is this meant to go from -0.15 to 8 (not 0.15 to 8)? This would also answer why later on we use 0.7 as an effective lower limit in the calculations.

posydon/popsyn/Moes_distributions.py
```diff
@@ -0,0 +1,470 @@
...
+                if(myM1 <= 3.0):
+                    eta = eta_3
+                if((myM1 > 3.0) and (myM1 <= 7.0)):
+                    eta = np.interp(np.log10(myM1), np.log10([3.0,7.0]),
+                                    [eta_3, eta_7])
+                if(myM1 > 7.0):
+                    eta = eta_7
+                # Given eta at the specified M1 and P, tabulate eccentricity
+                # distribution
+                if(10**mylogP <= 2.0):
```
sgossage, line 291: Can we explain more where this condition comes from? Is it based on a fit (why eta = -3.2 in `fe = self.ev**(-3.2)` below P = 2.0)?

posydon/popsyn/Moes_distributions.py
sgossage: This is a small detail maybe, but the equations in Moe and Di Stefano 2017 read in the way suggested above, rather than the way it is currently written in the code. Similarly for eqns. 10 and 11. Is there a reason for it to differ from the paper?

This is also the case in eqns. 14 and 15. Eqn 23 as written in code below also differs from the paper similarly.

posydon/popsyn/Moes_distributions.py
astroJeff: I looked into this. It looks like the equations blow up at the limit: log P = 0.5. This is clearly a problem, so I guess when Max coded it up, he added a lower limit of 0.7 instead of 0.5. There is a subsequent check for log P < 0.7 which just uses the value for eta at log P = 0.7. This seems reasonable to me.

posydon/popsyn/Moes_distributions.py
astroJeff: I don't know about this. There is a section in the M&dS paper describing short period systems having difficulty measuring the eccentricity distribution, but they list for periods less than 1 day. The negative slope should strongly preference circular orbits, which makes sense. I'm guessing that Max Moe put this together years ago and decided this value made more sense than what he put in the paper. I think we ought to keep it unless we have a reason to change it.

sgossage: Sounds good -- I think if we can provide some context or source we should, but sounds like that isn't available in this case. I'm OK with keeping it as is.

sgossage (approved):

posydon/popsyn/population_params_default.ini
```diff
@@ -293,254 +317,275 @@
...
+    # float (0, inf) (used only for orbital_scheme = 'separation')
+    # value should be within the separation range given by the used grids
+  orbital_separation_max = 1e5
+    # float (0, inf) (used only for orbital_scheme = 'separation')
+    # value should be within the separation range given by the used grids
+  log_orbital_separation_mean = None
+    # float (0, inf) (used only for orbital_separation_scheme ='log_normal')
+    # value should be within the separation range given above
+  log_orbital_separation_sigma = None
+    # float (0, inf) (used only for orbital_separation_scheme ='log_normal')
```
maxbriel, line 404: Why are these specific values suddenly uncommented?

maxbriel (commented):
This looks fine. I will be rebasing it on v2.1-dev.

The only thing I'm a bit worried about is our initial sampling being able to do multiple samples in one go, while the BinaryPopulation only does one at the time.

maxbriel:
For the rebasing, I had to manually cherry-pick the commits and then merge them to a new branch based on v2.1-dev.  PR #648 is this.

sgossage:
> For the rebasing, I had to manually cherry-pick the commits and then merge them to a new branch based on v2.1-dev. PR #648 is this.

Is this PR to be deprecated in favor of #648 ?

astroJeff:
Closing. This PR is superceded by PR #648, which is rebased into v2.1

PR #613, opened 2025-06-18
Initialize star properties

### LINKED ISSUE #354: Evolving a binary from an arbitrary combination
In V1, it was possible to evolve a binary from an arbitrary state, but in V2 you can only start a binary on a grid.
For example, the system in the tutorial NS+H-rich_core-h-burning can start on the CO-HMS_RLO grid, but not in the detached step.

This is caused by some values not properly being initialised.
I think it should be possible for someone to start a binary with a CO + a fresh star (ZAMS).
This is not currently possible.

### LINKED ISSUE #577: Initialize binaries with all parameters set to null values
When a binary is created, all of its values ought to be initialized. We ought to keep track of all the variables that are ever assigned to a binary object and make sure that they exist from the creation of the binary object.

This is related to issue #354

sgossage:
Initialize `SingleStar` class properties.

### DIFF
--- a/posydon/binary_evol/singlestar.py
+++ b/posydon/binary_evol/singlestar.py
@@ -15,6 +15,7 @@ __authors__ = [
     "Jeffrey Andrews <jeffrey.andrews@northwestern.edu>",
     "Emmanouil Zapartas <ezapartas@gmail.com>",
     "Devina Misra <devina.misra@unige.ch>",
+    "Seth Gossage <seth.gossage@northwestern.edu>"
 ]
 
 
@@ -22,6 +23,7 @@ import numpy as np
 import pandas as pd
 from posydon.utils.common_functions import check_state_of_star
 from posydon.grids.SN_MODELS import SN_MODELS
+from posydon.popsyn.io import STARPROPERTIES_DTYPES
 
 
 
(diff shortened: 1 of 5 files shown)


posydon/unit_tests/utils/test_common_functions.py
```diff
@@ -1742,7 +1747,7 @@ def mock_calculate_lambda_from_profile(profile, donor_star_state,\
                     assert np.isnan(getattr(star, attr))
                 elif (("stripped_He" in s) and ("pure_He" not in attr) and\
                       ("m_core" in attr)):
-                    assert getattr(star, attr) == star.mass
+                    assert np.isnan(getattr(star, attr)) and np.isnan(star.mass)
```
astroJeff, line 1750: Just checking this statement. Is this actually correct? It's a different statement than was written previously. Also, do we want to use `np.isnan` or `pd.isna`?

posydon/binary_evol/singlestar.py
```diff
@@ -133,7 +135,17 @@ def __init__(self, **kwargs):
...
+
+            # set default values when a kwarg is absent
+            dtype = STARPROPERTIES_DTYPES.get(item, '')
+            if dtype == 'float64' or dtype == 'int':
+                default = np.nan
+            elif dtype == 'string':
+                default = ''
+            else:
+                default = None
+
```
astroJeff, line 147: Highlighting this block to make sure there aren't any unintended consequences

astroJeff (changes requested):

sgossage:
This needs to be rebased or cherrypicked to merge into `v2.1-dev`

sgossage:
Currently no issues in regular code usage -- this is more a safeguard for cases where `SingleStar` or `BinaryStar` objects are setup by a user manually.

posydon/unit_tests/utils/test_common_functions.py
sgossage: This depends on what we want to use for the initial value. This case would be when we use np.nan -- we no longer use None (as that was part of the root issue, values being initialized to None), so np.isnan is the more explicit choice if we go with that (since null values are actually NaN). 

However, in general pd.isna is safer to allow things to catch any bad values. In a unit test, maybe we want to make the test explicit I'm thinking.

sgossage:
This was branched from `development`. Rather than rebase and merge the new `v2.1-dev`, I've cherrypicked this and moved it to [PR#686](https://github.com/POSYDON-code/POSYDON/pull/686)

PR #651, opened 2025-08-14
Pop params ini fix

maxbriel:
This removed the double definition of steps introduced with the rework of the matching step.
Must have happened during the rebasing.


astroJeff:
Editing to merge this into `main`, as it's a bug fix

astroJeff:
I see now - this is only a problem with `v2.1-dev`. Reverting to the old merge destination.

astroJeff (approved):

sgossage (approved):

PR #652, opened 2025-08-14
Spin in IF runs fix

### LINKED ISSUE #650: [BUG]: BH spin IF-interpolation
**Describe the bug**
When a CO-HMS binary undergoes RLOF, we update the final BH spin after the mass transfer with the mass gained during the evolution. We only do this in the case of the Nearest Neighbour runs, we do not do this for the initial-final interpolation. Below I link the code for the NN portion of the code:

https://github.com/POSYDON-code/POSYDON/blob/722f0cad8892a230c924acb638465588f45107a0/posydon/binary_evol/MESA/step_mesa.py#L604-L619

maxbriel:
In runs with the initial-final interpolation, the spin was not properly treated. Despite high accretion amounts, the spin remained low. Only in NN runs, was the spin adjusted based on the accretion. See issue #650 for more details.

This adds the same treatment to IF runs for the spin. You can see in the example model below that this increases the spin in a similar fashion as the NN did (see Issue #650 for NN run values). They're not exactly the same, but close enough and better than before.

image

A zip with a notebook to check the difference. Note that the ini files point to the conservative grid interpolator and grid.
[BH_spin_other_grids.zip](https://github.com/user-attachments/files/21773723/BH_spin_other_grids.zip)

### DIFF
--- a/posydon/binary_evol/MESA/step_mesa.py
+++ b/posydon/binary_evol/MESA/step_mesa.py
@@ -893,7 +893,7 @@ class MesaGridStep:
                         current = getattr(star, 'spin')
                         key_p = 'star_%d_mass' % (k+1)
                         # calculate new spin based on masses
-                        spin = cf.spin_stable_mass_transfer(current, prev_mass[k] , fv[key_p])
+                        spin = cf.spin_stable_mass_transfer(current, prev_mass[k], fv[key_p])
                         setattr(star, key, spin)
                         
                     elif key == 'log_R':
--- a/posydon/binary_evol/MESA/step_mesa.py
+++ b/posydon/binary_evol/MESA/step_mesa.py
@@ -843,6 +843,7 @@ class MesaGridStep:
         stars = [binary.star_1, binary.star_2]
         stars_CO = [star_1_CO, star_2_CO]
         fv = self.final_values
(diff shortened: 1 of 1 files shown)


posydon/binary_evol/MESA/step_mesa.py
```diff
@@ -882,15 +883,19 @@ def initial_final_interpolation(self, star_1_CO=False, star_2_CO=False):
...
                         setattr(star, key, None)
                     elif key == 'mass':
                         key_p = 'star_%d_mass' % (k+1)
                         setattr(star, key, fv[key_p])
                     elif key == 'spin':
                         current = getattr(star, 'spin')
-                        setattr(star, key, current)
+                        key_p = 'star_%d_mass' % (k+1)
+                        # calculate new spin based on masses
+                        spin = cf.spin_stable_mass_transfer(current, prev_mass[k] , fv[key_p])
```
astroJeff:
```suggestion
                        spin = cf.spin_stable_mass_transfer(current, prev_mass[k], fv[key_p])
```

astroJeff (approved):

sgossage (approved):

elizabethteng (approved):

PR #656, opened 2025-08-18
Evolution of secondary if primary is CO + S2 and S2 flip

### LINKED ISSUE #600: [BUG]: S2_mass_ZAMS/S1_ZAMS > 1
**Describe the bug**
In the population synthesis, I've seen binaries with a primary mass lower than the lower bound of the sampling limit, ie. $M_\mathrm{1,ZAMS} < M_\mathrm{lim}$. This should not be possible, but in these systems, $M_\mathrm{2,ZAMS}$ is larger than  $M_\mathrm{lim}$.
I've seen this with ~4000 in a run of 1.000.000 binaries.

Since $M_\mathrm{2,ZAMS}$ is larger than $M_\mathrm{1,ZAMS}$, I suspect that the stars have been flipped in their evolution. There's the function `flip_stars` ([the function](https://github.com/POSYDON-code/POSYDON/blob/ce0aeca0f88493dfce698b5e5ebb83853b6befa1/posydon/utils/common_functions.py#L1287C1-L1287C24)), which is used a few times inside the population synthesis (See [step_MESA](https://github.com/POSYDON-code/POSYDON/blob/development/posydon/binary_evol/MESA/step_mesa.py#L26))
It flips the stars in the binary around. Given that this happens inside the binary, the original S1 and S2 are swapped and they're not flipped back.

**To Reproduce**
Example systems, where S1 and S2 are given like they're in the oneline and history (so they're already flipped).

```
S1_mass = 105.786680
S2_mass =  115.428720 
orbital_period = 31.519519
```

```
S1_mass = 9.611029
S2_mass = 9.845907 
orbital_period = 3.820571
```

**Expected behavior**
Keep the original S1 and S2 the same throughout the evolution. I suspect that this will cause some part of the code to break. This should be done carefully.

maxbriel:
In solving issue #600 with the latest v2.1-dev version, I encountered an issue with `step_detached` introduced in v2.1-dev with the separation of the matching.

When the `primary` star has become a compact object, we replace the primary's properties in `step_detached` with the secondary. This makes it possible to use the solver correctly. However, it needs to grab the values from the `secondary` star, not the `primary` to do this. In the rewrite, this got lost.

Here I reimplement it, such that the solver gives back the correct ages.
Additionally, I also solved #600 by making sure that the star initiating the RLO is always the system that becomes the main object. This stops S1 and S2 from switching.

Code cleanup + examples on the way (or available below)

### DIFF
--- a/posydon/binary_evol/DT/step_detached.py
+++ b/posydon/binary_evol/DT/step_detached.py
@@ -353,7 +353,7 @@ class detached_step:
                                     dense_output=True)
 
             t_after_ODEsolution = time.time()
-            
+
             if self.verbose:
                 ivp_tspan = t_after_ODEsolution - t_before_ODEsolution
                 print(f"\nODE solver duration: {ivp_tspan:.6g} sec")
@@ -372,7 +372,7 @@ class detached_step:
             secondary.state = check_state_of_star(secondary, star_CO=False)
             for timestep in range(-len(t[:-1]), 0):
                 secondary.state_history[timestep] = check_state_of_star(secondary, i=timestep, star_CO=False)
-            
+
             if primary.state == "massless_remnant":
                 pass
             elif primary.co:
--- a/posydon/binary_evol/DT/step_detached.py
+++ b/posydon/binary_evol/DT/step_detached.py
@@ -305,7 +305,7 @@ class detached_step:
 
         # match stars to single star models for detached evolution
         primary, secondary, only_CO = self.track_matcher.do_matching(binary, "step_detached")
-
+        
         if only_CO:
             if self.verbose:
                 print("Binary system only contains compact objects."
--- a/posydon/binary_evol/DT/step_detached.py
+++ b/posydon/binary_evol/DT/step_detached.py
@@ -351,11 +351,8 @@ class detached_step:
                                         secondary.omega0, primary.omega0],
                                     args=(primary, secondary),
                                     dense_output=True)
-
+            
             t_after_ODEsolution = time.time()
-            print(primary.mass, secondary.mass)
-            print(primary.state, secondary.state)
-            print(binary.state)
             
             if self.verbose:
                 ivp_tspan = t_after_ODEsolution - t_before_ODEsolution
--- a/posydon/binary_evol/DT/step_detached.py
+++ b/posydon/binary_evol/DT/step_detached.py
@@ -305,7 +305,6 @@ class detached_step:
 
         # match stars to single star models for detached evolution
         primary, secondary, only_CO = self.track_matcher.do_matching(binary, "step_detached")
-        
         if only_CO:
             if self.verbose:
                 print("Binary system only contains compact objects."
@@ -372,7 +371,8 @@ class detached_step:
             secondary.state = check_state_of_star(secondary, star_CO=False)
             for timestep in range(-len(t[:-1]), 0):
                 secondary.state_history[timestep] = check_state_of_star(secondary, i=timestep, star_CO=False)
-
+            
+            
             if primary.state == "massless_remnant":
                 pass
             elif primary.co:
@@ -385,6 +385,7 @@ class detached_step:
                 primary.state = check_state_of_star(primary, star_CO=False)
                 for timestep in range(-len(t[:-1]), 0):
                     primary.state_history[timestep] = check_state_of_star(primary, i=timestep, star_CO=False)
+
 
             ## CHECK IF THE BINARY IS IN RLO
             if res.t_events[0] or res.t_events[1]:
@@ -444,6 +445,7 @@ class detached_step:
 
             ## CHECK IF STARS WILL UNDERGO CC
             elif res.t_events[2]:
+                print('HERE')
                 # reached t_max of track. End of life (possible collapse) of secondary
                 if secondary == binary.star_1:
                     binary.event = "CC1"
@@ -470,6 +472,7 @@ class detached_step:
                             "step, but do not have the same mass")
 
             elif res.t_events[3]:
+                print("HERE2")
                 # reached t_max of track. End of life (possible collapse) of primary
                 if secondary == binary.star_1:
                     binary.event = "CC2"
--- a/posydon/binary_evol/DT/step_detached.py
+++ b/posydon/binary_evol/DT/step_detached.py
@@ -351,7 +351,7 @@ class detached_step:
                                         secondary.omega0, primary.omega0],
                                     args=(primary, secondary),
                                     dense_output=True)
-            
+
             t_after_ODEsolution = time.time()
             
             if self.verbose:
@@ -372,7 +372,6 @@ class detached_step:
             secondary.state = check_state_of_star(secondary, star_CO=False)
             for timestep in range(-len(t[:-1]), 0):
                 secondary.state_history[timestep] = check_state_of_star(secondary, i=timestep, star_CO=False)
-            
             
             if primary.state == "massless_remnant":
                 pass
--- a/posydon/binary_evol/DT/step_detached.py
+++ b/posydon/binary_evol/DT/step_detached.py
@@ -305,6 +305,7 @@ class detached_step:
 
         # match stars to single star models for detached evolution
         primary, secondary, only_CO = self.track_matcher.do_matching(binary, "step_detached")
+
         if only_CO:
             if self.verbose:
                 print("Binary system only contains compact objects."
@@ -352,7 +353,10 @@ class detached_step:
                                     dense_output=True)
 
             t_after_ODEsolution = time.time()
-
+            print(primary.mass, secondary.mass)
+            print(primary.state, secondary.state)
+            print(binary.state)
+            
             if self.verbose:
                 ivp_tspan = t_after_ODEsolution - t_before_ODEsolution
                 print(f"\nODE solver duration: {ivp_tspan:.6g} sec")
@@ -385,7 +389,6 @@ class detached_step:
                 primary.state = check_state_of_star(primary, star_CO=False)
                 for timestep in range(-len(t[:-1]), 0):
                     primary.state_history[timestep] = check_state_of_star(primary, i=timestep, star_CO=False)
-
 
             ## CHECK IF THE BINARY IS IN RLO
             if res.t_events[0] or res.t_events[1]:
@@ -445,7 +448,6 @@ class detached_step:
 
             ## CHECK IF STARS WILL UNDERGO CC
             elif res.t_events[2]:
-                print('HERE')
                 # reached t_max of track. End of life (possible collapse) of secondary
                 if secondary == binary.star_1:
                     binary.event = "CC1"
@@ -472,7 +474,6 @@ class detached_step:
                             "step, but do not have the same mass")
 
             elif res.t_events[3]:
-                print("HERE2")
                 # reached t_max of track. End of life (possible collapse) of primary
                 if secondary == binary.star_1:
                     binary.event = "CC2"
--- a/posydon/binary_evol/DT/track_match.py
+++ b/posydon/binary_evol/DT/track_match.py
@@ -1498,7 +1498,7 @@ class TrackMatcher:
         interp1d["m0"] = match_m0
 
         if star.co:
-            kvalue["mass"] = np.zeros_like(kvalue["mass"]) + secondary.mass
+            kvalue["mass"] = np.zeros_like(kvalue["mass"]) + star.mass
             kvalue["R"] = np.zeros_like(kvalue["log_R"])
             kvalue["mdot"] = np.zeros_like(kvalue["mdot"])
             interp1d["mass"] = PchipInterpolator(age, kvalue["mass"])
--- a/posydon/binary_evol/DT/track_match.py
+++ b/posydon/binary_evol/DT/track_match.py
@@ -1914,8 +1914,9 @@ class TrackMatcher:
 
         if binary.non_existent_companion == 0: # both stars exist, detached step of a binary
 
-            # states match, either both H stars or both He stars
-            if (all(s_valid) and (all(s_htrack) or all(~s_htrack))):
+            # states match, either both H stars or both He stars and not any COs
+            # prevents He+CO going into here.
+            if (all(s_valid) and (all(s_htrack) or all(~s_htrack)) and not (any(s_CO))):
                 primary = s_arr[0]
                 primary.co = s_CO[0]
                 primary.htrack = s_htrack[0]
--- a/posydon/binary_evol/DT/track_match.py
+++ b/posydon/binary_evol/DT/track_match.py
@@ -8,7 +8,8 @@ __authors__ = [
     "Kyle Akira Rocha <kylerocha2024@u.northwestern.edu>",
     "Jeffrey Andrews <jeffrey.andrews@northwestern.edu>",
     "Camille Liotine <cliotine@u.northwestern.edu>",
-    "Seth Gossage <seth.gossage@northwestern.edu>"
+    "Seth Gossage <seth.gossage@northwestern.edu>",
+    "Max Briel <max.briel@gmail.com>",
 ]
 
 import os
--- a/posydon/binary_evol/DT/track_match.py
+++ b/posydon/binary_evol/DT/track_match.py
@@ -1362,7 +1362,7 @@ class TrackMatcher:
         return match_vals, best_sol
 
 
-    def get_star_match_data(self, binary, star, secondary_htrack=None,
+    def get_star_match_data(self, binary, star,
                             copy_prev_m0=None, copy_prev_t0=None):
         """
             Match a given component of a binary (i.e., a star) to a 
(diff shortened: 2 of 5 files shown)


maxbriel:
<details>
<summary> Setup details </summary>

```
from posydon.binary_evol.binarystar import BinaryStar, SingleStar
from posydon.binary_evol.simulationproperties import SimulationProperties
from posydon.popsyn.io import simprop_kwargs_from_ini


sim_kwargs = simprop_kwargs_from_ini('population_params_NN.ini', verbose=True)
metallicity = {'metallicity':1, 'verbose':True}

sim_kwargs['step_HMS_HMS'][1].update(metallicity)
sim_kwargs['step_CO_HeMS'][1].update(metallicity)
sim_kwargs['step_CO_HMS_RLO'][1].update(metallicity)
sim_kwargs['step_CO_HeMS_RLO'][1].update(metallicity)
sim_kwargs['step_detached'][1].update(metallicity)
sim_kwargs['step_disrupted'][1].update(metallicity)
sim_kwargs['step_merged'][1].update(metallicity)
sim_kwargs['step_initially_single'][1].update(metallicity)

sim_pop = SimulationProperties(**sim_kwargs)

sim_pop.load_steps(verbose=False)
```

</details>

# Example binary 1 (ZAMS mass flipped)

```
star_1 = SingleStar(**{'mass': 1.400713, 'state': 'H-rich_Core_H_burning','metallicity':1,
                'natal_kick_array':  [0, 4.190728383757787, 1.1521129697118118, 5.015343794234789]})
star_2 = SingleStar(**{'mass': 10.438541, 'state': 'H-rich_Core_H_burning','metallicity':1,
binary = BinaryStar(star_2, star_1, **{'time': 0.0,  'state': 'detached',  'event': 'ZAMS', 
                        'orbital_period':  9.824025,'eccentricity': 0.0}, properties = sim_pop)
```

> [!IMPORTANT]
> In the original evolution S1_ZAMS and S2_ZAMS are swapped! In this weird case, S2 is initiating the RLO and CE.
> This means that it should become the base of the merger (as intended in the original code). However, the original code also swapped the star histories around (but not the binary history!!!!). In this PR, the histories stay the same, and S2 becomes the merger.

## Original evolution
image

## New evolution
image


# Example binary 2 (no changes)
```
star_1= SingleStar(**{'mass': 30.845907 , 'state': 'H-rich_Core_H_burning','metallicity':1,
                'natal_kick_array':  [0, 4.190728383757787, 1.1521129697118118, 5.015343794234789]})
star_2 = SingleStar(**{'mass': 30.611029, 'state': 'H-rich_Core_H_burning','metallicity':1,
                        'orbital_period':  30.820571,'eccentricity': 0.0}, properties = sim_pop)
```
> ![WARNING]
> I messed up the natal kick for the secondary, so that evolution is different!!!

## Original Evolution
image

## New evolution
image


# Example binary 3 (Binary crashed!!

```
star_1 = SingleStar(**{'mass': 9.188942722072198, 'state': 'H-rich_Core_H_burning','metallicity':1.,
                'natal_kick_array': [919.6808229267539, 1.6877410132908262, 1.1499328001057614, 3.399595205766019]})
star_2 = SingleStar(**{'mass': 1.0634074981129351 , 'state': 'H-rich_Core_H_burning','metallicity':1.,
                        'orbital_period':  183152.99061144073,'eccentricity': 0.0}, properties = sim_pop)
```

Thrown Error: ValueError: Undefined next step given stars/binary states ('NS', 'H-rich_Core_H_burning', 'disrupted', 'CC1').

> [!NOTE]
> The original binary fails! It goes to `CC1`, which is incorrect. S1 is a compact object, not a star.


## Original Evolution

image


## New Evolution
image


# Binary 4 (flipped S1 & S2)

```
star_1= SingleStar(**{'mass': 9.845907 , 'state': 'H-rich_Core_H_burning','metallicity':1,
                'natal_kick_array':  [00, 4.190728383757787, 1.1521129697118118, 5.015343794234789]})
star_2 = SingleStar(**{'mass': 9.611029, 'state': 'H-rich_Core_H_burning','metallicity':1,
                        'orbital_period':  3.820571,'eccentricity': 0.0}, properties = sim_pop)
```

> [!IMPORTANT]
> S1 is initiated as the more massive star. But in the history, S2 has become the more massive one (at ZAMS).
> Again the system is swapped around, this time because S2 is in a slightly more "evolved" state (despite the initially lower mass)

## Original evolution
image


## New evolution
image


# Test notebooks

(messy)

[test_binaries.zip](https://github.com/user-attachments/files/21838852/test_binaries.zip)

posydon/binary_evol/DT/step_detached.py
```diff
@@ -353,7 +353,7 @@ def __call__(self, binary):
                                     dense_output=True)
 
             t_after_ODEsolution = time.time()
-
+
```
astroJeff: I'm guessing these changes are automatic. I generally prefer having no spaces in empty lines. It's a super minor thing, but can we revert this change?

posydon/binary_evol/DT/step_detached.py
```diff
@@ -372,7 +372,7 @@ def __call__(self, binary):
             secondary.state = check_state_of_star(secondary, star_CO=False)
             for timestep in range(-len(t[:-1]), 0):
                 secondary.state_history[timestep] = check_state_of_star(secondary, i=timestep, star_CO=False)
-
+
```
astroJeff: Same comment as above...

posydon/binary_evol/DT/step_detached.py
maxbriel: Sure. I prefer to have them there, since it's the same indentation level, but I will remove them

posydon/binary_evol/DT/track_match.py
```diff
@@ -1489,7 +1498,7 @@ def get_star_match_data(self, binary, star,
         interp1d["m0"] = match_m0
 
         if star.co:
-            kvalue["mass"] = np.zeros_like(kvalue["mass"]) + star.mass
+            kvalue["mass"] = np.zeros_like(kvalue["mass"]) + secondary.mass
```
sgossage:
```suggestion
            kvalue["mass"] = np.zeros_like(kvalue["mass"]) + star.mass
```
In testing, this line creates a new error in the calculation of the orbital separation; reverting to the previous behavior eliminates the error. I think this still needs to be `star.mass`, otherwise it assigns the secondary's mass, rather than the current star's. Is there a reason that this needs to change in the cases that this PR addresses?

sgossage: A binary to test this with is e.g., 

```
star1 = SingleStar(**{'mass': 7.552858,#29829485,
                      'natal_kick_array': [40.91509926587841, 2.6295454150818256, 1.6718337470964977, 6.0408769315244895]})
star2 = SingleStar(**{'mass': 6.742063, #481560266,

binary = BinaryStar(star1, star2,
                    **{'time': 0.0, 'state': 'detached', 'event': 'ZAMS', 'orbital_period': 17.957531550841225, 
                       'eccentricity': 0.0, 'history_verbose':True},
                        properties = sim_prop)
```

sgossage: In fact, I think the issue is more that we need to guard against the case where the `secondary` is `None`, in which case fall back to the old behavior.

astroJeff: This is already caught by the new if statements in lines 1410 in that same file. Is it possible that your binary finds an edge case where secondary isn't being passed when it should be?

maxbriel: @sgossage Thank you for checking and finding this example binary that fails! I changed it to `secondary.mass`, because that's how it was implemented in the previous version.
https://github.com/POSYDON-code/POSYDON/blob/5c1d02de3bf81e3823945c10016ac974dc50935a/posydon/binary_evol/DT/step_detached.py#L1010

However, I think that you're right though. It makes much more sense for the mass to remain the star mass (rather than the CO mass), since this goes into the differential equations, which include the orbital evolution. When the wrong mass is input, the solution is wack.
I don't know how this was working before.

maxbriel: Some voodoo magic is happening. The fixed version reproduces the same result as v2.0.0
# v2.0.0
image

# fixed version with `star.mass`
Screenshot 2025-08-18 at 20 21 52

sgossage: In the previous version where `star2.mass` is used, that corresponds actually to the new code's `star.mass`.  The current function is written to get the match data of the passed `star`. In the old code, we just passed if the secondary was a CO (that's what `co` refers to in that chunk of code back then). The new code is written to be more explicit, so now we check if `star.co` is `True`, and if so use the `star.mass`. In this case, `star.mass` is the CO mass (if `star` is a CO).

sgossage (dismissed):

posydon/binary_evol/DT/track_match.py
sgossage: @astroJeff , thanks, the `None` cases are guarded against as you pointed out and the issue was the usage of `secondary.mass`

sgossage:
@maxbriel , one thing that might be good is to instead just pass `secondary.htrack` rather than the entire `secondary` instance.

ezapartas (approved):

sgossage:
@maxbriel , I was looking into this a bit more and I'm not sure if the fix we implemented fully addresses the issues. I think it's fine to accept as is, esp. if its holding up other projects, but I think there may be something up with the logic chain in `step_detached` that handles the events that may need to be addressed in the future, and I wonder if what we implemented is more of a workaround. I'll outline what I'm thinking below.

For Example 3 above, it seems like things go more or less as they should, except that `solve_ivp` terminates due to `s.t_events[3]` triggering. This event causes termination of `step_detached` if the integration time exceeds the max age of the primary. Since the primary is a more massive star, it's very short compared to the secondary. So, that triggers properly.

Also, it is correct that `secondary != binary.star_1`, since `secondary` holds the less evolved star, so `binary.event` is set to `CC1`. That's erroneous when the primary is a CO though (it can't core collapse).

I think what we implemented works because we switch to using the H-rich grid in the case of a CO, which is typically of a smaller mass, and hence fairly long lived on the H-rich grid, allowing the detached step integration to proceed further and not terminate via `s.t_events[3]` .

I was checking into correcting the logic, but Quest analytics nodes have been moving at a snail's pace, so had to stop for today. In the end though, maybe switching the `htrack` is the best route.

maxbriel:
@sgossage I agree with you. I think that's how the logic works. It's definitely based on the `primary` being set to have the same final "age" as the `secondary`, which results in reaching `s.t_events[2]`. This is why we originally ended up with it reaching `s.t_events[3]`, where the `primary` reached the max_time (of its model) due to being grabbed from the He star grid.

We can hold off on it for a bit longer and investigate more if needed. I will run things of a branch, then, with multiple PRs.

posydon/binary_evol/DT/track_match.py
```diff
@@ -1421,11 +1424,17 @@ def get_star_match_data(self, binary, star,
...
             return None, None, None
-
-        if star.htrack:
-            self.grid = self.grid_Hrich
+        
+        if star.co:
+            if secondary.htrack:
+                self.grid = self.grid_Hrich
+            else:
+                self.grid = self.grid_strippedHe
```
sgossage:
```suggestion
```

Suggesting to make change elsewhere.

posydon/binary_evol/DT/track_match.py
```diff
@@ -1421,11 +1424,17 @@ def get_star_match_data(self, binary, star,
...
         else:
-            self.grid = self.grid_strippedHe
+            if star.htrack:
+                self.grid = self.grid_Hrich
+            else:
+                self.grid = self.grid_strippedHe
```
sgossage:
```suggestion
        if star.htrack:
            self.grid = self.grid_Hrich
        else:
            self.grid = self.grid_strippedHe
```

Suggesting to make change elsewhere.

sgossage:
hi @maxbriel , I think we got it right the first time after looking into it. I'd just suggest that we make the change here

```
            # states mismatch, one is a CO and other is an H or He star
            elif (all(s_valid) and (any(s_CO) and not all(s_CO))):

                ...

                # in these cases, we utilize the same htrack as the surviving secondary
                # when constructing the interp1d of the CO star
                primary.htrack = s_htrack[~CO_mask].item()
```

around line 1950 in `track_match.py` instead, but let me know what you think. This would set things so that whenever the primary is a CO and secondary is a normal star, we always use the `htrack` of the non-CO star.

We'd also need to remove the input of `secondary` to `get_track_match_data` if we do this.

I had a look at the logic in `step_detached`, and I do not think it is worth making changes there at this point. IMO we wouldn't have `ev_max_time2(self, t, y, primary, secondary)` as a termination event (checking if the time has exceeded that of the primary) when the primary is a CO. However, doing that introduces scaling errors. I think that's because the interpolators between the two stars possess very different scalings if they're very different stars.

BTW I found a slight (harmless in our cases at least) error in the wrapper for the `solve_ivp` events. It never actually sets the `terminal` property; it is hardcoded as `True`. At line 43 we have

```
def event(terminal, direction=0):
    """Return a helper function to set attributes for solve_ivp events."""
    def dec(f):
        f.terminal = True
        f.direction = direction
        return f
    return dec
```

but we should have `f.terminal = terminal` to allow someone to change things.

maxbriel:
I see. If we go that route, we also need to adapt the ones when there is a `massless_remnant`. 
It effectively comes down to the same solution, except where we implement it.
I don't have a strong preference. @sgossage, do you have a preferred approach?

sgossage:
That's right, it's the same solution, just in a different place. I agree about the massless remnant logic changing as well there. 

```
        # star 1 is a massless remnant, only star 2 exists
        elif binary.non_existent_companion == 1:
            # massless remnant
            ...
            primary.htrack = s_htrack[1]
           ...

        # star 2 is a massless remnant, only star 1 exists
        elif binary.non_existent_companion == 2:
            ...
            primary.htrack = s_htrack[0]
            ...
```

I prefer making the changes in `determine_star_states` if we can so that the logic handling these changes is all contained in one spot. Otherwise we have assignments happening for `htrack` there, and then again conditionally in `get_star_match_data`.

maxbriel:
@sgossage I've changed it to do it in `determine_star_state`. 

While testing with a population, I ran into more issues, which was why I made the `mb_matching_fix` branch and PR #657. I merged that PR into this branch, such that it's easier to keep track of everything.
It adds the following:

- Fixes the output structure for "later" matching attempts, such that the right information is passed along, if a matched failed.
- Renames a variable such that it's correctly used in the output. Otherwise, the variable is undefined.
- Fixes that interp1d might not be set for the primary or secondary, instead we check for the existence of the variable.

It mostly housekeeping of "failed" outputs to make sure the structure is similar to what's expected.

I also found that the `accreted_He_Core_He_burning` state was originally matched to Helium star, but failed here with an invalid state (see Issue #659 ). I added it back to the He matching list.

sgossage (dismissed):

maxbriel:
I am still having some issues. It looks like the minimizer is failing in some cases, when x0 is very close to a good solution. Then the scipy minimizer throws an error with this as the output:

```
Matching attempt completed:
Best solution: 2.184285249585575e-07 (tol = 0.01)
sol.success = False
Matching result: FAILED
Reason: Optimizer failed (sol.success = False)
Optimizer termination reason: Linear search failed
```

These used to evolve fine in v2.0.0.

maxbriel:
It looks like the result/track is already super close to the final value. I added a "tol=1e-6" to the scipy minimizer, which allows some of them to pass correctly with the Powell method (which was hitting maxfunc calls).
They still fail in the initial TNC minimisation due to the initial linear search failing.

So this is not really a good solution.

sgossage:
hey @maxbriel , can you please check if setting unbounded upper age limits for the minimization routine resolves the remaining issues? Lines 350 and 352 may need to change as below to achieve the same behavior as v2.0.0. Doing this seems to resolve the issue on my end.

```
348        # min/max ranges of ages for each grid
349        t_min_H = 0.0
350        t_max_H = None
351        t_min_He = 0.0
352        t_max_He = None
```

posydon/binary_evol/DT/track_match.py
```diff
@@ -1100,7 +1101,7 @@ def get_match_attrs(match_type="default", match_ok=True):
                         print(f"We cannot use match_type={match_type} " \
                                 "since star is on the MS. Skipping...")
                     match_ok = False
-                    return None, None, False, match_ok
+                    return (None, None, False, match_ok), (None, None, None)
```
sgossage:
```suggestion
                    match_attrs = (None, None, new_htrack, match_ok)
                    scls_bnds = (None, None, None)
                    return match_attrs, scls_bnds
```

posydon/binary_evol/DT/track_match.py
```diff
@@ -1183,7 +1184,7 @@ def get_match_params(match_type, match_ok=True):
             rescale_facs, bnds, scalers = scls_bnds
 
             if not match_ok:
-                return None, None, None, new_htrack, match_ok
+                return None, (new_htrack, None, None, None), None, match_ok
```
sgossage:
```suggestion
                fnc_args = (new_htrack, None, None, None)
                return None, fnc_args, None, match_ok
```

sgossage (approved):

maxbriel:
I ran a 1.000.000 population with this branch, and it works. @sgossage thank you for the help and insights on this PR! There are a few other failed binaries, for other reasons, for which I will make separate issues. These already existed in v2.0.0 as well.

astroJeff (approved):

PR #657, opened 2025-08-18
Matching output fixes

maxbriel:
1. Fixes the output structure for "later" matching attempts, such that the right information is passed along, if a matched failed.
2. Renames a variable such that it's correctly used in the output. Otherwise, the variable is undefined.
3. Fixes that interp1d might not be set for the `primary` or `secondary`, instead we check for the existence of the variable.

### DIFF
--- a/posydon/binary_evol/DT/track_match.py
+++ b/posydon/binary_evol/DT/track_match.py
@@ -1346,7 +1346,7 @@ class TrackMatcher:
                            "\nOptimizer termination reason: "
                            f"{best_sol.message}") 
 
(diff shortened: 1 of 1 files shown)


astroJeff:
Thanks for these checks. Just to make sure, these checks hardcode the number of values returned. We just need to make sure this doesn't cause problems in the future, if we change our matching algorithm.

maxbriel:
These make sure that each function returns the shape and number of objects it is expecting to be returned. It does use the shape in some, because it uses those specific values to determine "errors".

I am adding these to PR #656, because they're important to solving some of the issues there too. Closing this PR.

PR #658, opened 2025-08-18
add try/finally around evolution of single binary

maxbriel:
I noticed that while debugging and running individual binaries the kernel would crash after a failed binary.
I suspect that's because a `signal.alarm` is started, but never properly closed. 

We caught this before in the population runs, but this does it on an individual binary level. 
The `finally` always runs, even when an Error is raised. 
Should the `post-step` still be ran?

### DIFF
--- a/posydon/binary_evol/binarystar.py
+++ b/posydon/binary_evol/binarystar.py
@@ -238,7 +238,8 @@ class BinaryStar:
                     
         finally:
             signal.alarm(0)     # turning off alarm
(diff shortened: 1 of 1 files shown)


astroJeff:
Looks good to me. You mean the `post_evolve` step, right? I think it's fine - it should run after the `finally` block, right, or do we explicitly need an `except: pass` statement?

sgossage (approved):

astroJeff (approved):

PR #660, opened 2025-08-19
[PATCH v2.1.2] Simplify `SimulationProperties` setup, initial eccentricity for `BinaryStar`, tolerance controls for `step_detached`, and more

sgossage:
Simplifying some commonly used commands.

To run a `BinaryStar`, certain steps require a metallicity to pull data from. Right now you have to set this manually:

```
sim_kwargs = simprop_kwargs_from_ini('population_params.ini')

# manually add the metallicity to each step that requires it
metallicity = {'metallicity':1}

sim_kwargs['step_HMS_HMS'][1].update(metallicity)
sim_kwargs['step_CO_HeMS'][1].update(metallicity)
sim_kwargs['step_CO_HMS_RLO'][1].update(metallicity)
sim_kwargs['step_CO_HeMS_RLO'][1].update(metallicity)
sim_kwargs['step_detached'][1].update(metallicity)
sim_kwargs['step_disrupted'][1].update(metallicity)
sim_kwargs['step_merged'][1].update(metallicity)
sim_kwargs['step_initially_single'][1].update(metallicity)

sim_prop = SimulationProperties(**sim_kwargs)
# Load the steps and required data
sim_prop.load_steps(verbose=True)
```

Now you can either just give `sim_prop.load_steps()` a metallicity, e.g., 

```
sim_kwargs = simprop_kwargs_from_ini('population_params.ini')
sim_prop = SimulationProperties(**sim_kwargs)
sim_prop.load_steps(metallicity = 1.0, verbose=True) 
```

to do the same. Or you could use

`sim_props = SimulationPropeties.from_ini(path_to_ini, load_steps=True, metallicity=1, verbose=True)`

to do all this in one line.

You can also supply your own `kwargs` to override what is in the `.ini` file. For example, you can override steps like this

```
my_kwargs = {}
my_kwargs['step_detached'] = (my_detached_step, {}) # direct to custom step
my_kwargs['step_disrupted'] = (my_disrupted_step, {}) # direct to custom step

sim_props = SimulationProperties.from_ini(path_to_ini, load_steps=True, metallicity=1, verbose=True, **my_kwargs)
```

and/or you can still provide custom steps in the `.ini` itself file too.

Other things:

1. Adds the function `load_a_step()`, which loads an individual step. The older function `load_steps()` simply loops through all steps and calls this function to load each individually. This adds the flexibility to reload just one step with new settings for example, without having to reload all steps (e.g., `sim_props.load_a_step('step_disrupted', (DisruptedStep, {'verbose': True}))`). You can also load kwargs for a step from a .ini file: `sim_props.load_a_step(`step_disrupted`, from_ini="my_params.ini")`

2. This PR also includes a small fix that ensures that the eccentricity of a newly created `BinaryStar` is set to 0, rather than `None`, to avoid `TypeError` from occurring when arithmetic is performed using an unset eccentricity. This can happen if a user creates a `BinaryStar` without explicitly initializing the eccentricity themself and it enters `step_detached`. 

3. Also allows users to control the `matching_tolerance` and `matching_hard_tolerance` of `step_detached` (these params have been added o the population params `.ini` file).
4. Small fix in `step_SN` where `binary.event` is not set to `None` as it should be if `kick = False`
5. In `BinaryStar`, adds a `deepcopy` of both `SingleStar` objects that are passed at `__init__`. This is to avoid modifying the original memory reference of the `SingleStar` objects.
6. [BUG] This also fixes an error in `step_SN` where the orbital period and separation would not be updated if `kick = False`
7. [BUG] Fixes an error in the signs of G18 and M15 magnetic braking prescriptions. Adds `magnetic_braking_mode` as an option in `default_population_params.ini` for selecting prescriptions other than ``RVJ83``.
8. [BUG] The binary/star history was not properly flushed in cases where `step_mesa` was entered using IF/kNN interpolation and the MESA tracks final time exceeded the max Hubble time; these cases require track interpolation and the entire interpolated history would be appended. Now this history is properly flushed after interpolation.
9. All `interp1d` properties that evolve with time are now accessible in the `detached_evolution` class during evolution. This makes calculating what you want more flexible. Also the time and current timestep are tracked now.
10. Organizing `step_detached` so that `solve_ivp` is called by a helper function of `step_detached`. This makes it so that you can customize just parts of the `solve_ivp()` setup a bit more easily.
11. Compact objects are now updated separately from orbit/regular stars after detached evolution in their own update function. These stars are handled in a special way, so this gives more flexibility in how they are handled/updated after detached evolution.


astroJeff (approved):

sgossage:
I made this into a class method for `SimulationProperties` that has a bit more flexibility than what I implemented in this PR before. Although there's more that can be done, I'll save it for future PRs so this does not hang.

posydon/binary_evol/DT/step_detached.py
```diff
@@ -30,6 +30,7 @@
                                             check_state_of_star,
                                             set_binary_to_failed,
                                             zero_negative_values)
+
```
maxbriel:
```suggestion
```

posydon/binary_evol/DT/track_match.py
```diff
@@ -286,8 +288,8 @@ def __init__(
 
         self.metallicity = convert_metallicity_to_string(metallicity)
         self.matching_method = matching_method
-        self.matching_tolerance = 1e-2 # DEF: 1e-2
-        self.matching_tolerance_hard = 1e-1 # DEF: 1e-1
+        self.matching_tolerance = matching_tolerance # DEF: 1e-2
+        self.matching_tolerance_hard = matching_tolerance_hard # DEF: 1e-1
```
maxbriel:
```suggestion
        self.matching_tolerance = matching_tolerance # DEFAULT: 1e-2
        self.matching_tolerance_hard = matching_tolerance_hard # DEFAULT: 1e-1
```

posydon/binary_evol/binarystar.py
```diff
@@ -144,8 +144,15 @@ def __init__(self, star_1=None, star_2=None, index=None, properties=None,
...
-        self.star_2 = star_2 if star_2 is not None else SingleStar()
+        if star_1 is not None:
+            self.star_1 = copy.deepcopy(star_1)
+        else:
+            self.star_1 = SingleStar()
+
+        if star_2 is not None:
+            self.star_2 = copy.deepcopy(star_2)
+        else:
+            self.star_2 = SingleStar()
```
maxbriel, line 155: Is this to make sure that the input star is not "used" and instead a copy is evolved?

Can you mention this change in the PR?

posydon/binary_evol/simulationproperties.py
```diff
@@ -130,14 +132,13 @@ def __init__(self, flow=({}, {}),
...
             step_class, _ = step_tuple
-            if isinstance(step_class, NullStep):
-                if verbose: Pwarn(f"Step {key} not provided, skipping it.",
-                                  "POSYDONWarning")
-                continue   
-            else:
-                self.kwargs[key] = step_tuple
+            if verbose and isinstance(step_class, NullStep): 
+                Pwarn(f"Step {key} not provided, skipping it.",
+                                "POSYDONWarning")
```
maxbriel: Is there a more relevant warning available? or create a new one? Like StepWarning or EvolutionWarning?
Not sure what name/type would be good.

posydon/popsyn/io.py
maxbriel, line 1: Why can this be removed? Shouldn't this information be important to include?

posydon/binary_evol/binarystar.py
sgossage: that's right -- so that the original memory reference is not accessed and altered by e.g., `BinaryStar.evolve()`. Just added this to the PR description.

posydon/popsyn/io.py
sgossage: These changes are due to a circular import where `popsyn/io.py` imports `SimulationProperties`, preventing `simulationproperties.py` from importing `simprop_kwargs_from_ini` from `popsyn/io.py`. (The new `SimulationProperties.from_ini()` function uses `simprop_kwargs_from_ini`.)

Now the properties are loaded in `BinaryPopulation.from_ini()` instead of in `popsyn.io.py` to avoid the circular import.

Removing the commented code in `io.binarypop_kwargs_from_ini()`.

posydon/binary_evol/simulationproperties.py
sgossage: Thanks, added a new `StepWarning`.

posydon/binary_evol/binarystar.py
astroJeff: Agree with Seth's code update

posydon/binary_evol/DT/gravitational_radiation/default_gravrad.py
```diff
@@ -2,8 +2,8 @@
 
 def default_gravrad(a, e, primary, secondary, verbose = False):
 
-    m1 = primary["mass"]
-    m2 = secondary["mass"]
+    m1 = primary.latest["mass"]
```
maxbriel, line 41: @sgossage I don't fully understand where this `latest` comes from. Could you add a doc-string to the function to explain what the input parameter and type is?

And where this equation comes from?

posydon/binary_evol/DT/magnetic_braking/prescriptions.py
```diff
@@ -4,15 +4,15 @@
 
 def RVJ83_braking(primary, secondary, verbose = False):
```
maxbriel, line 5: Can you add a doc-string?

posydon/binary_evol/DT/magnetic_braking/prescriptions.py
```diff
@@ -51,22 +51,22 @@ def RVJ83_braking(primary, secondary, verbose = False):
 
 def M15_braking(primary, secondary, verbose = False):
```
maxbriel, line 82: Doc string?

posydon/binary_evol/DT/magnetic_braking/prescriptions.py
```diff
@@ -133,21 +133,21 @@ def M15_braking(primary, secondary, verbose = False):
 
 def G18_braking(primary, secondary, verbose = False):
```
maxbriel, line 197: Another missing doc-string

posydon/binary_evol/DT/magnetic_braking/prescriptions.py
```diff
@@ -175,19 +175,19 @@ def G18_braking(primary, secondary, verbose = False):
 
 def CARB_braking(primary, secondary, verbose = False):
```
maxbriel, line 272: For consistency, also putting a comment for doc-string here.

posydon/binary_evol/DT/step_detached.py
```diff
@@ -433,38 +428,46 @@ def __call__(self, binary):
...
+                
+                abs_diff_porb_str = f"\nabs_diff_porb = {abs_diff_porb:.4f}" + \
+                    f"\nbinary.orbital_period = {binary.orbital_period:.4f}" +\
+                    "\norbital_period_from_separation(binary.separation, secondary.mass, primary.mass) = " + \
+                    f"{orbital_period_from_separation(binary.separation, secondary.mass, primary.mass):.4f}"
+
+                if self.verbose:
+                    print(abs_diff_porb_str)
+
+                assert abs_diff_porb < 1e-2, \
```
maxbriel, line 455: Do we actually want this to "always" be outputted?

posydon/binary_evol/DT/step_detached.py
```diff
@@ -798,6 +811,49 @@ def update_after_evo(self, res, t, binary, primary, secondary):
 
                 setattr(obj, key, current)
                 getattr(obj, key + "_history").extend(history)
+                
+    def update_co_stars(self, t, primary, secondary):
+
```
maxbriel, line 816:
```suggestion
```

posydon/binary_evol/DT/step_detached.py
```diff
@@ -1007,32 +1049,69 @@ def ev_rel_rlo2(self, t, y, primary, secondary):
...
+        self.primary = primary
+        self.secondary = secondary
+        # update list of tracked physical properties
+        self.phys_keys = list(secondary.interp1d.keys())
+        # except we don't evolve these:
+        for prop in ["t0", "m0", "t_max", "max_time"]:
+            self.phys_keys.remove(prop)
+
+        # dictionaries to track current properties during evolution
+        self.primary.latest = {}
```
maxbriel, line 1102: Should this be initialised as part of the `SingleStar` class, not just here?

posydon/binary_evol/DT/tides/default_tides.py
```diff
@@ -4,29 +4,29 @@
 
 def default_tides(a, e, primary, secondary, verbose=False):
```
maxbriel, line 5: doc-string

posydon/binary_evol/DT/winds/default_winds.py
```diff
@@ -2,27 +2,17 @@
 
 def default_spin_from_winds(a, e, primary, secondary, verbose = False):
```
maxbriel, line 3: doc-string

posydon/binary_evol/DT/winds/default_winds.py
```diff
@@ -73,11 +63,11 @@ def default_sep_from_winds(a, e, primary, secondary, verbose = False):
     wind loss.
```
maxbriel: Input parameters in doc-string

posydon/binary_evol/SN/step_SN.py
```diff
@@ -239,7 +240,28 @@ class StepSN(object):
...
+                       use_core_masses = SN_MODEL['use_core_masses'],
+                       allow_spin_None = SN_MODEL['allow_spin_None'],
+                       approx_at_he_depletion = SN_MODEL['approx_at_he_depletion'],
+                       kick = SN_MODEL['kick'],
+                       kick_normalisation = SN_MODEL['kick_normalisation'],
+                       sigma_kick_CCSN_NS = SN_MODEL['sigma_kick_CCSN_NS'],
+                       sigma_kick_CCSN_BH = SN_MODEL['sigma_kick_CCSN_BH'],
+                       sigma_kick_ECSN = SN_MODEL['sigma_kick_ECSN'],
+                       verbose = SN_MODEL['verbose'],
+                       **kwargs):
```
maxbriel: Is there a specific reason to expand it like this?

It increases the code length by quite a bit.

posydon/popsyn/population_params_default.ini
```diff
@@ -149,6 +153,8 @@
     # True, False
   do_magnetic_braking = True
     # True, False
+  magnetic_braking_mode = 'RVJ83'
```
maxbriel, line 156: Can you also add these to the documentation page? 
https://posydon.org/POSYDON/latest/components-overview/pop_syn/population_params.html

Similarly with the other additions in this PR.

file location: `docs/components-overview/pop_syn/population_params.rst`

posydon/binary_evol/SN/step_SN.py
sgossage: thanks -- at one point I ran into an error without giving the `kwargs` their defaults. I've tested it things work reverting back to the old `kwarg` handling and they do, so I've reverted the code to the more concise handling.

posydon/popsyn/population_params_default.ini
sgossage: Added new options to the docs.

posydon/binary_evol/DT/winds/default_winds.py
sgossage: Done!

posydon/binary_evol/DT/winds/default_winds.py
sgossage: Done!

posydon/binary_evol/DT/tides/default_tides.py
sgossage: Done!

posydon/binary_evol/DT/magnetic_braking/prescriptions.py
sgossage: Done!

posydon/binary_evol/DT/magnetic_braking/prescriptions.py
sgossage: Done!

posydon/binary_evol/DT/magnetic_braking/prescriptions.py
sgossage: Done!

posydon/binary_evol/DT/magnetic_braking/prescriptions.py
sgossage: Done!

posydon/binary_evol/DT/gravitational_radiation/default_gravrad.py
sgossage: Added doc strings!

The `latest` attribute is new. It is mostly a more convenient way to interact with the interpolators in `step_detached` compared to what we had before. In `detached_evolution.update_props()`, we get the interpolated value at the current time of integration `primary.interp1d[key](t)`. So, instead of having to refer to all that, those values are stored in the `latest` attribute of the primary/secondary for that time step at time `t`. This is meant to make it more clear/user-friendly what e.g., `primary.interp1d['mass'](t)` refers to in calculations during `step_detached`. Now you can just use `primary.latest['mass']` in calculations instead.

This info has been added to the doc string for `update_props()`.

maxbriel (approved):

PR #663, opened 2025-08-21
First steps: Validation binaries

maxbriel:
Here's a test script that creates and runs a bunch of binaries at solar metallicity and shows their output in an easily comparable format. Each system has a maximum of ~20 lines to occupy. 15 for the binary history. 5 for any warnings.

This is a first step towards validation tests. It's basic, but all I have time for at the moment.

I suggest people keep this in their working directory and test their branch with it before making a PR.


sgossage:
I suggest renaming `evolve_binaries.py` to something like `1Zsol_binaries_suite.py` or just `1Zsol_binaries.py`. I think it would be good to establish us having distinct/descriptive names for different tests for when we add more in the future.

I also suggest that we rename `population_params_default.ini` to something else, since that's the same as what we have stored in `posydon/synpop`. Maybe something like `test_population.ini`?

maxbriel:
I've made the requested changes :)
And here's a txt file with the output of the main branch.

[evolve_binaries_main.txt](https://github.com/user-attachments/files/21939475/evolve_binaries_main.txt)

astroJeff:
I think this is great - something we've been wanting to implement for some time now. The downside, of course, is that if we change things (e.g., the default evolution parameters) we will have to update the outputs for these binaries. That's exactly what we want to happen, after all this is a validation test by design. However, I think it's worth remembering that these outputs may need to be updated

sgossage:
one more suggestion -- what about having this as a subfolder inside of a directory called e.g., `dev-tools` so regular users have some immediate cue that the test suite is a more advanced tool and not something they need to worry about. If we develop similar tools/scripts in the future they could go in here.

sgossage (approved):

astroJeff (approved):

PR #662, opened 2025-08-21
Fixes to binarystar.py/binarypopulation.py for BinaryPopulation().evolve() functionality

sgossage:
This PR addresses [Issue#661](https://github.com/POSYDON-code/POSYDON/issues/661). Right now manually setting up a `BinaryPopulation` and running its `evolve()` function fails.

This is because of issues saving the `nearest_neighbour_column` and because `S1/S2` data is not saved by default to the `oneline` dataframe.

New stuff:
- `optimize_ram` is now a keyword that can be applied to `BinaryPopulation.evolve()`, it is `False` by default now to enable simple functionality similar to v1 with data frame output after evolve. It is set to `True` (the old default) in `PopulationRunner()` to maintain old behavior there.
- Three new columns are added to enable panda to parse the binary star data column `nearest_neighbour_distance` (`NN_dist_x`, `NN_dist_y`, `NN_dist_z`).
- Null values of `nearest_neighbour_distance` are now set to `np.nan` to enable easy float conversion, rather than the string `'None'` that it was before.

> [!Note]
> These changes should perhaps also be patched into `main`. Here's a PR for that: [PR#664](https://github.com/POSYDON-code/POSYDON/pull/664).

PR #654, opened 2025-08-14
Quickstart docs page

elizabethteng:
- first-populations.rst
- first-grids.rst
- minor fixes to other related docs pages

PR #664, opened 2025-08-21
[PATCH v2.0.2] Fixes to binarystar.py/binarypopulation.py for BinaryPopulation().evolve() functionality, Fix MB

sgossage:
> [!NOTE]  
> This is a patch for POSYDON v2.0.X, merging into `main`. For POSYDON v2.1.X, see here: [PR#662](https://github.com/POSYDON-code/POSYDON/pull/662). Right now the `meta.yml` is set to update to `v2.0.2`, anticipating that [PR#654](https://github.com/POSYDON-code/POSYDON/pull/654) will be accepted first.

- `optimize_ram` is now a keyword that can be applied to `BinaryPopulation.evolve()`, it is `False` by default now to enable simple functionality similar to v1 with data frame output after evolve. It is set to `True` (the old default) in `PopulationRunner()` to maintain old behavior there.
- Three new columns are added to enable panda to parse the binary star data column `nearest_neighbour_distance` (`NN_dist_x`, `NN_dist_y`, `NN_dist_z`). 
- Null values of `nearest_neighbour_distance` are now set to `np.nan` to enable easy float conversion, rather than the string `'None'` that it was before.
- Apparently the update to stellar spin from magnetic braking was inside the indent of a code block that would only trigger if `verbose = True` in `step_detached`. This has been corrected

### DIFF
--- a/conda/meta.yaml
+++ b/conda/meta.yaml
@@ -1,5 +1,5 @@
 {% set name = "posydon" %}
-{% set version = "2.0.0" %}
+{% set version = "2.0.2" %}
 
 package:
   name: "{{ name|lower }}"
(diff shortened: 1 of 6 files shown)


sgossage:
Tested and ran successfully for a multi metallicity population run of 1000 binaries each.

astroJeff (approved):

astroJeff (approved):

maxbriel (approved):

PR #670, opened 2025-08-26
Zero bad negative values from detached evolution

sgossage:
Sometimes you can get negative angular rotation rates for example from detached evolution (this also happens in `main`).

1. Adds a new `zero_negative_values()` function to `common_functions.py`. This will zero the negative values in an array to keep things positive.
2. Adds the options for match recording and verbosity to step initially single and step disrupted in our .ini file.
3. Now the match recording will use the current step name, rather than always use `step_detached`.
4. As a part of point 3, `BinaryStar` also given 3 new helper functions to facilitate getting the total state (s1.state, s2.state, binary.state, binary.event), next step name, and next step class object from its flow. I.e., `binary.get_total_state()`, `binary.get_next_step_name()` and `binary.get_next_step()`.

### DIFF
--- a/posydon/binary_evol/DT/step_detached.py
+++ b/posydon/binary_evol/DT/step_detached.py
@@ -27,7 +27,8 @@ from posydon.utils.common_functions import (bondi_hoyle,
                                             orbital_period_from_separation,
                                             roche_lobe_radius,
                                             check_state_of_star,
-                                            set_binary_to_failed)
+                                            set_binary_to_failed,
+                                            zero_negative_values)
 from posydon.binary_evol.flow_chart import (STAR_STATES_CC, 
                                             STAR_STATES_H_RICH_EVOLVABLE,
                                             STAR_STATES_HE_RICH_EVOLVABLE)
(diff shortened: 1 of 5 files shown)


sgossage:
I was running some tests and found this binary achieves a negative angular rotation rate after detached evolution (see `S2_surf_avg_omega`):

```
star_1 = SingleStar(**{'mass': 9.2, 'state': 'H-rich_Core_H_burning','metallicity':1.,
star_2 = SingleStar(**{'mass': 1.0 , 'state': 'H-rich_Core_H_burning','metallicity':1.,
                        'orbital_period':  183152.99061144073,'eccentricity': 0.0}, properties = sim_prop)
```

Screenshot 2025-08-26 170849

In this case, this is due to `do_stellar_evolution_and_spin_from_winds`.

sgossage:
after zeroing the values, the evolution looks like this:

image

astroJeff:
@sgossage I added a POSYDON warning that triggers when this replacement occurs. Can you check it?

sgossage:
Looks good, thanks @astroJeff

astroJeff (approved):

elizabethteng (approved):

PR #675, opened 2025-08-31
Overhaul detached step to be more customizable

sgossage:
This significantly restructures step_detached.

- A new `evolution_handler` class that is called by `solve_ivp`. This class holds certain star and orbital properties, updating every timestep, and calls physical prescriptions for tides, etc. during evolution
- Tides, magnetic braking, gravitational radiation, winds are all moved into their own files. This is meant to improve customizability of these prescriptions by users. and aid organization
- Other small changes, such as giving the `PchipInterpolator2` object an offset attribute that is applied once (to offset a matched track to the current age), rather than having to offset the time each timestep, or every time the interpolator is called.


astroJeff (approved):

sgossage:
tested and ran with out validation binaries, producing the same output as the base `v2.1-dev` branch.

astroJeff:
Pushing these changes. Bypassing merge rules...

PR #676, opened 2025-09-01
[PATCH v2.0.3] Fixes to binarypopulation.py population saving functionality

### LINKED ISSUE #667: Population running docs improvements
**Is your feature request related to a problem? Please describe.**

- In our population docs, we state ([here](https://posydon.org/POSYDON/latest/tutorials-examples/population-synthesis/pop_syn.html)) that we should see all 8 metallicities present in the copied default population params ini file, but only solar is present, It's in the comment for that setting, but not the actual value.
- We give some guidance on setting `dump_rate` for a small population but should either link to documentation describing the `dump_rate` in more detail/ or give some rules of thumb on how this should be set for an arbitrary number of binaries here. People could come where wanting to run something other than what the tutorial says.
- In [our FAQ](https://posydon.org/POSYDON/latest/troubleshooting-faqs/code-questions.html) regarding the `dump_rate` we say what happens if we set the value too low, but what about too high? We should provide guidance there, maybe a ballpark value typically used for 100, 1,000, 10,000, 100,000, 1,000,000,000 binaries? I'm envisioning a table...
- TO CHECK: it seems like `posydon-popsyn --help` hangs for several seconds and fails if $PATH_TO_POSYDON_DATA is not set. We should check whether it is running excess processes that are unnecessary if the `--help` flag is used.
- In the `posydon-popsyn` usage examples, `--job-array` is used as a flag in the Quest case when it should be `--job_array`.

**Describe the solution you'd like**
We should correct these issues for v2.X.X.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

sgossage:
This addresses several issues associated with the functionality of `breakdown_to_df = True`.

1. Previously, the population meta data was not stored when the population is saved with `PopulationManager.save()`. This causes errors when trying to read the file in e.g., via `Population('saved_population.h5')`The meta data has now been added in so that the file can be read without error (`ValueError: batches/evolution.combined.h5 does not contain an ini_parameters table!`).
2. Populations were previously saved as `.h5` files called `evolution.combined*` files, without the `.h5` extension. This meant that they could not be loaded by the `Population` class which expects the extension. Now these files are saved as `evolution.combined*.h5` files to remedy this.
3. Previously `breakdown_to_df = True` would cause `BinaryPopulation.manager` to try and save the population still. This produced a bad population file without meta data before. With the meta data in, another error because binaries are not stored in the `binaries` attribute of the manager -- instead they are broken into data frames and stored there. Now, if `breadkdown_to_df=True`, `BinaryPopulation` will not try to save the population to a file. Instead, the binaries are just kept in memory via data frames.
4. Updates to the documentation to reflect the addition of the `.h5` extension to `evolution.combined*` files and adds some more guidance regarding the `dump_rate`. Also clarifies text on usage of `breakdown_to_df`.
5. Explicitly defines the `kwargs` of `SimulationProperties` and dynamically fetches them. Doc string updated for this class to describe the kwargs.


astroJeff (approved):

sgossage:
tested and ran successfully with 8 metallicity run of 1k binaries each.

docs/_source/tutorials-examples/population-synthesis/pop_syn.ipynb
```diff
@@ -44,14 +44,21 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "Open the `population_params.ini` file and you should see all 8 currently supported metallicities present. \n",
+    "Open the `population_params.ini` file and you should see that the metallicity is currently set to solar (represented by a value of ``1.``), but in the comment, you can see all 8 currently supported metallicities. \n",
     "\n",
     "```\n",
-    "metallicity = [2., 1., 0.45, 0.2, 0.1, 0.01, 0.001, 0.0001]\n",
+    "metallicity = [1.] #[2., 1., 0.45, 0.2, 0.1, 0.01, 0.001, 0.0001]\n",
```
philipp-rajah, line 50: Only one metallicity now? Maybe need to change line 47 to reflect change?

philipp-rajah (approved):

docs/_source/tutorials-examples/population-synthesis/pop_syn.ipynb
astroJeff: The latest addresses this point

PR #677, opened 2025-09-01
[PATCH v2.0.3] Updating Installation Docs

### LINKED ISSUE #666: Installation guide improvements
Here are some things to update in the installation guide, from the perspective of a novice user going through the conda installation instructions:
- The installation guide says "This documentation describes the POSYDON v2.0.0 code which is not yet available on Anaconda.” and also says the v2 data are not available on Zenodo. Both of these are no longer true and should be updated. 
- It would also be helpful to update the reference to the "latest pre-release dataset (241028)” on the Quest/UNIGE clusters to 250520_newSNe
- It should be stated in the installation guide what the PATH_TO_POSYDON should be for the conda-installed version: `/path/to/your/conda/envs/<ENV_NAME>/lib/python3.11/site-packages`
- Instead of suggeting that users add the PATH_TO_POSYDON and PATH_TO_POSYDON_DATA to their bashrc, they should be saved to the env. This will set these environment variables every time the env is activated, but it means you can have different posydon/data installations in different envs: 
```
conda env config vars set PATH_TO_POSYDON=/path/to/your/posydon/installation
PATH_TO_POSYDON_DATA=/path/where/you/want/to/store/data
```

sgossage:
This adds some improvements to our installation docs from users. This also increments the version number to v2.0.3 and is intended to be patched in alongside [PR#676](https://github.com/POSYDON-code/POSYDON/pull/676).

This also updates out install documentation to be labeled as for the Stable version (v2.0.X right now). Previously it was labeled as the Development version, which was how it was in the past.

### DIFF
--- a/docs/_source/getting-started/installation-guide.rst
+++ b/docs/_source/getting-started/installation-guide.rst
@@ -83,8 +83,7 @@ Anaconda (Recommended)
     .. warning::
         Executing the following command ``get-posydon-data`` will download the full DR2 data set. This includes data for 
         all eight of the available metallicities, plus auxillary data. This may be more than you want, the data for each 
-        metallicity requires about 10 GB of disk space, so to be safe, make sure you have at least 103 GB of disk 
-        space available.
+        metallicity requires about 10 GB of disk space. In total, the DR2  dataset requires 103 GB of disk space.
 
     .. code-block:: bash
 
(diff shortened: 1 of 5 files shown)


astroJeff (approved):

sgossage:
tested and ran successfully with 8 metallicity run of 1k binaries each.

elizabethteng (approved):

docs/_source/getting-started/installation-guide.rst
```diff
@@ -57,68 +63,139 @@ Anaconda (Recommended)
...
     .. code-block:: bash
 
         pip install -e .
 
-3. **Set Environment Variables and Download Data**
+3. **Set Environment Variables**
+
+    Next, export the required paths as environment variables. From the ``POSYDON`` directory that you just performed the installation in, 
+    you can execute ``pwd`` if you are unsure of the path name. Please change the location names accordingly below to your installation 
+    path:
```
philipp-rajah, line 159: Is there no way we can make the path setting automatic? It's very unfriendly to some users

philipp-rajah (approved):

docs/_source/getting-started/installation-guide.rst
astroJeff: In the future we ought to consider a better way to deal with environmental variables. For now, this PR improves our documentation

PR #681, opened 2025-09-03
Fixing some typos from merge conflict resolution during rebase of v2.1

sgossage:
This PR just fixes some syntax errors introduced by merge conflict resolution during the rebasing of `v2.1-dev` with `main`.

- Indentation error on block assigning AML from magnetic braking in detached step
- Removes `binary_fraction` variable in `binarypopulation.py` as in the update to include the Moe & di Stefano q distribution.

This PR also removes a conditional import of `BinaryStar` in `binarypopulation.py` that appears to cause an error after the update to v2.0.3. Now `BinaryStar` is imported unconditionally. This should also be implemented in v2.0.3 or v2.0.4.

### DIFF
--- a/conda/meta.yaml
+++ b/conda/meta.yaml
@@ -1,5 +1,5 @@
 {% set name = "posydon" %}
-{% set version = "2.0.4" %}
+{% set version = "2.1.0" %}
 
 package:
   name: "{{ name|lower }}"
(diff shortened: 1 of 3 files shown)


sgossage:
successfully ran test multi metallicity population

astroJeff:
Looks good to me. Forcing this patch to be adopted - bypassing rules

PR #682, opened 2025-09-04
[PATCH v2.0.4] Fix conditional import

sgossage:
This makes `binarypopulation.py` always import `BinaryStar`. Previously it was conditional, producing an error, failing to import the needed module.


astroJeff (approved):

astroJeff:
Forcing this patch to be merged - bypassing rules

PR #684, opened 2025-09-04
[PATCH v2.0.5] General fixes for population saving, binary failed state storage, MESA grid single star user inlist functionality

### LINKED ISSUE #649: [BUG]: Single Star MESA inlist stacking
**Describe the bug**
I would like to use a custom MESA inlist on top of the posydon inlist. However, when trying to do this for a single star, I cannot stack the inlists. For the binaries, this works correctly, but not for a single star grid.

The code related to this is here: https://github.com/POSYDON-code/POSYDON/blob/5c1d02de3bf81e3823945c10016ac974dc50935a/bin/posydon-setup-grid#L263-L280

A special single-star inlist is made, instead of doing the stacking of different inlists parameters in the section before for a non-single star grid.
I'm not super familiar with this part of the code, so I could be getting things wrong in the setup. I've tried using the `star1_controls_user` parameter in the posydon inlist to get it to stack, but was unable to get it to add the custom parameters to the final inlist.

sgossage:
- Use the `history` instead of the `oneline` to get simulated mass, because the `oneline` does not always have masses in it. 
- Set default saving mode of `BinaryPopulation.PopulationManager.save` to `w`, not `a`.
- Also remove an extra `mode` keyword argument that is passed to `PopulationManager.save()`. This `kwarg` is searched for within the arbitrary `**kwargs`, so this extra kwarg prevents explicitly supplying your own `mode`.
- `PopulationManager.save` previously did not store the history columns, preventing functionalities like `loaded_population.history.select()` from working because there were no columns. Now the columns are saved and files saved this way can interacted with using e.g., `select()`.
- Appends the failed state to binary history in case of failed binaries.
- Fixes to the `overwrite` ability of `PopulationRunner.merge_parallel_runs()`.
- Prevents override of user supplied inlist controls in `posydon-setup-grid` when using `posydon` as the source

Tested with multi-metallicity population runs and sub-unity binary fractions to appropriately calculate and store the simulated mass.

### DIFF
--- a/posydon/binary_evol/binarystar.py
+++ b/posydon/binary_evol/binarystar.py
@@ -689,6 +689,9 @@ class BinaryStar:
                                             extra_binary_dtypes_user=extra_binary_cols_dict,
                                             extra_S1_dtypes_user=extra_s1_cols_dict,
                                             extra_S2_dtypes_user=extra_s2_cols_dict)
+        
+        # remove any columns duplicated from user keys/full star properties
+        oneline_df = oneline_df.loc[:,~oneline_df.columns.duplicated()].copy()
 
         return oneline_df
 
--- a/posydon/binary_evol/binarystar.py
+++ b/posydon/binary_evol/binarystar.py
@@ -632,8 +632,6 @@ class BinaryStar:
         """Convert binary into a single row DataFrame."""
         if history:
             bin_kwargs = kwargs.copy()
-            bin_kwargs['include_S1'] = False
-            bin_kwargs['include_S2'] = False
             output_df = self.to_df(**bin_kwargs)
             initial_final_data = output_df.values[[0, -1], :]  # first/last row
             col_names = list(output_df.columns)
--- a/posydon/popsyn/binarypopulation.py
+++ b/posydon/popsyn/binarypopulation.py
@@ -514,7 +514,7 @@ class BinaryPopulation:
         oneline_min_itemsize = {key: val for key, val in
                                 ONELINE_MIN_ITEMSIZE.items()
                                 if key in oneline_cols}
-        mode = kwargs.get('mode', 'a')
+        mode = kwargs.get('mode', 'w')
         complib = kwargs.get('complib', 'zlib')
         complevel = kwargs.get('complevel', 9)
         
--- a/posydon/popsyn/binarypopulation.py
+++ b/posydon/popsyn/binarypopulation.py
@@ -536,10 +536,10 @@ class BinaryPopulation:
                     oneline = pd.read_hdf(f, key='oneline')
                     
                     # split weight between single and binary stars
-                    ZAMS_mask = history["event"] == "ZAMS"
-                    singles_mask = history[ZAMS_mask]["state"] == "initially_single_star"
-                    filtered_data_single = history[ZAMS_mask][singles_mask]
-                    filtered_data_binaries = history[ZAMS_mask][~singles_mask]
+                    init_step_mask = ~history.index.duplicated(keep="first")  # indices that are NOT duplicates of the first
+                    singles_mask = history["state"] == "initially_single_star"
+                    filtered_data_single = history[init_step_mask & singles_mask]
+                    filtered_data_binaries = history[init_step_mask & ~singles_mask]
                     
                     simulated_mass_binaries += np.nansum(filtered_data_binaries[["S1_mass", "S2_mass"]].to_numpy())
                     simulated_mass_single += np.nansum(filtered_data_single[["S1_mass"]].to_numpy())
@@ -875,10 +875,10 @@ class PopulationManager:
             try:
 
                 # split weight between single and binary stars
-                ZAMS_mask = history_df["event"] == "ZAMS"
-                singles_mask = history_df[ZAMS_mask]["state"] == "initially_single_star"
-                filtered_data_single = history_df[ZAMS_mask][singles_mask]
-                filtered_data_binaries = history_df[ZAMS_mask][~singles_mask]
+                init_step_mask = ~history_df.index.duplicated(keep="first") # indices that are NOT duplicates of the first
+                singles_mask = history_df["state"] == "initially_single_star"
+                filtered_data_single = history_df[init_step_mask & singles_mask]
+                filtered_data_binaries = history_df[init_step_mask & ~singles_mask]
 
                 simulated_mass_binaries += np.nansum(filtered_data_binaries[["S1_mass", "S2_mass"]].to_numpy())
                 simulated_mass_single += np.nansum(filtered_data_single[["S1_mass"]].to_numpy())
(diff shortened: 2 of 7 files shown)


maxbriel:
@sgossage, we cannot guarantee that `ZAMS` is unique in the `history` for each binary; the `ZAMS` state can be repeated for erroneous binaries. This is why we're using the `oneline` to get the simulated mass. It might be better to either fix the masses as an always outputted part of the oneline or do some sort of filtering to make sure that for each binary only a single ZAMS line from the history is used.

sgossage:
thanks @maxbriel , originally I wanted to add the masses to the `oneline` but there are a few complications that I see:

1. The masses get added to the `oneline` from the single star data.
2. Right now this is done via an "all-or-nothing" approach, where if `include_S1/S2 = True`, all of the single star data is added, or else none of it is.
3. We have the option `only_select_columns` to select e.g., just the `mass` being added, but perhaps we don't want to hardcode always adding parameters like mass to leave the functionality of that `kwarg` working in a clear way for users. We would have to clarify something like "you can use this option to add only certain parameters, but mass will always be added". That could work but might be less clear and less flexible from a user's perspective, and seems to run counter to the intended function of this option. Albeit, in my opinion you always want to have the mass in the `oneline`, at least. This is close to what I tried to do at first.
4. I noticed that the doc string for `to_df` in `BinaryStar` sounds like its saying `include_S1/S2` should be `True` by default -- is that so? Maybe they got flipped at some point in the past accidentally, but right now we have them set to `False` in `to_oneline_df` (see comments below responding to Jeff). We could simply make them `True` if that won't cause issues with data footprints, etc. In fact, as is, it looks like these `kwargs` will never trigger if `history = True` for `to_oneline_df()` in `BinaryStar`. I'm not sure if that's intentional.
5. We could add mass always via the binary data `star_1_mass`, etc., but then we have the potential for repeat data in the `oneline` with masses always added from the binary data, and then possibly added again from the single star data.
6. We could add the masses to the `oneline` just for this calculation and then remove it, but that seems clunky.

For now, I added a more secure filter to ensure we just get the initial step's data from the history, let me know what you think of that solution, or if there might be a better way.

I also removed the setting of `include_S1 = False` and `include_S2 = False` in `BinaryStar.to_oneline_df()` for now. If that was intentional code let's put it back in though.

sgossage:
I added some updates to the docs on `binarypopulation.py` and fixed some typos. To reflect the existing docs, the final automatic save option during `evolve` will occur only if `breakdown_to_df == True`, if `breakdown_to_df == False`, the population will not be automatically saved.

The default setting for `breakdown_to_df` has now been set to `False`. It is set to `True` within the default population `.ini` file, so functionality remains the same when using that, but otherwise this option must be enabled manually if running a population without the `.ini` file.

Added `kwargs` retrieval to `to_df()`, `to_oneline_df()` and `save()` so that `BinaryPopulation` will use its supplied kwargs when these are called outside of `evolve()`.

posydon/binary_evol/binarystar.py
```diff
@@ -632,8 +632,6 @@ def to_oneline_df(self, scalar_names=[], history=True, **kwargs):
         """Convert binary into a single row DataFrame."""
         if history:
             bin_kwargs = kwargs.copy()
-            bin_kwargs['include_S1'] = False
-            bin_kwargs['include_S2'] = False
```
astroJeff, line 636: Did we remove this functionality?

posydon/popsyn/binarypopulation.py
```diff
@@ -48,7 +47,7 @@
                                                binary_fraction_value)
 from posydon.popsyn.sample_from_file import (get_samples_from_file,
                                              get_kick_samples_from_file)
-from posydon.popsyn.normalized_pop_mass import initial_total_underlying_mass
```
astroJeff, line 51: Why was this removed?

sgossage: This module is unused in ``binarypopulation.py`` It's been this way since [PR#451](https://github.com/POSYDON-code/POSYDON/commit/92b0ae061081cdd821258e8f19b769a01b051ec4). I'm not sure when it was ever used.

posydon/binary_evol/binarystar.py
sgossage: I removed these because I'm not sure that they are intentional, but awaiting further input. It seems like these have existed since at least [PR#451](https://github.com/POSYDON-code/POSYDON/commit/92b0ae061081cdd821258e8f19b769a01b051ec4), but these codelines disable the usage of these `kwargs` if `history = True` in `BinaryStar.to_oneline_df()`. The doc strings elsewhere in `binarystar.py` seem to suggest that these should be `True` by default, and we set these to `True` both in our default `.ini` file and by default if this `kwarg` is unset. However, because we set these to `False` here, the `kwarg` is essentially always hardcoded to `False`, hence why I removed these lines for now. 

Instead `only_select_columns` supply specific column names to add to the `oneline_df` in our `.ini` file. If these `include_S1`/`include_S2` `kwargs` are deprecated, perhaps we should remove them all together.

sgossage: Right now removing the two lines above introduces an error w/ duplicated column names in the oneline df in certain cases -- looking into a solution.

sgossage: This happens because with those lines removed, we add all of the star data to the `oneline` once in lines `633-649` ( in `binarystar.py`, if `include_S* = True`), and then again if we also supplied `only_select_columns` (which we do in the `.ini` file).

The `only_select_columns` data are added in lines `657-669`. I added a line to remove duplicate columns in case both of these arguments (`include_S*` and `only_select_columns`) are used together and all seems to be working fine with the latest commit. This means that if you set `include_S* = True`, then all the star data gets added (as intended) and if you also set `only_select`,  any dupes added by it are eliminated.

I can also set the default of `include_S1/include_S2` to `False` in the `.ini` file, since it seems like the intention there is to only include a select few columns from the star data, rather than all of them (or we should remove the `only_select_columns` and just leave `include_S* = True`).

posydon/popsyn/binarypopulation.py
astroJeff: All sounds good - just checking

posydon/binary_evol/binarystar.py
astroJeff: Thanks for tracking this down. I suppose from a coding standpoint, it's easier to remove the duplication afterwards rather than just not add the new column if it already exists. I don't think the copy command will bloat the memory too much, but we should keep this in mind if we run into memory issues in the future, at the save step.

astroJeff (approved):

maxbriel (approved):
Thank you for the additional filtering and fixing the simulated mass.
This looks good to me now.

I am wondering if it's worth (at a later date) to create more abstract versions of the "saving/merging".

sgossage:
> I am wondering if it's worth (at a later date) to create more abstract versions of the "saving/merging".

Yes I agree -- I think we have a fair amount of overlap at the moment and can probably abstract a few methods that can be applied generically where needed.

PR #685, opened 2025-09-08
Additional kick options

maxbriel:
Small rework of the kick prescriptions and normalisations, allowing for more flexibility of the input parameters of the distributions.

1. It splits up `kick_normalisation` and `kick_prescription`. I needed to have a "one" normalised kick, but with "log_normal" drawn. This is not currently possible.
2.  Adds new parameters to step_SN in the population ini file.
    - 'kick_prescription'
    - `mean_kick_XXXX` for `BH`, `NS`, and `ECSN`
3. If `asym_ej` or `linear` are still used as part of the `kick_normalisation` parameters, they're automatically set as the 'kick_prescription' to allow older inlists to still be used. A DeprecationWarning is thrown as well.

I'm interested in hearing if we should keep these functions in the step_SN class or move them out, since they're basically independent.


maxbriel:
superseded by #717

PR #686, opened 2025-09-10
[v2.1.3] Initialize `SingleStar` properties

### LINKED ISSUE #354: Evolving a binary from an arbitrary combination

### LINKED ISSUE #577: Initialize binaries with all parameters set to null values

sgossage:
Initialize star properties with null values.

In the course of `BinaryPopulation.evolve()`, these properties are initialized, however when evolving a `BinaryStar` outside of that framework, things can fail because the current `None` null value causes trouble with math operations, so this introduces `np.nan` where possible/appropriate as the new null value. So, if a user isn't careful to initialize certain things in `SingleStar` that are required by the track matching process, the evolution is likely to fail. This is moreso the case when creating binaries where the stars are either initially CO or evolved stars.

This is a cherry picked remake of [PR#613](https://github.com/POSYDON-code/POSYDON/pull/613).

- Implements `SingleStar` initialization in cases where the star's state is not `H-rich_Core_H_burning`. `SingleStar.__init__()` now sets initial values for `SingleStar` that can be used by `track_matcher` to find an initial model. Previously if e.g., `center_h1`, `center_he4`, `log_R`, `he_core_mass` were not set, the star could not match to an initial model and could not be evolved.
- Many of the `float64` data type star properties are now initialized with `np.nan`, rather than `None` to avoid arithmetic `TypeError` occuring.
- The unit tests had to to be modified because of the new null values -- previously related tests checked for `None`, which is no longer relevant.
- It is now possible to evolve a `BinaryStar` with stars from arbitrary states. like `NS`, `stripped_He_Core_He_burning`. However, it is possible for a user to give initial conditions for a CO that does not fit within our internal classifications for COs. The CO type is inferred at initialization, given the initial conditions and POSYDON's inferred state replaces the user's given state. This is to avoid e.g., a WD becoming a NS midway through evolution, simply because POSYDON checks the state and reassigns it. TODO: warn that this happens.
- Previously, the `_history` of `SingleStar` and `BinaryStar` properties were not properly updated during `__init__`. This has been fixed for properties in `STARPROPERTIES` and `BINARYPROPERTIES` and for those properties supplied by a user.
- In `step_mesa` and `step_detached`, most property histories of a CO would be reset to `None`. Now just the old values from the last step are used instead (which are often null values anyways, unless modified by a custom hook in a step).

### DIFF
--- a/posydon/binary_evol/MESA/step_mesa.py
+++ b/posydon/binary_evol/MESA/step_mesa.py
@@ -662,6 +662,8 @@ class MesaGridStep:
                             getattr(star, key_h).append(cb_bh[key_p][0])
                         if track_interpolation:
                             getattr(star, key_h).extend(cb_bh[key_p][:-1])
+                    elif key in ['state', 'lg_mdot']:
+                        continue
                     else:
                         if self.save_initial_conditions:
                             getattr(star, key_h).append(old_h[0])
@@ -915,6 +917,8 @@ class MesaGridStep:
                     elif key in ['lg_mdot', 'lg_system_mdot', 'lg_wind_mdot']:
                         key_p = POSYDON_TO_MESA['star'][key]+'_%d' % (k+1)
                         setattr(star, key, fv[key_p])
+                    elif key == 'state':
+                        continue
 
         # infer stellar states
         interpolation_class = self.classes['interpolation_class']
--- a/posydon/binary_evol/MESA/step_mesa.py
+++ b/posydon/binary_evol/MESA/step_mesa.py
@@ -465,7 +465,7 @@ class MesaGridStep:
             length_binary_hist = len(cb_bh['age']) - 1
             length_star_hist = len(cb_hs[0]['center_he4']) - 1
             length_hist = min(length_binary_hist, length_star_hist)
-            empy_h = [None] * length_hist
+            empty_h = [None] * length_hist
             MESA_history_bug_fix = False
             if length_binary_hist != length_star_hist:
                 MESA_history_bug_fix = True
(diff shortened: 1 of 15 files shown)


sgossage:
In this PR, one can successfully evolve the binary in [Issue#354](https://github.com/POSYDON-code/POSYDON/issues/354)

image

posydon/binary_evol/MESA/step_mesa.py
```diff
@@ -656,8 +656,9 @@ def update_properties_NN(self, star_1_CO=False, star_2_CO=False,
                             getattr(star, key_h).extend(cb_bh[key_p][:-1])
                     elif key in ['state', 'lg_mdot']:
                         continue
-                    elif star.state == 'WD' and key in ['co_core_mass','he_core_mass','center_h1','center_he4','center_c12','center_n14','center_o16']:
-                        continue
```
sgossage, line 665: I am not sure that we want to fully remove this, but if you set up a `BinaryStar` where one of the stars is initially a WD, this code prevents the WD's history from being updated for these properties which prevents evolution, so it needs to be handled some way. I did not see harm in removing it, but please double check this.

sgossage:
These are some examples of a 10 solar mass star initialized using different initial states (sorry for the small font). In all cases, only an initial state and mass is given:

```
'H-rich_Core_H_burning',
'H-rich_non_burning',
'H-rich_Core_He_burning',
 'accreted_He_Core_H_burning',
'accreted_He_Shell_H_burning',
 'accreted_He_non_burning',
'stripped_He_Core_He_burning',
'stripped_He_non_burning',
'stripped_He_Core_He_depleted'
```

the `accreted_He_*` models are all like the `H-rich_*` models because in our matching process we do not use any surface values to determine which star to match to. In one case there's a bad match where the 10 Msol star is matched to a 17 Msol star. This is because we are making some simple guess right now about what the initial R is in a given state, for a given mass, so the initial match can be off by quite a bit if only the mass is given. There are better ways to get an initial guess perhaps, but this is the simplest solution perhaps for now. Although, if people have suggestions, please say so and it can be implemented/tested.

Screenshot 2025-09-17 235818

astroJeff (approved):

maxbriel (approved):

PR #687, opened 2025-09-10
[v2.0.6] Rescue script fix

maxbriel:
The rescue script was expecting the temporary batch files to end on their 'run' number. However, with PR #676 the `.h5` extension is added. This patch makes the rescue script grab the correct index again.


astroJeff (approved):

astroJeff:
Easy bug fix. Thanks for putting together, @maxbriel. Bypassing rules and merging

PR #699, opened 2025-09-19
Fix: posydon_github_root to skipped parameters

maxbriel:
I want to be able to use this parameter as part of the `mesa_extras` in the grid configuration files.
This can either be achieved by placing this parameter also in the `[mesa_extras]` section or creating a `[DEFAULT]` section and placing `posydon_github_root` there. 

Either way, `posydon_github_root` becomes part of the `[mesa_extras]` section and becomes one of the keys. It should be skipped, when one of the keys.

This is not the best fix, but at least allows `posydon_github_root` to be globally defined (since it's used in multiple section in the configuration file).


maxbriel:
I found a work around. You can refer to `${mesa-inlists:posydon_github_root}`. We can close this PR.

PR #700, opened 2025-09-21
[2.1.4] Fix massless remnant metallicity assignment

sgossage:
Massless remnants are given a null metallicity in BinaryPopulation which causes the new SingleStar initialization to fail. This edits the initialization to skip the check on metallicity for compact objects. It also adds more safeguards to make sure steps pick a metallicity.


astroJeff (approved):

astroJeff:
Bypassing rules to merge

PR #708, opened 2025-09-30
add pre-commit workflow for checks+isort and add to CI

maxbriel:
Testing out a pre-commit for basic checks and isort to fix imports.

This is opt-in and not on by default:
Developers can do `pip install '.[dev]'` to install the required packages.
Then with `pre-commit install`, you can initialize `pre-commit` to run before every commit you do.

This checks the following:
1. check for added large files 
2. check python ast
3. check docstring is first
4. check illegal windows names
5. check toml
6. debug statements (python)
7. fix end of files **consistent end of files**
8. fix requirements.txt
9. trim trailing whitespace **removes training whitespaces**
10. isort **this orders the imports in python files**

It's only been ran on the files in this PR, not on the general codebase.
We can include `pre-commit.ci` into the repository if we want automatic fixes on PRs.


maxbriel:
Additional pre-commit hooks: https://github.com/pre-commit/pre-commit-hooks

I was trying to figure out if it's useful including: `double-quote-string-fixer`.
We're currently using a mix of double and single quotes throughout our code. This would clean it up and make them all single quotes.

maxbriel:
I've added pre-commit to the POSYDON repo and run it on all the files in the repository to clean them up.
I've had to make a few manual changes to multi-line string comments being identified as module docstrings. And a debug breakpoint was present in the VHdiagrams.
For now, we do not automatically change the double quotes into single quotes, because we're unsure if this causes behaviour issues.

Once accepted, this will run `pre-commit.ci` on every PR, and make sure the basic code layout checks are done.
**I suggest we merge this into `main` and then propagate this to `v2.2-dev` immediately.**

I checked on an online rST text editor that when there's no whitespace at the end of a line, it automatically adds a space in the rendered file. Example: https://rsted.info.ucl.ac.be/?theme=basic&n=299bcfb0feb2e1f9cf343aaae14ab248

astroJeff (approved):

sgossage (approved):

PR #710, opened 2025-09-30
[feat] SHA for test binaries

sgossage:
This adds the ability to supply a SHA to run our test suite of binaries with. So, now you can do

```
./evolve_binaries.sh main c09b3260dc451692dc9ed3a6de595f9ed9e6d246
```

for example to run with/compare to that specific commit on `main`. 

Some issues:

 + Since the `1Zsun_binaries_suite.py` script uses certain options to evolve the binaries, there are no guarantees currently that the commit you select will be compatible with those options and the script may fail to run.
+ I have seen that if you select especially old commits (e.g., pre v2), POSYDON installation may fail.

### DIFF
--- a/dev-tools/evolve_binaries.sh
+++ b/dev-tools/evolve_binaries.sh
@@ -8,8 +8,12 @@
 BRANCH=${1:-main}
 REPO_URL="https://github.com/POSYDON-code/POSYDON"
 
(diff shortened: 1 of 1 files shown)


maxbriel (approved):
LGTM!

PR #712, opened 2025-10-02
[fix] `BinaryStar.to_df()` for failed binaries

sgossage:
Right now, failed binaries are not properly being broken down and saved in `BinaryStar.to_df()`. Often, this error is thrown:

image

The reason why is that if a user has set `only_select_columns` (as is the case by default in our population .ini file), in the case of a failed binary, the method `equalize_columns` (from `bianrystar.py`) would be called to append null values and create equal length columns. However, all binary property keys would be passed to this function, meanwhile, `data_to_save` is cut to include data only according to `only_select_columns`, so the keys array and the data array are mismatched in this case.

This PR just uses the `keys_to_save` array, which was already in the code, which is cut according to `only_select_columns`, instead of all binary property keys, eliminating this error.


maxbriel (approved):
LGTM!

maxbriel (approved):
I merged v2.1.8 back into this branch for you, @sgossage. Changes should be the same as before. The changes look good to me

PR #713, opened 2025-10-03
[fix] grid bound checks in `step_mesa`

sgossage:
We have several checks to make sure the star masses are inside the grid bounds in the case where the star order is flipped (star 1 becomes star 2). So, e.g.,

```
        elif ((self.p_min <= p <= self.p_max and
               (m1 < self.m1_min or m1 > self.m1_max))):
            set_binary_to_failed(self.binary)
            raise GridError(f'The mass of m1 ({m1}) is outside the grid,'
                                ' while the period is inside the grid.')
```

becomes

```
        elif ((self.p_min <= p <= self.p_max and
               (m2 < self.m1_min or m2 > self.m1_max))):
            set_binary_to_failed(self.binary)
            raise GridError(f'The mass of m1 ({m1}) is outside the grid,'
                                ' while the period is inside the grid.')
```

and we we never flip the grid bounds (`self.m1_min` or `self.m1_max`, for example). This causes binaries to fail when they should not. For example, the binary

```
star_1 = SingleStar(**{'mass': 4.118113, 'state': 'H-rich_Core_H_burning','metallicity':1,
            'natal_kick_array':  [None, None, None, None]})
star_2 = SingleStar(**{'mass': 3.221320, 'state': 'H-rich_Core_H_burning','metallicity':1,
                        'orbital_period':  3.808629,'eccentricity': 0.0}, properties = sim_prop)
```

fails because the primary becomes a 0.59 Msol WD, is flipped in `step_CO_HMS_RLO`, and is checked against the grid bounds of `m2_min` and `m2_max` for the CO_HMS grid. These bounds are for the normal (HMS) star, which are inappropriate for the CO; the minimum HMS mass in this grid is 1 Msol, while the minimum CO mass is 0.5. The check should still be against `m1_min` and `m1_max`, so this PR removes the flipped star grid bound checks to avoid these failures.


sgossage:
This is the original evolution of the binary mentioned in the PR description. You can see that it fails unnecessarily because it says that the WD mass of 0.59 is outside of the grid bounds. It is not, the minimum CO mass is 0.5 Msol in our grids.

Screenshot 2025-10-02 222224

sgossage:
This is the evolution after the fix in this PR. The binary still fails now, but it fails correctly because the HeMS star is 0.43 Msol, which is outside the grid bounds (the minimum He star mass in our grids is 1 Msol).

Screenshot 2025-10-02 223057

maxbriel:
@sgossage  I updated the branch. 

I don't fully get the reason that we can remove the check. Can you provide a bit more basic explanation?

I thought that we hadn't yet flipped the stars when entering this check. So if `flip_star_before_step=True`, we need to check m1 vs m2 limits for the grid.
Is the problem here that for the CO-HMS grid, the values are the other way around? ie. m1=star and m2=CO?
Why should we also apply this to the HMS-HMS grid?

sgossage:
Thanks @maxbriel for the update.

The basic issue I see is that the grid bounds are fixed, so for example, we should always check the CO star against `m1` bounds, since `m1` is always the CO star in these grids.

The grid bounds are always:

Screenshot 2025-10-06 111316

Right now, we have a check that says if `state_1` is a CO state (which should be the case in normal evolution when entering `step_CO_Star_RLO`), set `flip_stars_before_step = True`, which then causes the CO mass to be checked against the `m2` bounds, which is inappropriate.

The same issue applies to the HMS-HMS grid. Unless the `binary.star_*` memory references have actually been flipped, we should not check e.g., `star_2` against `m1` bounds. I think the only time that these references could be flipped when entering `step_MS_MS` is if the user explicitly flipped them (set star 1 to be less massive than star 2 for example). 

We might consider keeping the check there to allow for that functionality, or in case `star_2` forms a CO first, maybe from reverse MT. Likewise, we could keep the `CO_Star` grid checks, but will need to modify the logic because it currently leads to the above error.

In any case, I do not think we want e.g., this logic:

```
        elif (state_1 in ["WD", "NS", "BH"] and (state_2 in FOR_RLO_STATES)
                and event == "oRLO2"):
            self.flip_stars_before_step = True
```

It is almost always true, and therefore almost always makes the CO star check against the wrong grid bounds.

astroJeff:
@sgossage one thing we need to consider with this is what happens when we add a HeMS-HMS grid? We will need to include some functionality to make sure things are flipped correctly. Will we have to re-create the changes you're suggesting?

sgossage:
@astroJeff @maxbriel , this is no longer an issue. Somehow I convinced myself that the grid bounds were reversed, but they are as they should be.

PR #714, opened 2025-10-03
[feat] New `posydon-setup-env` command

sgossage:
This adds a command `posydon-setup-env` that takes care of setting up a `conda` environment with POSYDON.

TODO: Intending to rework this and make it more functional for the use case where a user has already installed POSYDON with conda or git and wants to setup a conda environment with that installation. Also needs unit test coverage to be addressed.


posydon/utils/install.py
```diff
@@ -0,0 +1,102 @@
+import os
+import subprocess
+from pathlib import Path
+
+HOME = Path.home()
+MINICONDA_DIR = HOME / "miniconda"
+MINICONDA_SCRIPT = HOME / "miniconda.sh"
+ENV_NAME = "posydon_env"
+DATA_DIR = "/projects/e33022/POSYDON-shared/data"
```
astroJeff, line 9: Can we come up with a user input here?

posydon/utils/install.py
```diff
@@ -0,0 +1,102 @@
...
+
+def run(cmd, check=True, shell=True, **kwargs):
+    """Run a shell command."""
+    print(f"Running: {cmd}")
+    subprocess.run(cmd, shell=shell, check=check, **kwargs)
+
+def install_miniconda():
+    if not MINICONDA_DIR.exists():
+        print("📦 Installing Miniconda...")
+        run(f"curl https://repo.anaconda.com/miniconda/Miniconda3-latest-Linux-x86_64.sh --output {MINICONDA_SCRIPT}")
```
astroJeff, line 22: This is specific to Linux. We should run a test for other OS's

posydon/utils/install.py
```diff
@@ -0,0 +1,102 @@
...
+            lines = f.read().splitlines()
+        if conda_cmd not in lines:
+            with open(bashrc, "a") as f:
+                f.write(f"\n{conda_cmd}\n")
+
+def setup_ipython_config(posydon_dir):
+    if IPYCONFIG_PATH.exists():
+        IPYCONFIG_PATH.unlink()
+    run("ipython profile create")
+    tex_path = "/software/texlive/2020/bin/x86_64-linux/"
```
astroJeff, line 77: Same here - check for different OS's

astroJeff:
Online documentation is probably good enough for installation by new users. Closing.

PR #715, opened 2025-10-03
[fix] CE re-adjusting the inflated radius of the companion

kasdaglie:
This PR introduces the adjust_secondary_radius function to adjust the inflated radius of the companion.

 It uses a TrackMatcher object to find the closed single star track and returns the radius at the time of the CE. 

! This PR introduces metallicity to step CE -> Tutorials and scripts for evolving a single binary will have to be changed


kasdaglie:
@astroJeff

sgossage:
Thank you @kasdaglie , is it possible to provide a screenshot or two of example binary star evolution before and after this fix?

astroJeff:
@maxbriel is the pre-commit.ci error being thrown because Eirini needs to merge the latest version into her branch?

maxbriel:
@astroJeff That's correct. @kasdaglie I've updated the branch with v2.1.8

kasdaglie:
Screenshot of the same binary. It merges when it's evolved with the main branch and continues as detached with the radius re-adjustment for another 16 Myr before failing due to He-HMS interaction. 
Screenshot 2025-10-09 at 9 25 35 PM

sgossage:
@kasdaglie , I made an edit to `simulation_properties.py` to remove `step_CE` from the list of steps that are ignored when setting a step's metallicity.

posydon/binary_evol/CE/step_CEE.py
```diff
@@ -141,6 +148,7 @@ def __init__(
             CEE_tolerance_err=MODEL['CEE_tolerance_err'],
             mass_loss_during_CEE_merged=MODEL['mass_loss_during_CEE_merged'],
             verbose=MODEL['verbose'],
+            metallicity = MODEL['metallicity'],
```
sgossage, line 152:
```suggestion
            metallicity = MODEL['metallicity'],
            record_matching = MODEL['record_matching'],
```

posydon/binary_evol/CE/step_CEE.py
```diff
@@ -61,11 +68,11 @@
          "CEE_tolerance_err": 0.001,
          "verbose": False,
          "common_envelope_option_after_succ_CEE": 'two_phases_stableMT',
-         "mass_loss_during_CEE_merged": False # If False, then no mass loss from this step for a merged star
+         "mass_loss_during_CEE_merged": False, # If False, then no mass loss from this step for a merged star
                                               # If True, then we remove mass according to the alpha-lambda prescription
                                               # assuming a final separation where the inner core RLOF starts.
          # "one_phase_variable_core_definition" for core_definition_H_fraction=0.01
-
+         "metallicity": None,
```
sgossage, line 75:
```suggestion
         "metallicity": None,
         "record_matching": False
```

posydon/binary_evol/CE/step_CEE.py
```diff
@@ -167,10 +175,31 @@ def __init__(
             self.common_envelope_option_after_succ_CEE = \
                 common_envelope_option_after_succ_CEE
             self.mass_loss_during_CEE_merged = mass_loss_during_CEE_merged
-
+        self.metallicity = metallicity
```
sgossage, line 180:
```suggestion
        self.metallicity = metallicity
        self.record_matching = record_matching
```

posydon/binary_evol/CE/step_CEE.py
```diff
@@ -167,10 +175,31 @@ def __init__(
...
+                                    grid_name_strippedHe = None,
+                                    path=PATH_TO_POSYDON_DATA,
+                                    metallicity = metallicity,
+                                    matching_method = "minimize",
+                                    matching_tolerance=1e-2,
+                                    matching_tolerance_hard=1e-1,
+                                    list_for_matching_HMS = list_for_matching_HMS,
+                                    list_for_matching_HeStar = None,
+                                    list_for_matching_postMS = None,
+                                    record_matching = False,
```
sgossage:
```suggestion
                                    record_matching = self.record_matching,
```

posydon/binary_evol/CE/step_CEE.py
```diff
@@ -167,10 +175,31 @@ def __init__(
...
+        list_for_matching_HMS = [
+                        ["mass", "center_h1", "he_core_mass"],
+                        [20.0, 1.0, 10.0],
+                        ["log_min_max", "min_max", "min_max"],
+                        [0.1, 300], [0.0, None]
+                    ]
+        self.track_matcher = TrackMatcher(grid_name_Hrich = None,
+                                    grid_name_strippedHe = None,
+                                    path=PATH_TO_POSYDON_DATA,
+                                    metallicity = metallicity,
```
sgossage:
```suggestion
                                    metallicity = self.metallicity,
```

posydon/binary_evol/CE/step_CEE.py
```diff
@@ -1408,3 +1440,60 @@ def CEE_adjust_binary_upon_merger(self, binary, donor, comp_star, m1_i,
...
+            elif comp_star.state in STAR_STATES_HE_RICH:
+                comp_star.htrack = False
+            else:
+                raise ValueError("state = %s of donor of CEE not recognized"
+                           % comp_star.state)
+            if self.verbose:
+                print(f"The binary overfilled its RL due to the companion's inflated radius \n"
+                      f"The Roche lobe of the companion is: {RL2}, The radius is: {rc2_i}")
+            t_i = binary.time
+            m0, t0 =self.track_matcher.get_star_match_data(binary, comp_star)
```
sgossage:
```suggestion
            _, _, only_CO = self.track_matcher.do_matching(binary, step_name='step_CE')
            t0 = comp_star.interp1d['t0']
```

This will let us see the matching info if `record_matching = True`

posydon/binary_evol/CE/step_CEE.py
```diff
@@ -1408,3 +1440,60 @@ def CEE_adjust_binary_upon_merger(self, binary, donor, comp_star, m1_i,
...
+            rc2_i = 10**comp_star.interp1d['log_R'](t_i)
+            print('Time,radius of sec:',t0,rc2_i,comp_star.interp1d['log_R'](t_i))
```
sgossage: Is this `print` needed, or debugging info? Can it be moved into the `verbose` block below?

posydon/binary_evol/CE/step_CEE.py
sgossage: We should also probably add the ability to alter the other options of the matching, like the method, etc., as we have for `step_detached`.

sgossage: We should also add the `record_matching` option to `step_CE` in our default `.ini` file for users to turn on/off from there.

sgossage: Additionally, we can add something to allow you to only match one star (in this case the companion) and not both.

kasdaglie: I initially used the `do_matching` function but I ended up changing it. The reason I use `get_star_match_data` is because this only matches the companion and not the entire binary and ends up being more computationally efficient. I do agree that the `record_matching` option is a better way to do this because now the verbose for the CE prints all the verbose for the matching as well, and it's not ideal for someone that doesn't care about the matching information.

posydon/binary_evol/CE/step_CEE.py
kasdaglie: It was used for the debugging purposes, I forgot to delete it.

posydon/binary_evol/CE/step_CEE.py
sgossage: Sounds good. I pushed an update to allow one or the other star to be matched with `do_matching` (so it does not always have to be both.

posydon/binary_evol/DT/track_match.py
```diff
@@ -1683,22 +1701,25 @@ def update_rotation_info(self, primary, secondary):
...
                 primary.surf_avg_omega_div_omega_crit = 0.0
                 primary.surf_avg_omega = 0.0
                 primary.log_total_angular_momentum = 0.0
-        else:
-            # omega of compact objects or massless remnant
-            # (won't be used for integration)
-            omega0_pri = omega0_sec
+
 
         return omega0_pri, omega0_sec
```
kasdaglie, line 1724: These new changes have introduce the error: 
`UnboundLocalError: cannot access local variable 'omega0_pri' where it is not associated with a value`

sgossage: thanks @kasdaglie , I accidentally removed an `else` statement in the last commit. I put it back -- can you confirm that the above issue no longer occurs?

kasdaglie: The problem has been fixed @sgossage, thanks!

sgossage (dismissed):

posydon/binary_evol/CE/step_CEE.py
```diff
@@ -167,10 +177,32 @@ def __init__(
...
             self.mass_loss_during_CEE_merged = mass_loss_during_CEE_merged
-
+        self.metallicity = metallicity
+        self.record_matching = record_matching
         self.verbose = verbose
         self.path_to_posydon = PATH_TO_POSYDON
 
+
+        #m_min_H = np.min(self.grid_Hrich.grid_mass)
+        #m_max_H = np.max(self.grid_Hrich.grid_mass)
```
maxbriel:
```suggestion
```

posydon/binary_evol/CE/step_CEE.py
```diff
@@ -1408,3 +1443,73 @@ def CEE_adjust_binary_upon_merger(self, binary, donor, comp_star, m1_i,
...
+            elif comp_star.state in STAR_STATES_HE_RICH:
+                comp_star.htrack = False
+            else:
+                raise ValueError("state = %s of donor of CEE not recognized"
+                           % comp_star.state)
+            if self.verbose:
+                print(f"The binary overfilled its RL due to the companion's inflated radius \n"
+                      f"The Roche lobe of the companion is: {RL2}, The radius is: {rc2_i}")
+            t_i = binary.time
+            #m0, t0 =self.track_matcher.get_star_match_data(binary, comp_star)
```
maxbriel:
```suggestion
```

posydon/binary_evol/CE/step_CEE.py
```diff
@@ -930,6 +962,9 @@ def CEE_simple_alpha_prescription(
             print("separation_i in Rsun", separation_i/const.Rsun)
             print("separation_postCEE in Rsun", separation_postCEE/const.Rsun)
 
+        if ((not double_CE) and (comp_star.state not in STAR_STATES_CO)):
```
maxbriel, line 963: Do we really want this always to run when a CE occurs with a stellar companion?

Or do we only want to apply it if the systems would have merged otherwise?

maxbriel (dismissed):

posydon/binary_evol/CE/step_CEE.py
```diff
@@ -1408,3 +1441,72 @@ def CEE_adjust_binary_upon_merger(self, binary, donor, comp_star, m1_i,
...
+                comp_star.htrack = False
+            else:
+                raise ValueError("state = %s of donor of CEE not recognized"
+                           % comp_star.state)
+            if self.verbose:
+                print(f"The binary overfilled its RL due to the companion's inflated radius \n"
+                      f"The Roche lobe of the companion is: {RL2}, The radius is: {rc2_i}")
+                print("Adjusting the companion's radius by matching to single star.")
+            t_i = binary.time
+            #m0, t0 = self.track_matcher.get_star_match_data(binary, comp_star)
```
maxbriel:
```suggestion

```

astroJeff (approved):

PR #719, opened 2025-10-06
[fix] Add additional checks for `posydon-popsyn` `check` and `rescue` functions

### LINKED ISSUE #716: [BUG]: `posydon-popsyn check` on empty folder/population run
**Describe the bug**
When `posydon-popsyn check` is run on an empty folder/population run (failed or not submitted), the code fails. It should fail/return a graceful error.

maxbriel:
Add verification that files and folders are present, and if no batch folder is found, the script exists gracefully. closes #716

### DIFF
--- a/bin/posydon-popsyn
+++ b/bin/posydon-popsyn
@@ -493,59 +493,73 @@ def check_batches(run_folder, metallicity, batch_folder_name):
         # Get the jobID
         jobIDs = glob.glob(os.path.join(run_folder,
                                         f'{str_met}_logs/popsyn_*.out'))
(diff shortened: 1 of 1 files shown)


astroJeff:
I'm not sure what the python standard is, but I find these deeply nested if statements difficult to read and follow. Rather than 

```
if len(jobIDs) == 0:
   print("print a failure message")
else:
   # Now here's the rest of the code
```

I prefer something closer to:
```
if len(jobIDs) == 0:
   print("print a failure message")
   return

# Now here's the rest of the code
```

PR #725, opened 2025-10-07
[feat] Make `step_dco` a child class of `step_detached`

### LINKED ISSUE #80: step DCO should be rewriten to inherit the new step detached class
Given the latest makeover of the detached step, one thing that should be done for the v2 code release would be to remove the step_DCO, a copy of the old detached step, and replace it with a method that inherits from the step_detached.

I am assigning this issue to the people working on the new detached step. Please comment here if you think/want to take over the task.

sgossage:
This makes `step_dco` a child class of `step_detached`. Outside of reworking `step_dco` as a child of `step_detached`, I've kept things as they were with the previous implementation of `step_dco`. There may be areas to improve this step still.

This also makes all natal kick arrays have float values in the binary test suite to avoid errors complaining about not having a float e.g., for the mean anomaly.


sgossage:
While working through this, the doc string for the GW radiation equations said that a reference would be added. Is there a reference that people know of for these equations?

maxbriel:
Normally, Peters 1964 is referenced (https://link.aps.org/doi/10.1103/PhysRev.136.B1224). I assume that's the reference used here for the equations. Though it might also be in Peters+Mathews 1963 (https://ui.adsabs.harvard.edu/abs/1963PhRv..131..435P/abstract)

maxbriel (dismissed):

maxbriel:
Just some merge conflicts to resolve, but otherwise looks good to go to me

maxbriel (dismissed):

maxbriel:
The validation binaries look alright. I propagated the changes of `n_o_steps_interval` to the documentation and tutorials too.

maxbriel (dismissed):

maxbriel:
I'm finding changes in the outcome of the step_DCO. Instead of going to CO_contact, binaries are going to maxtime.
Example from the validation set:
https://www.diffchecker.com/7xTTwmtN/

sgossage:
> I'm finding changes in the outcome of the step_DCO. Instead of going to CO_contact, binaries are going to maxtime. Example from the validation set: https://www.diffchecker.com/7xTTwmtN/

Thanks, for some reason the contact event isn't triggering with this binary although the condition is met numerically -- looking into now.

Update: actually it is triggering, but the event is just not being displayed properly.

posydon/binary_evol/DT/double_CO.py
```diff
@@ -22,242 +23,139 @@
...
-            except:
-                set_binary_to_failed(binary)
-                raise NumericalError("SciPy encountered termination edge case while solving GR equations")
-            t_inspiral += s.t[-1]
-            status = s.status
-            n += 1
-            a = s.y[0][-1]
-            e = s.y[1][-1]
-            solt.append(s.t)
-            sol.append(s)
```
sgossage, line 86: As a note: I changed the solver algorithm here. You can see in this old block of code we would call `solve_ivp` up to 6 times, repeating the call if the solver failed (`sol.status == -1`).

Each retry you restart from the last timestep in the previous integration (`sol.t[-1]`) and with the previous integration's last separation and eccentricity.

Do we want to keep this? It's quite different from how the solver is handled in other (non-dco) cases and I'm not sure that it is needed for DCO cases (although we already seem to need to modify the solver to handle DCO evolution, adjusting the tolerance and scale of variables -- those modifications have been retained). In the testing that I have done, I have not seen a difference with/without it.

sgossage:
@maxbriel , I updated this step to display contact events like it should. Can you please confirm that the issue is fixed on your end?

maxbriel:
I'm not seeing the issue anymore! The `CO_contact` is now there.

I'm not too worried with the multi-try being removed. This might be worth asking in the `developers` channel if other might know, why this was there.

maxbriel (dismissed):

posydon/binary_evol/DT/double_CO.py
sgossage: From @ZepeiX on Slack:

> I saw the solver failed sometimes because t_inspiral dropped to very small values, spanning several orders of magnitude. Normally, the loop ensures it done within 2–3 iterations.

sgossage: I've reimplemented this scheme as it appears necessary in some cases upon further testing.

sgossage:
If I can get a second check on the binary evolution test suite from someone, that would be good. I only see a difference with the binary

```
    star1 = SingleStar(**{'mass': 11.862930493162692,
                          'natal_kick_array': [0.0, 0.0, 0.0, 0.0]})
    star2 = SingleStar(**{'mass': 1.4739109294156703,
                        'natal_kick_array': [0.0, 0.0, 0.0, 0.0]})

    binary = BinaryStar(star1, star2,
                        **{'time': 0.0, 'state': 'detached', 'event': 'ZAMS', 'orbital_period': 4111.083887312003, 'eccentricity':0.0},
                        properties = sim_prop)
```

where I'm seeing a timeout when testing (120 seconds limit) on my laptop. Running outside of the test suite script, this runs fine though.

Besides this, from my side everything seems in order.

maxbriel:
@sgossage I ran the binaries and don't see a time-out.
Instead, I get a grid match failure for the exact same binary.


```
🚨 Binary Evolution Failed!
Exception: MatchingError
Message: The mass 0.3687689143581279 is out of the single star grid range and cannot be matched to a track.

6  step_CO_HMS_RLO  detached     CC2       NS    1.878  H-rich_Shell_H_burning    0.370         273.151
7          step_SN  detached    <NA>       NS    1.878                      WD    0.369         273.499
8              nan       ERR  FAILED       NS    1.878                      WD    0.369         273.499

```

It seems to happen after the step_SN. Can it be that it thinks the WD is a star?

sgossage:
> @sgossage I ran the binaries and don't see a time-out. Instead, I get a grid match failure for the exact same binary.
> 
> ```
> 🚨 Binary Evolution Failed!
> Exception: MatchingError
> Message: The mass 0.3687689143581279 is out of the single star grid range and cannot be matched to a track.
> 
> 6  step_CO_HMS_RLO  detached     CC2       NS    1.878  H-rich_Shell_H_burning    0.370         273.151
> 7          step_SN  detached    <NA>       NS    1.878                      WD    0.369         273.499
> 8              nan       ERR  FAILED       NS    1.878                      WD    0.369         273.499
> ```
> 
> It seems to happen after the step_SN. Can it be that it thinks the WD is a star?

Thanks @maxbriel, that was the case. With `step_dco` rewritten, we can have the case where a DCO system has to be handled by `track_match.py` now, where previously it expected at least the secondary star was not a CO. I pushed a commit that improves the handling of DCO systems in `track_match.py`.

This particular failure happened because the `TrackMatcher` object checks that stars satisfy the grid mass boundaries, which it does not need to do for CO stars, so I added a condition to skip this check for CO stars.

I also added a bit of code for handling DCO systems with the `PchipInterpolator` objects. In the old code when we have e.g., a CO primary and normal star for the secondary, we would copy the mass/age of the secondary and use that to construct these objects. We can not do that necessarily when we have two CO objects and ensure we get a valid "dummy" track to use for the CO star.

I added something that just gives a dummy age array that goes from 0 to 14e9 years and sets values of 0 for properties like the moment of inertia (the "kvalues"). These are passed to the `PchipInterpolator` objects for the CO stars. These values are evntually stored in the `interp1d` objects, and eventually they get overwritten anyways for CO stars.

astroJeff (approved):

PR #730, opened 2025-10-16
[feat] Rewrite `posydon-popsyn` CLI interface

### LINKED ISSUE #716: [BUG]: `posydon-popsyn check` on empty folder/population run

maxbriel:
This makes the `posydon-popsyn` command line interface more manageable and modular.

1. `posydon-popsyn check` and  `posydon-popsyn rescue` now do the same. They were already sharing a lot of logic and I've extended the `check` to do the `rescue` stuff, as well. Both are still present for backwards compatibility. 
1. I've moved the majority of the code from `bin/posydon-popsyn` into its own sub-module under `utils`.


posydon/CLI/io.py
```diff
@@ -0,0 +1,575 @@
...
+
+        if __name__ == "__main__":
+            parser = argparse.ArgumentParser()
+            parser.add_argument("metallicity", type=float)
+            args = parser.parse_args()
+            ini_kw = binarypop_kwargs_from_ini("{ini_file}")
+            ini_kw["metallicity"] = args.metallicity
+            str_met = convert_metallicity_to_string(args.metallicity)
+            ini_kw["temp_directory"] = str_met + "_Zsun_" + ini_kw["temp_directory"]
+            synpop = BinaryPopulation(**ini_kw)
```
sgossage, line 89: just checking -- we don't need to provide `SimulationProperties` here (like with `create_run_script`) because we aren't evolving the population, is that right?

posydon/CLI/io.py
```diff
@@ -0,0 +1,575 @@
...
+            str_met = convert_metallicity_to_string(args.metallicity)
+            ini_kw['temp_directory'] = str_met+'_Zsun_' + ini_kw['temp_directory']
+            sim_props = SimulationProperties.from_ini('{ini_file}')
+            synpop = BinaryPopulation(population_properties=sim_props, **ini_kw)
+            synpop.evolve()
+    ''')
+    return text
+
+def create_run_script(ini_file):
+    '''Creates a run script for the population synthesis run.
```
sgossage, line 124: Possible to add brief description of what this script (and the merge script below) does?

posydon/CLI/io.py
maxbriel: That's right! We just need to merge them, which doesn't require the `SimulationProperties` :)

posydon/CLI/io.py
maxbriel: Added.

posydon/CLI/popsyn/check.py
```diff
@@ -0,0 +1,829 @@
...
+    if not all_match:
+        print_error("BINARY COUNT CHECK     ....................ERROR")
+    else:
+        clear_previous_lines(len(metallicities) + 3)
+        print_success("BINARY COUNT CHECK     ....................OK")
+
+    return all_match, counts_dict
+
+def check_run_status(run_folder, metallicities, number_of_binaries):
+    """Check the status of population files and binary counts.
```
sgossage:
```suggestion
    """Check the whether population files exist and if they have 
    the expected number of binary counts.
```

sgossage:
This looks good to me!

sgossage (dismissed):

sgossage (dismissed):

sgossage (approved):

PR #731, opened 2025-10-16
[feat] natal_kick_array renaming

### LINKED ISSUE #155: Consistency in where to store natal kick properties

maxbriel:
Move from a `natal_kick_arary` to the individual components as part of the `SingleStar` object.

This is still backwards compatible, ie. if `natal_kick_array` is given in the configuration file, it will still be outputted in the original format, but a deprecation warning will be raised.
It will also output the new format at the same time. 

I've named the new components:
- natal_kick_velocity
- natal_kick_azimuthal_angle
- natal_kick_polar_angle
- natal_kick_mean_anomaly

### DIFF
--- a/posydon/binary_evol/SN/step_SN.py
+++ b/posydon/binary_evol/SN/step_SN.py
@@ -1497,8 +1497,8 @@ class StepSN(object):
             M_companion = binary.star_2.mass
 
             # check if a kick is passed, otherwise generate it
-            if not binary.star_1.natal_kick_array[0] is None:
-                Vkick = binary.star_1.natal_kick_array[0]
+            if not binary.star_1.natal_kick_velocity is None:
+                Vkick = binary.star_1.natal_kick_velocity
             else:
                 # Draw a random orbital kick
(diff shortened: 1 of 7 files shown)


sgossage:
@maxbriel , confirming that the `metallicity` assignment via `kwargs` is appropriate.

sgossage (approved):

PR #740, opened 2025-10-24
[fix] give 2nd natal kick in validation suite

maxbriel:
This sets the second natal kick to 0 to produce a more consistent output for the validation binaries.


sgossage (dismissed):

maxbriel:
I ended up missing a few where there wasn't a natal_kick present.

astroJeff (approved):

PR #749, opened 2025-10-28
[fix] Adding a case for post HeMS matching and fix `step_CE` surface He assignment

sgossage:
This adds a new case for post HeMS stars that ignores the central he4 abundance in matching. This parameter appears to cause issues in matching when it is extremely small in these core He depleted stars.

This also comments out some of the verbose statements in the ODEs of detached evolution. They print out every timestep which may be a bit much, but we can revert back if that is preferred. It may be more suitable to enable these outputs with e.g., a debugging mode instead of simple verbose output.

This also includes a small fix to `step_CE` in the assignment of surface helium. Previous code used `star.metallicity`, which for solar is `1`, for example. The fix multiplies by our current metallicity scaling, 0.0142 (Asplund et al. 2009) to calculate the abundance in the intended way. Previously this led to negative surface abundance being assigned to the star.

Tested w/ our test suite, fixing some of the previous grid match failures.

### DIFF
--- a/posydon/binary_evol/CE/step_CEE.py
+++ b/posydon/binary_evol/CE/step_CEE.py
@@ -56,6 +56,7 @@ from posydon.utils.common_functions import (
     check_state_of_star,
 )
 from posydon.utils.posydonwarning import Pwarn
+from posydon.utils.constants import Zsun
 
 MODEL = {"prescription": 'alpha-lambda',
          "common_envelope_efficiency": 1.0,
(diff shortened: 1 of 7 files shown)


posydon/binary_evol/CE/step_CEE.py
```diff
@@ -1156,7 +1156,7 @@ def CEE_adjust_binary_upon_ejection(self, binary, donor, mc1_f, rc1_f,
                 if star.metallicity is None:
                     star.surface_he4 = 1 - star.surface_h1 - 0.0142
                 else:
-                    star.surface_he4 = 1.0-star.surface_h1-star.metallicity
+                    star.surface_he4 = 1.0-star.surface_h1-star.metallicity * 0.0142
```
maxbriel:
```
from posydon.utils.constants import Zsun
```

```
star.surface_he4 = 1.0-star.surface_h1-star.metallicity * Zsun
```

sgossage: Thanks, this is now implemented.

astroJeff (approved):
LGTM

PR #750, opened 2025-10-28
[fix] replace `np.array_equal` with `np.allclose` in `unit_tests`

maxbriel:
Replace to avoid errors in float comparisons


maxbriel:
Cannot just replace all instances of `array_equal` with `allclose`.  closing PR

PR #755, opened 2025-10-31
[fix] natal kick input for SingleStar

maxbriel:
I noticed that giving the `natal_kick_array` as an input to a `SingleStar` object was not properly being initiated by the class.
As such, the natal kick properties were not properly set and a random one was given to the star.

This is explicitly for backwards compatibility.


astroJeff (approved):
LGTM

PR #756, opened 2025-10-31
[feat] Rework of interpolators in `step_detached`

maxbriel:
TLDR: This PR speeds up the ODE solve by speeding up interpolation of the SingleStar tracks.

`step_detached` matches the individual stars and then solves the ODE equation using `solve_ivp` from `scipy`.
 Every ODE step, we request the new stellar parameters from the `interp1d` of the star, i.e. `secondary.interp1d['mass'](t)`, where `interp1d` is a dictionary of `PchipInterpolator2` objects. These give back the interpolated value at `t` and is based on the `PchipInterpolator` from `scipy`.
These functions are called many times during the solve because we need to update the stars.

In the original version, we had an individual `PchipInterpolator2` object per property we want to interpolate. However, this is very inefficient, since the `PchipInterpolator` from `scipy` allows for interpolation of multiple output dimensions simultaneously, i.e., ['mass', 'log_R'] etc. can all be requested at the same time, when set up correctly. This significantly speeds up the calculation with less time spent in Python code.

This PR does exactly that. It reworks the interpolator to return all parameters ['mass', 'log_R', 'R', etc] in one go, including forcing positive values and derivatives. It does this by separating the parameter into 4 different `PchipInterpolator` objects based on forcing positive values or requested derivatives. Per these objects, a single `PchipInterpolator` call is done for multiple parameters. This drastically reduces the overhead of this function call.

Below I have included my initial profiling and the result for a few binaries. 
_Note: I run each binary twice before profiling, because this will put the single star track into memory, reducing the IO effect_

> [!IMPORTANT]  
> **In general, this PR cuts down the ODE solve time by 2/3rds and sometimes more than an order of magnitude.**

> [!WARNING]
> There shouldn't be a difference between the outcomes below. There are some small differences of which I am not too worried, see for example the late stage evolution of _Example Binary 1_.

# Example Binary 1

<details>
<summary> binary setup </summary>

```
STAR1 = SingleStar(**{'mass': 30.01386549347856 ,
                      'natal_kick_velocity': 0.0,})
STAR2 = SingleStar(**{'mass': 25.106381375203597 ,
                      'natal_kick_velocity': 0.0})

                    **{'time': 0.0, 'state': 'detached', 'event': 'ZAMS', 'orbital_period':344.5337684440942  , 'eccentricity': 0.0},
                    properties = sim_pop)
```

</details>

## POSYDON v2.2: [1.97s]

image

<details>
<summary> Profiler </summary>

image

</details>

 ## This PR [0.6s]
image

<details>
<summary> Profiler </summary>
image

</details>

# Example binary 2

> [!NOTE]
> This is a difficult binary to solve, but shouldn't happen "too" often. We can maybe alter the solver tolerances to make it finish faster.

<details>
<summary> binary setup </summary>

```
star_1 = SingleStar(**{'mass':  7.082482087008959 , 'state': 'H-rich_Core_H_burning','metallicity':0.1,
                'natal_kick_velocity': 0.0,})
star_2 = SingleStar(**{'mass': 2.6379543729221413 , 'state': 'H-rich_Core_H_burning','metallicity':0.1,
                'natal_kick_velocity': 0.0})
                        'orbital_period': 115.08514438464961 ,'eccentricity': 0.0}, properties = sim_pop)
```
</details>

## POSYDON v2.2 [123s+failure]

**Hits 120 seconds, maximum evolve time per binary**

image

<details>
<summary> Profiler </summary>
image

</details>

## This PR [54s]

image

<details>
<summary> Profiler </summary>
image

</details>

# Example binary 3

<details>
<summary> binary setup </summary>

```
STAR1 = SingleStar(**{'mass': 161.7903633300293 ,
                      'natal_kick_velocity': 0.0,})
STAR2 = SingleStar(**{'mass': 83.70740542817626 ,
                      'natal_kick_velocity': 0.0})

                    **{'time': 0.0, 'state': 'detached', 'event': 'ZAMS', 'orbital_period':12.399212118998499 , 'eccentricity': 0.0},
                    properties = sim_pop)
```
</details>

## POSYDON v2.2 [1.58s]

image

<details>
<summary> Profiler </summary>

image

</details>

## This PR [0.42s]
image

<details>
<summary> Profiler </summary>
image

</details>


astroJeff:
What is the output when applied to the validation binaries?

posydon/binary_evol/CE/step_CEE.py
```diff
@@ -1501,8 +1501,10 @@ def adjust_secondary_radius(self,comp_star,binary,mc1_i,mc2_i,rc1_i,rc2_i,separa
             _, _, only_CO = self.track_matcher.do_matching(binary, step_name="step_CE",
                                                            match_primary=match_primary,
                                                            match_secondary=match_secondary)
-
-            rc2_i = 10**comp_star.interp1d['log_R'](t_i)
+            if comp_star.interp1d:
+                rc2_i = comp_star.interp1d(t_i)["R"]
+            else:
+                rc2_i = rc2_i  # no change
```
astroJeff, line 1507: This looks like a problem to me `do_matching` should raise an error if the matching step fails. At this point in the function `comp_star.interp1d` should always exist

posydon/binary_evol/DT/step_detached.py
```diff
@@ -1115,10 +1125,10 @@ def set_stars(self, primary, secondary, t0=0.0):
         self.primary = primary
         self.secondary = secondary
         # update list of tracked physical properties
-        self.phys_keys = list(secondary.interp1d.keys())
+        self.phys_keys = list(secondary.interp1d.keys)
         # except we don't evolve these:
-        for prop in ["t0", "m0", "t_max", "max_time"]:
-            self.phys_keys.remove(prop)
+        #for prop in ["t0", "m0", "t_max", "max_time"]:
+        #    self.phys_keys.remove(prop)
```
astroJeff: Can remove these lines

posydon/utils/interpolators.py
```diff
@@ -121,59 +121,115 @@ def __call__(self, x_new):
             raise NotImplementedError(f"kind = {self.kind} is not supported")
 
 
-class PchipInterpolator2:
-    """Interpolation class."""
-    def __init__(self, *args, positive=False, derivative=False, **kwargs):
+class StellarInterpolator:
```
astroJeff: Can we call this SingleStarInterpolator?

maxbriel:
@astroJeff I ran the validation binaries, and the outcome is the exact same for non-kicked binaries. The other ones look alright.

 I am still having some issues with the kicked systems, which keep getting a randomly drawn kick for some supernova steps. I'm not sure why it's still being drawn at the moment. @sgossage, you seemed to have solved this in your branch?

elizabethteng:
I have triumphed over interpolators.py and its unit test

posydon/binary_evol/CE/step_CEE.py
sgossage: we have 

```
        if (secondary.interp1d == None and self.match_secondary) or \
           (primary.interp1d == None and self.match_primary):
            failed_state = binary.state
            set_binary_to_failed(binary)
            raise MatchingError("Grid matching failed for "
                                f"{failed_state} binary. "
                                f"\nsecondary.interp1d = {secondary.interp1d}"
                                f"\nprimary.interp1d = {primary.interp1d}")
```

in `track_match.py`. In `step_CE`, we only match the secondary (companion star), so if the `interp1d` does not exist for the secondary, this error should trigger before we get to the above chunk of code in `step_CE`. Do we need the `else` statement?

astroJeff (dismissed):
LGTM

sgossage:
> @astroJeff I ran the validation binaries, and the outcome is the exact same for non-kicked binaries. The other ones look alright.
> 
> I am still having some issues with the kicked systems, which keep getting a randomly drawn kick for some supernova steps. I'm not sure why it's still being drawn at the moment. @sgossage, you seemed to have solved this in your branch?

@maxbriel I still see this issue. We opened up a bug report [Issue #764](https://github.com/POSYDON-code/POSYDON/issues/764) to address this in the future.

sgossage:
I pushed an update to fix an issue that prevented binary evolution from proceeding due to an indexing error. 

```
--> 831     current = interp1d[self.translate[key]][-1].item()
    832     history = interp1d[self.translate[key]][:-1]

IndexError: invalid index to scalar variable.
```

The problem was that `update_after_evo`  (in `step_detached.py`) was not built to interact with the new functionality of the new `interp1d`  objects. Since the new objects use the time array to create interpolated values, we do not need to index the interpolated values again in this function (it's done elsewhere when calling the interpolator object).

I updated that function to utilize the time arrays according to this new functionality. I also simplified the code in this function a bit. Tested with default settings and `dt` /`n_o_steps_history`  set and binary evolution is working on my end.
I also made it so that the function only appends to the binary/star history if the `dt`  or `n_o_steps_history`  variables are set (w/o this, the history can get excess entries).

sgossage (dismissed):

sgossage (approved):

PR #758, opened 2025-11-01
[fix] Fixes for indentation for created scripts for popsyn

maxbriel:
This fixes a bug introduced during the rewrite of the `posyon-popsyn` commands, where the indentation of the created scripts during `setup` is wrong.


astroJeff (approved):
LGTM

PR #760, opened 2025-11-03
[fix] minimum redshift bin Chruslinska+21 data

maxbriel:
Adds a fix for when the minimum requested redshift bin is lower than the Chruslinska+21 data has available.
Uses the lowest redshift bin as the appropriate proxy.

### DIFF
--- a/posydon/popsyn/star_formation_history.py
+++ b/posydon/popsyn/star_formation_history.py
@@ -652,7 +652,8 @@ class IllustrisTNG(SFHBase):
 
         Note:
         We are using the closest redshift bin in the IllustrisTNG simulation
-        to extract the metallicity distribution for the requested redshift.
+        that is smaller than the requested redshift to extract the metallicity
+        distribution for the requested redshift.
 
         Parameters
         ----------
--- a/posydon/popsyn/star_formation_history.py
+++ b/posydon/popsyn/star_formation_history.py
@@ -826,7 +826,17 @@ class Chruslinska21(SFHBase):
             Fraction of the SFR in the given metallicity bin at the given redshift.
         """
         # only use data within the metallicity bounds (no lower bound)
-        redshift_indices = np.array([np.where(self.redshifts <= i)[0][0] for i in z])
+        if np.min(z) < np.min(self.redshifts):
+            Pwarn('Some redshift values are lower than the minimum '
+                    'redshift in the Chruslinska+21 data. '
+                    'Using the minimum redshift available.',
+                    'SFHModelWarning')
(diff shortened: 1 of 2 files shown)


posydon/popsyn/star_formation_history.py
```diff
@@ -826,7 +826,17 @@ def fSFR(self, z, metallicity_bins):
             Fraction of the SFR in the given metallicity bin at the given redshift.
         """
         # only use data within the metallicity bounds (no lower bound)
-        redshift_indices = np.array([np.where(self.redshifts <= i)[0][0] for i in z])
+        if np.min(z) < np.min(self.redshifts):
+            Pwarn('Some redshift values are lower than the minimum '
+                    'redshift in the Chruslinska+21 data. '
+                    'Using the minimum redshift available.',
+                    'SFHModelWarning')
+            redshift_indices = np.array([-1 if i < np.min(self.redshifts)
```
astroJeff, line 844: `-1` because the Chruslinska data is ordered such that the `-1` value is the lowest redshift value in the set?

posydon/popsyn/star_formation_history.py
```diff
@@ -826,7 +826,17 @@ def fSFR(self, z, metallicity_bins):
...
+                                    else np.where(self.redshifts <= i)[0][0]
+                                    for i in z])
+        else:
+            redshift_indices = np.array([np.where(self.redshifts <= i)[0][0] for i in z])
```
astroJeff, line 848: I'm also realizing that the way we map onto the Chruslinska data is we are choosing the first bin where `z >= self.redshifts`. This is essentially a ceiling function, no?

posydon/popsyn/star_formation_history.py
maxbriel: That's correct.

posydon/popsyn/star_formation_history.py
maxbriel: Yes. This was how we were doing the Illustris data, so it's the same implementation. We could move to something more advanced?

astroJeff: No, I think it's fine. But can you add a comment somewhere higher up in the code, so we have this stated clearly for the user. Ideally, in either the comments for the method using the Chruslinska data or in the ini file for running the population?

maxbriel: @astroJeff added this as part of the docstring of this function.

elizabethteng (approved):

astroJeff (approved):

elizabethteng (approved):

PR #769, opened 2025-12-04
[fix] History array handling in step_detached.py

sgossage:
Fixes array access handling when accessing star properties after detached evolution. This also bumps the version number to 2.2.1


astroJeff (approved):
LGTM

PR #771, opened 2025-12-05
Jeff m d s log p fix

astroJeff:
Fixing a bug in the Moe & Di Stefano sampler, so the binary periods are returned, not the log binary periods


sgossage (approved):

kasdaglie:
I have tested this PR with a small population, everything seems to be working @astroJeff

PR #776, opened 2025-12-12
[fix] Making it so `SimulationProperties` accepts arbitrary `kwargs`

### LINKED ISSUE #763: [bug] Custom steps cannot be loaded
Due to the definitions of each step by name as an input, it's no longer possible to load in custom steps defined inside a configuration file.
Can we allow for arbitrary names to be loaded in?

https://github.com/POSYDON-code/POSYDON/blob/03690af948425748bd09137536a3dab7a54ad30f/posydon/binary_evol/simulationproperties.py#L30-L43

sgossage:
Adding a `kwargs` argument to the `SimulationProperties` class. This allows the class to accept custom step names for example that are not a part of the standard set.

### DIFF
--- a/conda/meta.yaml
+++ b/conda/meta.yaml
@@ -1,5 +1,5 @@
 {% set name = "posydon" %}
-{% set version = "2.2.2" %}
+{% set version = "2.2.3" %}
 
 package:
   name: "{{ name|lower }}"
--- a/posydon/binary_evol/simulationproperties.py
+++ b/posydon/binary_evol/simulationproperties.py
@@ -43,7 +43,7 @@ class SimulationProperties:
                        step_CE = (NullStep(), {}),
                        step_end = (NullStep(), {}),
                        extra_hooks = [],
-                       verbose=False):
+                       **kwargs):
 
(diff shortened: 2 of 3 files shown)


maxbriel:
This fixes the issue of loading in custom steps.

Is it needed to keep the default steps?
It will raise a warning when replacing a default step name with a default one.

sgossage:
We discussed today at the dev meeting a bit and just logging the discussion here.

An advantage of keeping the default steps is to maintain an explicit listing of the default flow. Throwing a warning could be nice to just alert that a default has changed (which may be intentional).

We may want to instead store the default step list in our flow chart code and import it here (or where needed).

I think we can remove the defaults but should consider whether it is better to keep them for code verbosity. Ideally we don't sacrifice flexibility though, or make things verbose to the point of obscuration.

maxbriel:
Since the warnings are just outputted in the verbose output, I'm okay to accept this PR.

maxbriel (approved):

sgossage:
Bypassing rules to squash and merge,

PR #780, opened 2025-12-18
Validate test binaries upon code changes

elizabethteng:
- [x] enhance 1Z_sun_binaries_suite.py to save evolution data to h5 file in ./outputs
- [x] update evolve_binaries.sh to reflect output saving
- [x] write validate_binaries.sh (script for single-line command for comparing branches)
    - [x] call evolve_binaries.sh for new branch
    - [x] call compare_runs.py and save output to file in ./outputs for difference report 
- [x] extract data from runs and save to file
- [x] write compare_runs.py (compares two h5 files and reports differences)
    - [x] outline
    - [x] quantitative differences beyond thresholds
    - [x] qualitative differences to evolution
    - [x] differences to warnings and errors
- [x] generate baseline.h5
- [x] extend to other metallicities
- [x] tweaks suggested during Mar 12 meeting (They are in your physical notebook!!)


dev-tools/script_data/1Zsun_binaries_suite.py
```diff
@@ -153,18 +153,13 @@ def warning_handler(message, category, filename, lineno, file=None, line=None):
             grp.create_dataset(col, data=df[col].values)
 
     except Exception as e:
-        # turn off binary alarm in case of exception
-        signal.alarm(0)
```
maxbriel, line 158: @elizabethteng I think these are now alright to remove, because we've added the following to `binarystar.evolve()`:

```
        finally:
            signal.alarm(0)     # turning off alarm
```

dev-tools/outputs/baseline.h5
maxbriel, line 1: Do you need this file still?

dev-tools/compare_runs.py
```diff
@@ -0,0 +1,412 @@
...
+        if cand_only_cols:
+            struct_diffs.append(f"Binary {bid}: columns only in candidate: {sorted(cand_only_cols)}")
+
+        # ── Per-column comparison ─────────────────────────────────────
+        common_cols = sorted(set(b.columns) & set(c.columns) - {'binary_id'})
+        min_rows = min(len(b), len(c))
+
+        for col in common_cols:
+            b_col = b[col].iloc[:min_rows]
+            c_col = c[col].iloc[:min_rows]
```
maxbriel, line 119: Is there a specific reason to only loop over the minimum number of columns?
What should happen if the other is much longer?

dev-tools/compare_runs.py
```diff
@@ -0,0 +1,412 @@
...
+
+    return diffs
+
+
+def read_table_safe(store, key):
+    """Read a table from HDFStore, returning None if it doesn't exist."""
+    try:
+        if key in store:
+            return store[key]
+    except Exception:
```
maxbriel, line 305: Is there a specific reason to return `None` here?

maxbriel (commented):
This generally looks really good! 

I do need a bit more explanation still on how to run it exactly and why the transition to binaries_suite for multiple metallicities? 
Although the test binaries originally came from Zsun populations, I guess we can run them on all metallicities.

Would you be able to switch to merging into `v2.3` branch? Or at least update it to the latest of `main`.

dev-tools/compare_runs.py
elizabethteng: this is about rows (evolution steps) - the step count difference is flagged as a difference above as "evolution step count differs". Per-column comparison only covers overlapping steps, extra steps are silently ignored other than the note about difference in number of steps.

dev-tools/compare_runs.py
elizabethteng: If something goes wrong reading the table from h5 file, it returns None instead of crashing. So "None" could mean either that the table wasn't there or it was but something went wrong reading it. This makes it so that the comparison script always runs to completion, which may help if for example one metallicity has problems but the others don't.

elizabethteng:
I added a README file to give more guidance about running. Each file has a header with more details. I wanted to make it possible to do multiple metallicities so that you wouldn't have to run multiple scripts, but it is possible to do evolution on a subset of metallicities or a single metallicity using command line arguments.

sgossage:
@elizabethteng , during a dev meeting we saw that there were some differences between values when comparing two branches according to these tests. I'm planning to help by looking into the source of those differences. Can you please post here which two branches produced those differences if you have them on hand?

sgossage:
At least in my testing, I noticed differences between branches that would occur only for `S1_lg_mdot` during `step_detached`.

This difference indeed seems to be caused by the randomization used in `bondi_hoyle` during this step. I was able to resolve the differences using [PR#826](https://github.com/POSYDON-code/POSYDON/pull/826). Creating a base line with the branch `sg_v2.3_rng_for_steps` and fixing the RNG seed, then comparing to the same branch with the same and different seeds produced: 0 differences when using the same seed and similar differences that I saw when comparing e.g. `main` and `v2.3` when using seeds different from the base line.

dev-tools/README.md
```diff
@@ -0,0 +1,130 @@
...
+python compare_runs.py baseline.h5 candidate.h5 [--loose] [--rtol VALUE] [--atol VALUE] [--verbose]
+```
+
+### `binaries_params.ini`
+
+Configuration file for `SimulationProperties`. Defines the POSYDON evolution steps, supernova prescriptions, common envelope parameters, and output column selections. Metallicity is overridden at runtime by `binaries_suite.py`.
+
+## Running Scripts Manually
+
+The shell scripts handle cloning, environment setup, and orchestration. If you already have POSYDON installed in your current environment, you can run the Python scripts directly.
```
sgossage:
```suggestion
The shell scripts handle cloning, environment setup, orchestration, and execution. If you already have POSYDON installed in your current environment, you can execute the Python scripts directly.
```

dev-tools/README.md
```diff
@@ -0,0 +1,130 @@
+Validation suite for POSYDON binary evolution. Evolves a fixed set of test binaries on a candidate branch and compares results against a stored baseline to catch regressions.
```
sgossage:
```suggestion
Validation suite for POSYDON binary evolution. Evolves a fixed set of test binaries on a candidate branch and compares results against a stored baseline to catch regressions. A baseline can be formed from any branch (`main` by default) and is represented by a set of results from `binary_suite.py`, saved HDF5 files, all stored in `dev-tools/baselines/<branch-name>`.
```

dev-tools/README.md
```diff
@@ -0,0 +1,130 @@
...
+```bash
+./generate_baseline.sh main "" "1 0.45"
+./validate_binaries.sh feature/my-branch main "1 0.45"
+```
+
+## Scripts
+
+### `validate_binaries.sh`
+
+Top-level entry point. Evolves test binaries on a candidate branch, then compares results against an existing baseline.
```
sgossage:
```suggestion
Top-level entry point. Evolves test binaries on a candidate branch, then compares results against an existing baseline. This script will look for baseline HDF5 files stored in `dev-tools/baseline/<branch-name>`, where `main` is the default `<branch-name>`.
```

dev-tools/README.md
```diff
@@ -0,0 +1,130 @@
...
+
+```bash
+./validate_binaries.sh <candidate_branch> [baseline_branch] [metallicities] [--loose] [--rtol VALUE] [--atol VALUE]
+```
+
+By default, comparison is exact (rtol=0, atol=0). Use `--loose` for relaxed floating-point tolerances (rtol=1e-12, atol=1e-15), or set `--rtol`/`--atol` explicitly as per [np.allclose](https://numpy.org/devdocs/reference/generated/numpy.allclose.html).
+
+### `generate_baseline.sh`
+
+Generates baseline HDF5 files from a designated branch or tag. Can also promote existing outputs to baseline with `--promote`.
```
sgossage:
```suggestion
Generates baseline HDF5 files from a designated branch name and optionally a SHA to specify a commit. Can also promote existing outputs to baseline with `--promote`.
```

dev-tools/README.md
```diff
@@ -0,0 +1,130 @@
...
+
+Top-level entry point. Evolves test binaries on a candidate branch, then compares results against an existing baseline.
+
+```bash
+./validate_binaries.sh <candidate_branch> [baseline_branch] [metallicities] [--loose] [--rtol VALUE] [--atol VALUE]
+```
+
+By default, comparison is exact (rtol=0, atol=0). Use `--loose` for relaxed floating-point tolerances (rtol=1e-12, atol=1e-15), or set `--rtol`/`--atol` explicitly as per [np.allclose](https://numpy.org/devdocs/reference/generated/numpy.allclose.html).
+
+### `generate_baseline.sh`
```
sgossage, line 44: In this section (`generate_baseline.sh`) I think the `--promote` option should be briefly explained just a bit more.

It was not immediately clear to me when I read through the README what this does. I suggest explaining the workflow required to use the `promote` option.

```
If you already have results from prior runs of `evolve_binaries.sh` saved as HDF5 files in `dev-tools/outputs/<branch-name>/`, you can use these as a baseline with the `--promote` option:

./generate_baseline.sh --promote <branch> [metallicities]
```

elizabethteng: done! thx for suggestion :^)

sgossage (approved):

astroJeff (approved):

PR #782, opened 2026-01-08
Automate conda version by extracting it from setup.py

maxbriel:
Extracts the versioneer version using conda's build in function to add it inside `meta.yaml`. 
I wasn't able to test properly, because my local `conda` build fails (also without this change). 

See here for an example: https://docs.conda.io/projects/conda-build/en/stable/resources/define-metadata.html#loading-data-from-other-files

### DIFF
--- a/conda/meta.yaml
+++ b/conda/meta.yaml
@@ -1,9 +1,9 @@
 {% set name = "posydon" %}
-{% set version = "2.2.3" %}
+{% set data = load_setup_py_data(setup_file='../setup.py', from_recipe_dir=True) %}
 
 package:
   name: "{{ name|lower }}"
-  version: "{{ version }}"
+  version: "{{ data.version }}"
 
 source:
   path: ..
--- a/conda/meta.yaml
+++ b/conda/meta.yaml
@@ -3,7 +3,7 @@
 
 package:
   name: "{{ name|lower }}"
-  version: "{{ data.version }}"
+  version: "{{ data.get('version') }}"
 
 source:
   path: ..


sgossage:
@maxbriel, how do we update the project version with this approach? Is the idea that we'd edit

```
setup(
    name=DISTNAME,
    provides=[PACKAGENAME],
    version=__version__,
    ...
```

in `setup.py` so that `version = NEW_VERSION`?

maxbriel:
This branch still has `versioneer.py` installed, so it will use the github tag of the release to set the version in `setup.py`. Then the `load_setup_py_data` from `conda` loads in the `setup.py` with the associated variables. This contains the version set by `versioneer.py`. It then uses that tag to set the `conda` tag. #784 would replace this with a different mechanism to achieve the same using `setuptools-scm`.

It should be possible to test this on Linux with `conda build`.
There it should be possible to verify the tag (I will try to get my local github actions working again).

This basically removes the manual step of updating the `conda` version number for each release.

sgossage:
thanks @maxbriel  I think this looks good. I tested this locally and it appears to pick up the correct version.

I believe we could still have e.g., custom versions for development branches if we wanted, using e.g.,

```
git tag <new-tag>
```

to set a custom tag for the branch locally. Otherwise, the local branch should always report the nearest version tag associated with that branch's commit.

sgossage (approved):

astroJeff (approved):

elizabethteng (approved):

PR #784, opened 2026-01-08
[enhancement] Clean POSYDON repository

maxbriel:
This pull request:
- Updates the setuptools we use.
- Moves from versioneer.py to setuptools-scm for easier intergration.
- Automatically updates the conda.yaml version.
- Moves project definitions and requirements into pyproject.toml instead of defining them inside setup.py.
- Cleans up test CI by adding project parameters to pyproject.toml.


astroJeff:
Needs proper testing. @sgossage can you help out with this?

conda/meta.yaml
```diff
@@ -1,9 +1,8 @@
 {% set name = "posydon" %}
-{% set version = "2.2.4" %}
 
 package:
   name: "{{ name|lower }}"
-  version: "{{ version }}"
+  version: {{ GIT_DESCRIBE_TAG }}
```
sgossage, line 5: Should we revert to the old `version: "{{ version }}"`?

Otherwise I think we can test if this works for us with a release after merging.

sgossage: Things seem to be working well. In my local test of `publish_conda`, the tag `v2.2.4` gets picked up, I think because that is the latest tag this branch can grab.

astroJeff (approved):

conda/meta.yaml
```diff
@@ -1,9 +1,8 @@
 {% set name = "posydon" %}
-{% set version = "2.2.6" %}
 
 package:
   name: "{{ name|lower }}"
-  version: "{{ version }}"
+  version: {{ GIT_DESCRIBE_TAG }}
```
astro-abhishek, line 5: minor stylistic suggestion: leave the quotes on 

> version: "{{ GIT_DESCRIBE_TAG }}"

sgossage:
There have been a few updates to the `ubuntu-latest` image in March it looks like and now the Github actions involving that image are failing. These could be related. To restore things, it looks like we need to update the test coverage.

sgossage (approved):

maxbriel:
@sgossage are you sure these merges went okay? Do you want me to cherry pick this PR into `v2.3` instead of `main`?

sgossage:
> @sgossage are you sure these merges went okay? Do you want me to cherry pick this PR into `v2.3` instead of `main`?

Yes, I merged `v2.3` into this thinking this PR was directed there, but then realized it's pointed at `main`. So I reverted back to [`https://github.com/POSYDON-code/POSYDON/commit/ff972409838de370dcf982e07e726acb8ff3345b`](https://github.com/POSYDON-code/POSYDON/commit/ff972409838de370dcf982e07e726acb8ff3345b) (before the merge) and continued from there.

As long as you can confirm that the files changed look as you remember them, this should be fine and we can squash and merge. Besides that, I added a fix for a unit test that would fail on Ubuntu and fixed a tab/space issue in `compress_mesa_files.py`.

The unit test was failing because 

```python
        if len(MESA_runs)>0:
            for i in [0, -1]:
                MESA_run_dir = os.path.join(MESA_dir, MESA_runs[i])
                os.symlink(MESA_run_dir, os.path.join(MESA_dir, f"link{i}"))
                
                # this line is not guaranteed to get a file, e.g., could be a directory
                MESA_run_file = os.path.join(MESA_run_dir,\
                                             os.listdir(MESA_run_dir)[0])

                os.symlink(real_file, os.path.join(MESA_dir, f"link{i}.file0"))
```

would sometimes fail to get and create a symlink for a file on `ubuntu-latest`, meaning test coverage wouldn't reach 100%.

However, I do think that this might fit better in `v2.3` because this is probably the last patch we push to `main` (v2.2). So, this will probably get (relatively) quickly overshadowed by that new minor version update, which contains many changes. I'm fine which whatever is more convenient here though.

Having said that, [PR#838](https://github.com/POSYDON-code/POSYDON/pull/838) is now affected by the same unit test failure, so may be worth it to push this as a fix to `main` anyways so that PR can be resolved.

PR #785, opened 2026-01-15
[fix] Corrected/enhanced merger process

ezapartas:
Implementing and solving issues already discussed in a previous PR version #311 which is now deprecated.

- No deepcopy during merger (which was working, but led to memory and speed issues)
- However, correct caclulations of the mass weighted averages of abundances BEFORE mass changes.
- calculating center_gamma and avg_c_in_c_core for mergers whenever possible (as the core value if the cores do not merge and np.nap and mass-weighted average respectively, if the cores merge).
- added co_core_mass as a potential option in the matching criteria at track_match (mostly for mergers)

Reminder : The resulting merger product is continuing as a singlestar instance at the place of the engulfing star (i.e. if oMerging1, then star_1 would be the merged product) UNLESS a NS/BH is involved, where the latter is the only one surviving (potential Thorne-Zytkov object).

### DIFF
--- a/posydon/binary_evol/DT/step_merged.py
+++ b/posydon/binary_evol/DT/step_merged.py
@@ -176,7 +176,7 @@ class MergedStep(IsolatedStep):
                 M2 = getattr(star2, "he_core_mass") - getattr(star2, "co_core_mass")
             else:
                 M2 = getattr(star2, mass_weight2)
-            
+
             try:
             	mass_weighted_avg_value=(A1*M1+A2*M2)/(M1+M2)
             except TypeError:
@@ -193,7 +193,7 @@ class MergedStep(IsolatedStep):
         if ( s1 in LIST_ACCEPTABLE_STATES_FOR_HMS
             and s2 in LIST_ACCEPTABLE_STATES_FOR_HMS):
                 #these stellar attributes change value
-                
+
                 #TODO for key in ["center_h1", "center_he4", "center_c12", "center_n14","center_o16"]:
                 merged_star.center_h1 = mass_weighted_avg()
                 merged_star.center_he4 = mass_weighted_avg(abundance_name = "center_he4")
@@ -210,7 +210,7 @@ class MergedStep(IsolatedStep):
                 merged_star.surface_o16 = mass_weighted_avg(abundance_name = "surface_o16")
 
                 # The change of masses occurs after the calculation of weighted averages
-                merged_star.mass = (star_base.mass + comp.mass) * (1.-self.rel_mass_lost_HMS_HMS) 
+                merged_star.mass = (star_base.mass + comp.mass) * (1.-self.rel_mass_lost_HMS_HMS)
 
                 for key in STARPROPERTIES:
                     # these stellar attributes become np.nan
@@ -234,7 +234,7 @@ class MergedStep(IsolatedStep):
                 merged_star.surface_o16 = mass_weighted_avg(abundance_name = "surface_o16", mass_weight1="H-rich_envelope_mass", mass_weight2="mass")
 
                 # The change of masses occurs after the calculation of weighted averages
-                merged_star.mass = star_base.mass + comp.mass 
+                merged_star.mass = star_base.mass + comp.mass
 
                 for key in STARPROPERTIES:
                     # these stellar attributes become np.nan
--- a/posydon/binary_evol/DT/step_merged.py
+++ b/posydon/binary_evol/DT/step_merged.py
@@ -269,8 +269,7 @@ class MergedStep(IsolatedStep):
                     for substring in ["log_", "lg_", "surf_", "conv_", "lambda", "profile"]:
                         if (substring in key) :
                             setattr(merged_star, key, np.nan)
-                        if key in [ "c12_c12", "center_gamma",
-                                   "avg_c_in_c_core", "total_moment_of_inertia", "spin"]:
+                        if key in [ "c12_c12", "total_moment_of_inertia", "spin"]:
                             setattr(merged_star, key, np.nan)
 
                 merged_star.log_LHe = comp.log_LHe
--- a/posydon/binary_evol/DT/step_merged.py
+++ b/posydon/binary_evol/DT/step_merged.py
@@ -208,7 +208,8 @@ class MergedStep(IsolatedStep):
                 merged_star.surface_n14 = mass_weighted_avg(abundance_name = "surface_n14")
                 merged_star.surface_o16 = mass_weighted_avg(abundance_name = "surface_o16")
 
-                merged_star.mass = (star_base.mass + comp.mass) * (1.-self.rel_mass_lost_HMS_HMS)
+                # The change of masses occurs after the calculation of weighted averages
+                merged_star.mass = (star_base.mass + comp.mass) * (1.-self.rel_mass_lost_HMS_HMS) 
 
                 for key in STARPROPERTIES:
                     # these stellar attributes become np.nan
@@ -224,8 +225,6 @@ class MergedStep(IsolatedStep):
         elif (s1 in LIST_ACCEPTABLE_STATES_FOR_POSTMS
             and s2 in LIST_ACCEPTABLE_STATES_FOR_HMS):
 
-                merged_star.mass = star_base.mass + comp.mass #TODO: in step_CEE we need to eject part of the (common) envelope
-
                 # weigheted mixing on the surface abundances of the whole comp with the envelope of star_base
                 merged_star.surface_h1 = mass_weighted_avg(abundance_name = "surface_h1", mass_weight1="H-rich_envelope_mass", mass_weight2="mass")
                 merged_star.surface_he4 = mass_weighted_avg(abundance_name = "surface_he4", mass_weight1="H-rich_envelope_mass", mass_weight2="mass")
@@ -233,13 +232,15 @@ class MergedStep(IsolatedStep):
                 merged_star.surface_n14 = mass_weighted_avg(abundance_name = "surface_n14", mass_weight1="H-rich_envelope_mass", mass_weight2="mass")
                 merged_star.surface_o16 = mass_weighted_avg(abundance_name = "surface_o16", mass_weight1="H-rich_envelope_mass", mass_weight2="mass")
 
+                # The change of masses occurs after the calculation of weighted averages
+                merged_star.mass = star_base.mass + comp.mass 
+
                 for key in STARPROPERTIES:
                     # these stellar attributes become np.nan
                     for substring in ["log_", "lg_", "surf_", "conv_", "lambda", "profile"]:
                         if (substring in key) :
                             setattr(merged_star, key, np.nan)
-                        if key in [ "c12_c12", "center_gamma",
-                                   "avg_c_in_c_core", "total_moment_of_inertia", "spin"]:
+                        if key in [ "c12_c12", "total_moment_of_inertia", "spin"]:
                             setattr(merged_star, key, np.nan)
 
                 merged_star.log_LHe = star_base.log_LHe
--- a/posydon/binary_evol/DT/step_merged.py
+++ b/posydon/binary_evol/DT/step_merged.py
@@ -93,12 +93,13 @@ class MergedStep(IsolatedStep):
         if self.verbose:
             print("Before Merger", binary.star_1.state, binary.star_2.state,
                   binary.state, binary.event)
-            print("M1 , M2, he_core_mass1, he_core_mass2: ",
+            print("M1 , M2, he_core_mass1, he_core_mass2, co_core_mass1, co_core_mass2: ",
                   binary.star_1.mass, binary.star_2.mass,
-                  binary.star_1.he_core_mass, binary.star_2.he_core_mass)
-            print("star_1.center_he4, star_2.center_he4, star_1.surface_he4, "
-                  "star_2.surface_he4: ", binary.star_1.center_he4,
-                  binary.star_2.center_he4, binary.star_1.surface_he4,
+                  binary.star_1.he_core_mass, binary.star_2.he_core_mass, binary.star_1.co_core_mass, binary.star_2.co_core_mass)
+            print("star_1.center_h1, star_2.center_h1, star_1.center_he4, star_1.center_c12, star_2.center_c12, "
+                  "star_2.center_he4, star_1.surface_he4, star_2.surface_he4: ",
+                  binary.star_1.center_h1, binary.star_2.center_h1, binary.star_1.center_he4,
+                  binary.star_2.center_he4, binary.star_1.center_c12, binary.star_2.center_c12, binary.star_1.surface_he4,
                   binary.star_2.surface_he4)
 
         if binary.state == "merged":
@@ -126,10 +127,16 @@ class MergedStep(IsolatedStep):
         if self.verbose:
             print("After Merger", binary.star_1.state, binary.star_2.state,
                   binary.state, binary.event)
-            print("M_merged , he_core_mass merged: ", binary.star_1.mass,
-                  binary.star_1.he_core_mass)
-            print("star_1.center_he4, star_1.surface_he4: ",
-                  binary.star_1.center_he4, binary.star_1.surface_he4)
+            if binary.event == 'oMerging1':
+                print("M_merged , he_core_mass merged, co_core_mass merged: ", binary.star_1.mass,
+                  binary.star_1.he_core_mass, binary.star_1.co_core_mass)
+                print("star_1.center_h1, star_1.center_he4, star_1.center_c12, star_1.surface_he4: ",
+                  binary.star_1.center_h1, binary.star_1.center_he4, binary.star_1.center_c12, binary.star_1.surface_he4)
+            elif binary.event=='oMerging2':
+                print("M_merged , he_core_mass merged, co_core_mass merged: ", binary.star_2.mass,
+                  binary.star_2.he_core_mass, binary.star_2.co_core_mass)
+                print("star_2.center_h1, star_2.center_he4, star_2.center_c12, star_2.surface_he4: ",
+                  binary.star_2.center_h1, binary.star_2.center_he4, binary.star_2.center_c12, binary.star_2.surface_he4)
 
         super().__call__(binary)
 
--- a/posydon/binary_evol/DT/step_merged.py
+++ b/posydon/binary_evol/DT/step_merged.py
@@ -175,13 +175,24 @@ class MergedStep(IsolatedStep):
                 M2 = getattr(star2, "he_core_mass") - getattr(star2, "co_core_mass")
             else:
                 M2 = getattr(star2, mass_weight2)
-            return (A1*M1 + A2*M2 ) / (M1+M2)
+            
+            try:
+            	mass_weighted_avg_value=(A1*M1+A2*M2)/(M1+M2)
+            except TypeError:
+            	mass_weighted_avg_value= np.nan
+
+            if self.verbose:
+                print(abundance_name, mass_weight1, mass_weight2)
+                print("A_base, M_base_abundance, A_comp, M_comp_abundanc", A1, M1, A2, M2)
+                print("mass_weighted_avg= ", mass_weighted_avg_value)
+
+            return mass_weighted_avg_value
 
         # MS + MS
         if ( s1 in LIST_ACCEPTABLE_STATES_FOR_HMS
             and s2 in LIST_ACCEPTABLE_STATES_FOR_HMS):
                 #these stellar attributes change value
-                merged_star.mass = (star_base.mass + comp.mass) * (1.-self.rel_mass_lost_HMS_HMS)
+                
                 #TODO for key in ["center_h1", "center_he4", "center_c12", "center_n14","center_o16"]:
                 merged_star.center_h1 = mass_weighted_avg()
                 merged_star.center_he4 = mass_weighted_avg(abundance_name = "center_he4")
@@ -196,6 +207,8 @@ class MergedStep(IsolatedStep):
                 merged_star.surface_c12 = mass_weighted_avg(abundance_name = "surface_c12")
                 merged_star.surface_n14 = mass_weighted_avg(abundance_name = "surface_n14")
                 merged_star.surface_o16 = mass_weighted_avg(abundance_name = "surface_o16")
+
+                merged_star.mass = (star_base.mass + comp.mass) * (1.-self.rel_mass_lost_HMS_HMS)
 
                 for key in STARPROPERTIES:
                     # these stellar attributes become np.nan
--- a/posydon/binary_evol/DT/step_merged.py
+++ b/posydon/binary_evol/DT/step_merged.py
@@ -383,7 +383,7 @@ class MergedStep(IsolatedStep):
                     merged_star.center_n14 = mass_weighted_avg(abundance_name = "center_n14", mass_weight1="he_core_mass")
                     merged_star.center_o16 = mass_weighted_avg(abundance_name = "center_o16", mass_weight1="he_core_mass")
                 elif (star_base.co_core_mass == 0 and comp.co_core_mass > 0): # star_base is the HeMS Star and comp has a CO core
-                    pass # the central abundances are kept as the ones of star_base
+                    pass # the central abundances are kept as the ones of comp
                 else:
                     Pwarn("weird compbination of CO core masses during merging", "EvolutionWarning")
                 # surface abundances from comp
@@ -498,40 +498,46 @@ class MergedStep(IsolatedStep):
                         if key in [ "c12_c12", "total_moment_of_inertia", "spin"]:
                             setattr(merged_star, key, np.nan)
                 merged_star.state = check_state_of_star(merged_star, star_CO=False)  # TODO for sure this needs testing!
-                massless_remnant = convert_star_to_massless_remnant(comp)
+                massless_remnant = convert_star_to_massless_remnant(star_base)
 
         # HeStar + HeStar
         elif (s1 in STAR_STATES_HE_RICH
             and s2 in STAR_STATES_HE_RICH):
-
-                # add total and core masses
-                for key in ["mass", "he_core_mass", "c_core_mass", "o_core_mass", "co_core_mass"]:
-                    current = getattr(star_base, key) + getattr(comp, key)
-                    setattr(merged_star, key,current)
 
                 # weighted central abundances if merging cores. Else only from star_base
                 if star_base.co_core_mass == 0 and comp.co_core_mass == 0: # two stars with Helium cores
                     merged_star.center_h1 = mass_weighted_avg(mass_weight1="he_core_mass")
                     merged_star.center_he4 = mass_weighted_avg(abundance_name = "center_he4", mass_weight1="he_core_mass")
                     merged_star.center_c12 = mass_weighted_avg(abundance_name = "center_c12", mass_weight1="he_core_mass")
+                    merged_star.avg_c_in_c_core = mass_weighted_avg(abundance_name = "avg_c_in_c_core", mass_weight1="he_core_mass")
                     merged_star.center_n14 = mass_weighted_avg(abundance_name = "center_n14", mass_weight1="he_core_mass")
                     merged_star.center_o16 = mass_weighted_avg(abundance_name = "center_o16", mass_weight1="he_core_mass")
+                    setattr(merged_star, "center_gamma", np.nan)
                 elif (star_base.co_core_mass > 0 and comp.co_core_mass == 0): # star_base with CO core and the comp has a He core
                     pass # the central abundances are kept as the ones of star_base
                 elif (comp.co_core_mass > 0 and star_base.co_core_mass == 0): # comp with CO core and the star_base has a He core
                     merged_star.center_h1 = comp.center_h1
                     merged_star.center_he4 = comp.center_he4
                     merged_star.center_c12 = comp.center_c12
+                    merged_star.avg_c_in_c_core = comp.avg_c_in_c_core
                     merged_star.center_n14 = comp.center_n14
                     merged_star.center_o16 = comp.center_o16
+                    merged_star.center_gamma =  comp.center_gamma
                 elif (star_base.co_core_mass > 0 and comp.co_core_mass > 0):
                     merged_star.center_h1 = mass_weighted_avg(mass_weight1="co_core_mass")
                     merged_star.center_he4 = mass_weighted_avg(abundance_name = "center_he4", mass_weight1="co_core_mass")
                     merged_star.center_c12 = mass_weighted_avg(abundance_name = "center_c12", mass_weight1="co_core_mass")
+                    merged_star.avg_c_in_c_core = mass_weighted_avg(abundance_name = "avg_c_in_c_core", mass_weight1="co_core_mass")
                     merged_star.center_n14 = mass_weighted_avg(abundance_name = "center_n14", mass_weight1="co_core_mass")
                     merged_star.center_o16 = mass_weighted_avg(abundance_name = "center_o16", mass_weight1="co_core_mass")
+                    setattr(merged_star, "center_gamma", np.nan)
                 else:
                     Pwarn("weird compbination of CO core masses during merging", "EvolutionWarning")
+
+                # add total and core masses
+                for key in ["mass", "he_core_mass", "c_core_mass", "o_core_mass", "co_core_mass"]:
+                    current = getattr(star_base, key) + getattr(comp, key)
+                    setattr(merged_star, key,current)
 
                 # weigheted mixing on the surface abundances based on the He-rich envelopes of the two stars
                 merged_star.surface_h1 = mass_weighted_avg(abundance_name = "surface_h1", mass_weight1="He-rich_envelope_mass", mass_weight2="He-rich_envelope_mass")
@@ -545,8 +551,7 @@ class MergedStep(IsolatedStep):
                     for substring in ["log_", "lg_", "surf_", "conv_", "lambda", "profile"]:
                         if (substring in key) :
                             setattr(merged_star, key, np.nan)
-                        if key in [ "c12_c12", "center_gamma",
-                                   "avg_c_in_c_core", "total_moment_of_inertia", "spin"]:
+                        if key in [ "c12_c12", "total_moment_of_inertia", "spin"]:
                             setattr(merged_star, key, np.nan)
                 merged_star.state = check_state_of_star(merged_star, star_CO=False)  # TODO for sure this needs testing!
                 massless_remnant = convert_star_to_massless_remnant(comp)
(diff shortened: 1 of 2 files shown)


posydon/binary_evol/DT/step_merged.py
```diff
@@ -168,13 +176,24 @@ def mass_weighted_avg(star1=star_base,star2=comp, abundance_name="center_h1", ma
                 M2 = getattr(star2, "he_core_mass") - getattr(star2, "co_core_mass")
             else:
                 M2 = getattr(star2, mass_weight2)
-            return (A1*M1 + A2*M2 ) / (M1+M2)
+
+            try:
+            	mass_weighted_avg_value=(A1*M1+A2*M2)/(M1+M2)
+            except TypeError:
+            	mass_weighted_avg_value= np.nan
```
astro-abhishek: Is setting the mass weighted average to np.nan going to cause issues later down the pipeline?

posydon/binary_evol/DT/step_merged.py
```diff
@@ -168,13 +176,24 @@ def mass_weighted_avg(star1=star_base,star2=comp, abundance_name="center_h1", ma
                 M2 = getattr(star2, "he_core_mass") - getattr(star2, "co_core_mass")
             else:
                 M2 = getattr(star2, mass_weight2)
-            return (A1*M1 + A2*M2 ) / (M1+M2)
+
+            try:
+            	mass_weighted_avg_value=(A1*M1+A2*M2)/(M1+M2)
+            except TypeError:
```
astro-abhishek: do we need to check for a division by zero error in case of zero envelope mass?

posydon/binary_evol/DT/track_match.py
```diff
@@ -285,7 +285,8 @@ def __init__(
                 "surface_he4",
                 "surface_h1",
                 "log_R",
-                "center_c12"
+                "center_c12",
+                "co_core_mass"
```
astro-abhishek, line 289: what if the co_core_mass doesn't exist? (there's a comment about something similar right above this)

maxbriel:
There are two testing binaries in the suite that change with this PR. 
@ezapartas can you confirm that these changes are expected from the change in the PR?

image


image

posydon/binary_evol/DT/step_merged.py
maxbriel: Is it possible to avoid a try/except?
Is it instead possible to verify the type of the values used?

posydon/binary_evol/DT/step_merged.py
```diff
@@ -514,34 +560,40 @@ def mass_weighted_avg(star1=star_base,star2=comp, abundance_name="center_h1", ma
...
-                # add total and core masses
-                for key in ["mass", "he_core_mass", "c_core_mass", "o_core_mass", "co_core_mass"]:
-                    current = getattr(star_base, key) + getattr(comp, "mass")
-                    setattr(merged_star, key,current)
+                # WD is considered a stripped CO core
+                # WD would always be the comp, it cannot be the engulfing star (so no need to do the opposite stars case below)
+
                 # weighted central abundances if merging cores. Else only from star_base
-                if (comp.co_core_mass > 0 and star_base.co_core_mass == 0): # comp with CO core and the star_base has not
+                if (comp.co_core_mass is not None and star_base.co_core_mass == 0): # comp with CO core and the star_base has not
```
maxbriel: Why is there this change from values to None? It happens later too.

Are there specific binaries that crash because the co_core_mass is undefined when reaching this stage?

Same n line 575.

posydon/binary_evol/DT/step_merged.py
maxbriel: Can we remove the old code here?

posydon/binary_evol/DT/step_merged.py
```diff
@@ -168,13 +176,24 @@ def mass_weighted_avg(star1=star_base,star2=comp, abundance_name="center_h1", ma
...
+
+            return mass_weighted_avg_value
 
         # MS + MS
         if ( s1 in LIST_ACCEPTABLE_STATES_FOR_HMS
             and s2 in LIST_ACCEPTABLE_STATES_FOR_HMS):
                 #these stellar attributes change value
-                merged_star.mass = (star_base.mass + comp.mass) * (1.-self.rel_mass_lost_HMS_HMS)
+
                 #TODO for key in ["center_h1", "center_he4", "center_c12", "center_n14","center_o16"]:
```
maxbriel: is this TODO still needed?

posydon/binary_evol/DT/step_merged.py
```diff
@@ -343,75 +373,82 @@ def mass_weighted_avg(star1=star_base,star2=comp, abundance_name="center_h1", ma
...
                     for substring in ["log_", "lg_", "surf_", "conv_", "lambda", "profile"]:
                         if (substring in key) :
                             setattr(merged_star, key, np.nan)
-                        if key in [ "c12_c12", "center_gamma",
-                                   "avg_c_in_c_core", "total_moment_of_inertia", "spin"]:
+                        if key in [ "c12_c12", "total_moment_of_inertia", "spin"]:
                             setattr(merged_star, key, np.nan)
 
                 merged_star.state = check_state_of_star(merged_star, star_CO=False)  # TODO for sure this needs testing!
-                massless_remnant = convert_star_to_massless_remnant(comp)
```
maxbriel, line 373: Why are we converting `star_base` to the massless remnant?

We are already switching the stars before going into this function depending on which one initiated RLO, such that the donor is always `star_base`.

posydon/binary_evol/DT/step_merged.py
```diff
@@ -343,75 +373,82 @@ def mass_weighted_avg(star1=star_base,star2=comp, abundance_name="center_h1", ma
         elif (s1 in  LIST_ACCEPTABLE_STATES_FOR_HeMS
             and s2 in LIST_ACCEPTABLE_STATES_FOR_POSTMS):
 
-                # add total and core masses
-                for key in ["mass", "he_core_mass", "c_core_mass", "o_core_mass", "co_core_mass"]:
-                    current = getattr(star_base, key) + getattr(comp, key)
-                    setattr(merged_star, key,current)
+                merged_star = comp
```
maxbriel, line 482: Why are we changing what the merged star is here?

This is different than the definition of which one starts the RLO. This will make the companion the remaining star.

posydon/binary_evol/DT/step_merged.py
```diff
@@ -420,25 +457,25 @@ def mass_weighted_avg(star1=star_base,star2=comp, abundance_name="center_h1", ma
         elif (s1 in LIST_ACCEPTABLE_STATES_FOR_POSTHeMS
             and s2 in LIST_ACCEPTABLE_STATES_FOR_POSTMS):
 
-                # add total and core masses
-                for key in ["mass", "he_core_mass", "c_core_mass", "o_core_mass", "co_core_mass"]:
-                    current = getattr(star_base, key) + getattr(comp, key)
-                    setattr(merged_star, key,current)
+                merged_star = comp
```
maxbriel, line 581: Change in what star becomes the merged star. Same question as before.

posydon/binary_evol/DT/step_merged.py
```diff
@@ -448,50 +485,60 @@ def mass_weighted_avg(star1=star_base,star2=comp, abundance_name="center_h1", ma
...
                     for substring in ["log_", "lg_", "surf_", "conv_", "lambda", "profile"]:
                         if (substring in key) :
                             setattr(merged_star, key, np.nan)
-                        if key in [ "c12_c12", "center_gamma",
-                                   "avg_c_in_c_core", "total_moment_of_inertia", "spin"]:
+                        if key in [ "c12_c12", "total_moment_of_inertia", "spin"]:
                             setattr(merged_star, key, np.nan)
                 merged_star.state = check_state_of_star(merged_star, star_CO=False)  # TODO for sure this needs testing!
-                massless_remnant = convert_star_to_massless_remnant(comp)
+                massless_remnant = convert_star_to_massless_remnant(star_base)
```
maxbriel, line 632: Same as before.

posydon/binary_evol/DT/step_merged.py
ezapartas: Having both M1 and M2  = 0 would be weird, but I added the ZeroDivisionError case, in the  the try-except loop.

posydon/binary_evol/DT/step_merged.py
ezapartas: 1) @astro-abhishek  the np.nan may cause issues later, but I think is the most pythonic way of saying "the function was called byt physically there are issues at the calculation so we have no computed result, we don't know the value". If the question is whether it will make it crash, I am not sure, I think not, as np.nan is accepted as a value in calculations, it will just produce more "unknown" np.nan quantities wherever it is used in a calcualation.

2) @maxbriel The following is a different way more exact checking, instead of try-except loop. Is this what you are thinking? If did not add it yet, but if you agree, please just add it as a commit directly. 

den = M1 + M2
if not (np.isfinite(A1) and np.isfinite(A2) and np.isfinite(M1) and np.isfinite(M2)):
    mass_weighted_avg_value = np.nan
elif den == 0:
    mass_weighted_avg_value = np.nan
else:
    mass_weighted_avg_value = (A1*M1 + A2*M2) / den
    
------------------
In both cases, if you have better coding ideas for this part (or any other) feel free to add them directly.

posydon/binary_evol/DT/step_merged.py
ezapartas: Hmm, I don' remember exactly the reasoning. I think I thought that instead of raising a TypeError in case of comp.co_core_mass = None and crashing, I preferred it to go to the else: Pwarn("weird compbination of CO core masses during merging", "EvolutionWarning"). 

But again, if you think of a better handling it , feel free to directly change it back

posydon/binary_evol/DT/step_merged.py
ezapartas: No, you are right, I took the line out

posydon/binary_evol/DT/step_merged.py
ezapartas: Hmm, actually this occurs even at line 256. Indeed, it "violates" the engulfind star (that starts the RLO) is our base of the merger, making the base to be the star with the most layers. So for example in a postMS+HMS merger, the postMS star becomes the base always. Same between postMS+HeMS. Even if it is not the postMS that engulfs. This is because the final merged star with have a more complex layering (envelope, core, etc). Of course in most cases it will indeed be the postMS star that is also initiating the RLO and the engulfing, but we allow for the other case too, giving the same outcome. 

If needed we can restructure the code to remake the calculation using the engulfing always as the base, but this is only for the sake of clarity. It will not change the merged outcome in the end. 

summary: The outcome in the end should not depend on which star engulfs (unless omeone argues otherwise). it is just easier to start the merger construction from that star in most cases, as it usually is the more "complex" star in the perspective of layers, i.e. it is a giant star

posydon/binary_evol/DT/track_match.py
ezapartas: Hmm, this list assumes that the single star models and the respective EEPs have this column. If a user uses their one sinlge star and EEPs, not run with POSYDON, so not calculating a co_core (having only a c_core or an o_core for example), there may be a mismatch. For our POSYDON runs this is ok for now. If we want to generalize it so much, I think much of the structure should change anyway. I would leave it for now, keeping the above comment there for future reference.

ezapartas:
@astro-abhishek @maxbriel Thanks for the detailed comments. I tried to answer all of them and address whichever where doable. Feel free to directly change the coding way if you think a better option exists. 

Note as soon as this PR is accepted, it is better to be pulled by the #788 before that PR is evaluated and merged.

posydon/binary_evol/DT/step_merged.py
```diff
@@ -232,59 +252,62 @@ def mass_weighted_avg(star1=star_base,star2=comp, abundance_name="center_h1", ma
...
-
                 # weighted central abundances if merging cores. Else only from star_base
                 if star_base.co_core_mass == 0 and comp.co_core_mass == 0: # two stars with Helium cores
                     merged_star.center_h1 = mass_weighted_avg(mass_weight1="he_core_mass")
                     merged_star.center_he4 = mass_weighted_avg(abundance_name = "center_he4", mass_weight1="he_core_mass")
                     merged_star.center_c12 = mass_weighted_avg(abundance_name = "center_c12", mass_weight1="he_core_mass")
+                    merged_star.avg_c_in_c_core = mass_weighted_avg(abundance_name = "avg_c_in_c_core", mass_weight1="he_core_mass")
                     merged_star.center_n14 = mass_weighted_avg(abundance_name = "center_n14", mass_weight1="he_core_mass")
                     merged_star.center_o16 = mass_weighted_avg(abundance_name = "center_o16", mass_weight1="he_core_mass")
+                    setattr(merged_star, "center_gamma", np.nan)
```
astroJeff, line 383: Can we not use the parameters of the helium stars to set `center_gamma`?

posydon/binary_evol/DT/step_merged.py
```diff
@@ -232,59 +252,62 @@ def mass_weighted_avg(star1=star_base,star2=comp, abundance_name="center_h1", ma
...
                     merged_star.center_o16 = comp.center_o16
+                    merged_star.center_gamma =  comp.center_gamma
                 elif (star_base.co_core_mass > 0 and comp.co_core_mass > 0):
                     merged_star.center_h1 = mass_weighted_avg(mass_weight1="co_core_mass") #TODO : maybe he_core_mass makes more sense?
                     merged_star.center_he4 = mass_weighted_avg(abundance_name = "center_he4", mass_weight1="co_core_mass")
                     merged_star.center_c12 = mass_weighted_avg(abundance_name = "center_c12", mass_weight1="co_core_mass")
+                    merged_star.avg_c_in_c_core = mass_weighted_avg(abundance_name = "avg_c_in_c_core", mass_weight1="co_core_mass")
                     merged_star.center_n14 = mass_weighted_avg(abundance_name = "center_n14", mass_weight1="co_core_mass")
                     merged_star.center_o16 = mass_weighted_avg(abundance_name = "center_o16", mass_weight1="co_core_mass")
+                    setattr(merged_star, "center_gamma", np.nan)
```
astroJeff, line 411: Same comment as above about `center_gamma`

posydon/binary_evol/DT/step_merged.py
```diff
@@ -311,30 +337,32 @@ def mass_weighted_avg(star1=star_base,star2=comp, abundance_name="center_h1", ma
...
-
                 # weighted central abundances if merging cores. Else only from star_base
                 if star_base.co_core_mass == 0 and comp.co_core_mass == 0: # two stars with only Helium cores, not CO cores
                     merged_star.center_h1 = mass_weighted_avg(mass_weight1="he_core_mass")
                     merged_star.center_he4 = mass_weighted_avg(abundance_name = "center_he4", mass_weight1="he_core_mass")
                     merged_star.center_c12 = mass_weighted_avg(abundance_name = "center_c12", mass_weight1="he_core_mass")
+                    merged_star.avg_c_in_c_core = mass_weighted_avg(abundance_name = "avg_c_in_c_core", mass_weight1="he_core_mass")
                     merged_star.center_n14 = mass_weighted_avg(abundance_name = "center_n14", mass_weight1="he_core_mass")
                     merged_star.center_o16 = mass_weighted_avg(abundance_name = "center_o16", mass_weight1="he_core_mass")
+                    setattr(merged_star, "center_gamma", np.nan)
```
astroJeff, line 456: Same comment as above. Why set to `nan` instead of grabbing the value from either `star_base` or `comp`?

posydon/binary_evol/DT/step_merged.py
```diff
@@ -343,75 +371,82 @@ def mass_weighted_avg(star1=star_base,star2=comp, abundance_name="center_h1", ma
...
-
                 # weighted central abundances if merging cores. Else only from star_base
                 if star_base.co_core_mass == 0 and comp.co_core_mass == 0: # two stars with Helium cores
                     merged_star.center_h1 = mass_weighted_avg(mass_weight1="he_core_mass")
                     merged_star.center_he4 = mass_weighted_avg(abundance_name = "center_he4", mass_weight1="he_core_mass")
                     merged_star.center_c12 = mass_weighted_avg(abundance_name = "center_c12", mass_weight1="he_core_mass")
+                    merged_star.avg_c_in_c_core = mass_weighted_avg(abundance_name = "avg_c_in_c_core", mass_weight1="he_core_mass")
                     merged_star.center_n14 = mass_weighted_avg(abundance_name = "center_n14", mass_weight1="he_core_mass")
                     merged_star.center_o16 = mass_weighted_avg(abundance_name = "center_o16", mass_weight1="he_core_mass")
+                    setattr(merged_star, "center_gamma", np.nan)
```
astroJeff, line 534: See above

posydon/binary_evol/DT/step_merged.py
```diff
@@ -343,75 +371,82 @@ def mass_weighted_avg(star1=star_base,star2=comp, abundance_name="center_h1", ma
...
                     merged_star.center_o16 = comp.center_o16
+                    merged_star.center_gamma =  comp.center_gamma
                 elif (star_base.co_core_mass > 0 and comp.co_core_mass > 0):
                     merged_star.center_h1 = mass_weighted_avg(mass_weight1="co_core_mass")
                     merged_star.center_he4 = mass_weighted_avg(abundance_name = "center_he4", mass_weight1="co_core_mass")
                     merged_star.center_c12 = mass_weighted_avg(abundance_name = "center_c12", mass_weight1="co_core_mass")
+                    merged_star.avg_c_in_c_core = mass_weighted_avg(abundance_name = "avg_c_in_c_core", mass_weight1="co_core_mass")
                     merged_star.center_n14 = mass_weighted_avg(abundance_name = "center_n14", mass_weight1="co_core_mass")
                     merged_star.center_o16 = mass_weighted_avg(abundance_name = "center_o16", mass_weight1="co_core_mass")
+                    setattr(merged_star, "center_gamma", np.nan)
```
astroJeff, line 560: see above

posydon/binary_evol/DT/step_merged.py
```diff
@@ -420,25 +455,25 @@ def mass_weighted_avg(star1=star_base,star2=comp, abundance_name="center_h1", ma
...
+
                 # weighted central abundances if merging cores. Else only from star_base
                 if star_base.co_core_mass == 0 and comp.co_core_mass == 0: # two stars with Helium cores
                     merged_star.center_h1 = mass_weighted_avg(mass_weight1="he_core_mass")
                     merged_star.center_he4 = mass_weighted_avg(abundance_name = "center_he4", mass_weight1="he_core_mass")
                     merged_star.center_c12 = mass_weighted_avg(abundance_name = "center_c12", mass_weight1="he_core_mass")
+                    merged_star.avg_c_in_c_core = mass_weighted_avg(abundance_name = "avg_c_in_c_core", mass_weight1="he_core_mass")
                     merged_star.center_n14 = mass_weighted_avg(abundance_name = "center_n14", mass_weight1="he_core_mass")
                     merged_star.center_o16 = mass_weighted_avg(abundance_name = "center_o16", mass_weight1="he_core_mass")
+                    setattr(merged_star, "center_gamma", np.nan)
```
astroJeff, line 596: see above

posydon/binary_evol/DT/step_merged.py
```diff
@@ -448,50 +483,60 @@ def mass_weighted_avg(star1=star_base,star2=comp, abundance_name="center_h1", ma
...
-
                 # weighted central abundances if merging cores. Else only from star_base
                 if star_base.co_core_mass == 0 and comp.co_core_mass == 0: # two stars with Helium cores
                     merged_star.center_h1 = mass_weighted_avg(mass_weight1="he_core_mass")
                     merged_star.center_he4 = mass_weighted_avg(abundance_name = "center_he4", mass_weight1="he_core_mass")
                     merged_star.center_c12 = mass_weighted_avg(abundance_name = "center_c12", mass_weight1="he_core_mass")
+                    merged_star.avg_c_in_c_core = mass_weighted_avg(abundance_name = "avg_c_in_c_core", mass_weight1="he_core_mass")
                     merged_star.center_n14 = mass_weighted_avg(abundance_name = "center_n14", mass_weight1="he_core_mass")
                     merged_star.center_o16 = mass_weighted_avg(abundance_name = "center_o16", mass_weight1="he_core_mass")
+                    setattr(merged_star, "center_gamma", np.nan)
```
astroJeff, line 650: see above

posydon/binary_evol/DT/step_merged.py
```diff
@@ -448,50 +483,60 @@ def mass_weighted_avg(star1=star_base,star2=comp, abundance_name="center_h1", ma
...
                     merged_star.center_o16 = comp.center_o16
+                    merged_star.center_gamma =  comp.center_gamma
                 elif (star_base.co_core_mass > 0 and comp.co_core_mass > 0):
                     merged_star.center_h1 = mass_weighted_avg(mass_weight1="co_core_mass")
                     merged_star.center_he4 = mass_weighted_avg(abundance_name = "center_he4", mass_weight1="co_core_mass")
                     merged_star.center_c12 = mass_weighted_avg(abundance_name = "center_c12", mass_weight1="co_core_mass")
+                    merged_star.avg_c_in_c_core = mass_weighted_avg(abundance_name = "avg_c_in_c_core", mass_weight1="co_core_mass")
                     merged_star.center_n14 = mass_weighted_avg(abundance_name = "center_n14", mass_weight1="co_core_mass")
                     merged_star.center_o16 = mass_weighted_avg(abundance_name = "center_o16", mass_weight1="co_core_mass")
+                    setattr(merged_star, "center_gamma", np.nan)
```
astroJeff, line 675: see above

astroJeff:
Can we spin off some of this into new functions? There is quite a bit of repeated code

posydon/binary_evol/DT/step_merged.py
ezapartas: @astroJeff I think all the cases that I put center_gamma==np.nan is where we have mixxing of the central core layer of the merged project. E.g. here the two helium cores are mixed. I refrained from saying the degeneracy level of a mixed layer. In any case, for now this is erased in the mathching of the merger part that continues after. It becomes important only we want ot use the central_gamma as a matching parameter which I assume we will refrain from.

maxbriel:
Given some discussion we've had during the dev-meeting, I'm refactoring some of the functions here to make them testable and easier to maintain at a later date

maxbriel:
Here's the output of the branch currently:
[manos merger copiedattributes](https://github.com/user-attachments/files/25979101/evolve_binaries_manos_merger_copiedattributes.txt)

[main](https://github.com/user-attachments/files/25976461/evolve_binaries_main.txt)

The diff: https://www.diffchecker.com/7IurVLUA/

posydon/binary_evol/DT/step_merged.py
```diff
@@ -126,423 +127,585 @@ def __call__(self,binary):
...
+            if binary.event == 'oMerging1':
+                print("M_merged , he_core_mass merged, co_core_mass merged: ", binary.star_1.mass,
+                  binary.star_1.he_core_mass, binary.star_1.co_core_mass)
+                print("star_1.center_h1, star_1.center_he4, star_1.center_c12, star_1.surface_he4: ",
+                  binary.star_1.center_h1, binary.star_1.center_he4, binary.star_1.center_c12, binary.star_1.surface_he4)
+            elif binary.event=='oMerging2':
+                print("M_merged , he_core_mass merged, co_core_mass merged: ", binary.star_2.mass,
+                  binary.star_2.he_core_mass, binary.star_2.co_core_mass)
+                print("star_2.center_h1, star_2.center_he4, star_2.center_c12, star_2.surface_he4: ",
+                  binary.star_2.center_h1, binary.star_2.center_he4, binary.star_2.center_c12, binary.star_2.surface_he4)
```
sgossage:
```suggestion
            elif binary.event=='oMerging2':
                print(f"M_merged = {binary.star_2.mass}\n"
                      f"he_core_mass merged = {binary.star_2.he_core_mass}\n"
                      f"co_core_mass merged = {binary.star_2.co_core_mass}\n")
                print(f"star_2.center_h1 = {binary.star_2.center_h1}\n"
                      f"star_2.center_he4 = {binary.star_2.center_he4}\n"
                      f"star_2.center_c12 = {binary.star_2.center_c12}\n"
                      f"star_2.surface_he4 = {binary.star_2.surface_he4}\n")
```

posydon/binary_evol/DT/step_merged.py
```diff
@@ -126,423 +127,585 @@ def __call__(self,binary):
...
                   binary.state, binary.event)
-            print("M_merged , he_core_mass merged: ", binary.star_1.mass,
-                  binary.star_1.he_core_mass)
-            print("star_1.center_he4, star_1.surface_he4: ",
-                  binary.star_1.center_he4, binary.star_1.surface_he4)
+            if binary.event == 'oMerging1':
+                print("M_merged , he_core_mass merged, co_core_mass merged: ", binary.star_1.mass,
+                  binary.star_1.he_core_mass, binary.star_1.co_core_mass)
+                print("star_1.center_h1, star_1.center_he4, star_1.center_c12, star_1.surface_he4: ",
+                  binary.star_1.center_h1, binary.star_1.center_he4, binary.star_1.center_c12, binary.star_1.surface_he4)
```
sgossage:
```suggestion
            if binary.event == 'oMerging1':
                print(f"M_merged = {binary.star_1.mass}\n"
                      f"he_core_mass merged = {binary.star_1.he_core_mass}\n"
                      f"co_core_mass merged = {binary.star_1.co_core_mass}\n")
                print(f"star_1.center_h1 = {binary.star_1.center_h1}\n"
                      f"star_1.center_he4 = {binary.star_1.center_he4}\n"
                      f"star_1.center_c12 = {binary.star_1.center_c12}\n"
                      f"star_1.surface_he4 = {binary.star_1.surface_he4}\n")
```

posydon/binary_evol/DT/step_merged.py
```diff
@@ -126,423 +127,585 @@ def __call__(self,binary):
         if self.verbose:
             print("After Merger", binary.star_1.state, binary.star_2.state,
                   binary.state, binary.event)
```
sgossage:
```suggestion
            print("After Merger:\n"
                  f"_____________\n"
                  f"star_1.state = {binary.star_1.state}\n"
                  f"star_2.state = {binary.star_2.state}\n"
                  f"binary.state = {binary.state}\n"
                  f"binary.event = {binary.event}\n")
```

posydon/binary_evol/DT/step_merged.py
```diff
@@ -93,12 +93,13 @@ def __call__(self,binary):
...
-                  binary.star_1.he_core_mass, binary.star_2.he_core_mass)
-            print("star_1.center_he4, star_2.center_he4, star_1.surface_he4, "
-                  "star_2.surface_he4: ", binary.star_1.center_he4,
-                  binary.star_2.center_he4, binary.star_1.surface_he4,
+                  binary.star_1.he_core_mass, binary.star_2.he_core_mass, binary.star_1.co_core_mass, binary.star_2.co_core_mass)
+            print("star_1.center_h1, star_2.center_h1, star_1.center_he4, star_1.center_c12, star_2.center_c12, "
+                  "star_2.center_he4, star_1.surface_he4, star_2.surface_he4: ",
+                  binary.star_1.center_h1, binary.star_2.center_h1, binary.star_1.center_he4,
+                  binary.star_2.center_he4, binary.star_1.center_c12, binary.star_2.center_c12, binary.star_1.surface_he4,
                   binary.star_2.surface_he4)
```
sgossage:
```suggestion
            print(f"star_1.center_h1 = {binary.star_1.center_h1}\n"
                  f"star_2.center_h1 = {binary.star_2.center_h1}\n"
                  f"star_1.center_he4 = {binary.star_1.center_he4}\n"
                  f"star_2.center_he4 = {binary.star_2.center_he4}\n"
                  f"star_1.center_c12 = {binary.star_1.center_c12}\n"
                  f"star_2.center_c12 = {binary.star_2.center_c12}\n"
                  f"star_1.surface_he4 = {binary.star_1.surface_he4}\n"
                  f"star_2.surface_he4 = {binary.star_2.surface_he4}\n"
                  f"_____________\n")
```

posydon/binary_evol/DT/step_merged.py
```diff
@@ -93,12 +93,13 @@ def __call__(self,binary):
...
             print("Before Merger", binary.star_1.state, binary.star_2.state,
                   binary.state, binary.event)
-            print("M1 , M2, he_core_mass1, he_core_mass2: ",
+            print("M1 , M2, he_core_mass1, he_core_mass2, co_core_mass1, co_core_mass2: ",
                   binary.star_1.mass, binary.star_2.mass,
-                  binary.star_1.he_core_mass, binary.star_2.he_core_mass)
-            print("star_1.center_he4, star_2.center_he4, star_1.surface_he4, "
-                  "star_2.surface_he4: ", binary.star_1.center_he4,
-                  binary.star_2.center_he4, binary.star_1.surface_he4,
+                  binary.star_1.he_core_mass, binary.star_2.he_core_mass, binary.star_1.co_core_mass, binary.star_2.co_core_mass)
```
sgossage:
```suggestion
            print(f"M1 = {binary.star_1.mass}\n"
                  f"M2 = {binary.star_2.mass}\n"
                  f"he_core_mass1 = {binary.star_1.he_core_mass}\n"
                  f"he_core_mass2 = {binary.star_2.he_core_mass}\n"
                  f"co_core_mass1 = {binary.star_1.co_core_mass}\n"
                  f"co_core_mass2 = {binary.star_2.co_core_mass}\n")
```

posydon/binary_evol/DT/step_merged.py
```diff
@@ -93,12 +93,13 @@ def __call__(self,binary):
         if self.verbose:
             print("Before Merger", binary.star_1.state, binary.star_2.state,
                   binary.state, binary.event)
```
sgossage:
```suggestion
            print("Before Merger:\n"
                  f"_____________\n"
                  f"star_1.state = {binary.star_1.state}\n"
                  f"star_2.state = {binary.star_2.state}\n"
                  f"binary.state = {binary.state}\n"
                  f"binary.event = {binary.event}\n")
```

posydon/binary_evol/DT/step_merged.py
```diff
@@ -126,423 +127,585 @@ def __call__(self,binary):
...
+        if not (np.isfinite(A1) and np.isfinite(A2) and np.isfinite(M1) and np.isfinite(M2)):
+            mass_weighted_avg_value = np.nan
+        elif den == 0:
+            mass_weighted_avg_value = np.nan
+        else:
+            mass_weighted_avg_value = (A1 * M1 + A2 * M2) / den
+
+        if self.verbose:
+            print(abundance_name, mass_weight1, mass_weight2)
+            print("A_base, M_base_abundance, A_comp, M_comp_abundance", A1, M1, A2, M2)
```
sgossage: Should we say `*_base` or e.g., `*_1` or `*_pri`? I think other parts of the code use one of the latter. Also, should `M_*_abundance` be something like `M_*_mass` because M1/M2 hold masses?

posydon/binary_evol/DT/step_merged.py
```diff
@@ -126,423 +127,585 @@ def __call__(self,binary):
...
-                for key in STARPROPERTIES:
-                    # these stellar attributes become np.nan
-                    for substring in ["log_", "lg_", "surf_", "conv_", "lambda", "profile"]:
-                        if (substring in key) :
-                            setattr(merged_star, key, np.nan)
-                        if key in [ "c12_c12", "center_gamma",
-                                   "avg_c_in_c_core", "total_moment_of_inertia", "spin"]:
-                            setattr(merged_star, key, np.nan)
+                # Set parameters that are not expected to be meaningful after a merger to np.nan
+                self._apply_nan_attributes(merged_star)
```
sgossage, line 327: Some of this is repeated code in this `if` chain. I suggest generalizing this and creating a helper function for clarity/brevity that is called here and elsewhere, as appropriate.

posydon/binary_evol/DT/step_merged.py
```diff
@@ -126,423 +127,596 @@ def __call__(self,binary):
...
+            mass_weighted_avg_value = np.nan
+        elif den == 0:
+            mass_weighted_avg_value = np.nan
+        else:
+            mass_weighted_avg_value = (A1 * M1 + A2 * M2) / den
+
+        if self.verbose:
+            print(abundance_name, mass_weight1, mass_weight2)
+            print("A_base, M_base_abundance, A_comp, M_comp_abundance", A1, M1, A2, M2)
+            print("mass_weighted_avg= ", mass_weighted_avg_value)
```
sgossage:
```suggestion
        if self.verbose:
            print(f"Mass weighted average for {abundance_name}:\n"
                  f"weight 1: {mass_weight1}\n"
                  f"weight 2: {mass_weight2}\n"
                  f"A_base = {A1}\n"
                  f"M_base_abundance = {M1}\n"
                  f"A_comp = {A2}\n"
                  f"M_comp_abundance = {M2}\n"
                  f"mass_weighted_avg = {mass_weighted_avg_value}\n")
```

posydon/binary_evol/DT/step_merged.py
sgossage: This line (`binary.event = None`) means that the `print` statements based on `binary.event` (e.g., `if binary.event == 'oMerging1':`) below never fire when `self.verbose = True`. Should this be set to `None` after the verbose print? Is this line needed?

posydon/binary_evol/DT/step_merged.py
```diff
@@ -514,34 +560,40 @@ def mass_weighted_avg(star1=star_base,star2=comp, abundance_name="center_h1", ma
...
                 for key in STARPROPERTIES:
                     # these stellar attributes become np.nan
                     for substring in ["log_", "lg_", "surf_", "conv_", "lambda", "profile"]:
                         if (substring in key) :
                             setattr(merged_star, key, np.nan)
-                        if key in [ "c12_c12", "center_gamma",
-                                   "avg_c_in_c_core", "total_moment_of_inertia", "spin"]:
+                        if key in [ "c12_c12", "total_moment_of_inertia", "spin"]:
                             setattr(merged_star, key, np.nan)
                 merged_star.state = check_state_of_star(merged_star, star_CO=False)  # TODO for sure this needs testing!
```
sgossage, line 742: The comment says that this for sure needs to be tested -- has it been?

posydon/binary_evol/DT/step_merged.py
sgossage: I commented the `binary.event = None` line for now.

sgossage:
I was working on generalizing the logic and creating functions to apply these merging rules and noticed something.

For post-HMS mergers, we have this logic (in `step_merged.py`):

```
                      ...
385                # star_base with CO core and the comp has a He core
386                elif (star_base.co_core_mass > 0 and comp.co_core_mass == 0):
387                    pass # the central abundances are kept as the ones of star_base
                      ...
```

whereas for HMS-HMS mergers, we do not apply the same rule. Should we?

The following binary evolution from our test suite contains an HMS-HMS merger where the base star has a non-zero CO core mass while the companion has no CO core.

```
star_1= SingleStar(**{'mass': 9.845907 , 'state': 'H-rich_Core_H_burning','metallicity':1,
            'natal_kick_array':  [0.0, 4.190728383757787, 1.1521129697118118, 5.015343794234789]})
star_2 = SingleStar(**{'mass': 9.611029, 'state': 'H-rich_Core_H_burning','metallicity':1,
                'natal_kick_array': [0.0, 0.0, 0.0, 0.0]})
                        'orbital_period':  3.820571,'eccentricity': 0.0}, properties = sim_prop)
```

(The branch that I'm working on the logic in is `sg_merger_copiedattributes`. After we sort this out, if the changes there look good and the logic looks correct, we could merge it in to this branch. I've tested it with the test suite, but noticed that this binary produces a different outcome compared to `manos_merger_copiedattribues` because of this issue and how I currently wrote the logic.)

maxbriel:
> I was working on generalizing the logic and creating functions to apply these merging rules and noticed something.
> 
> For post-HMS mergers, we have this logic (in `step_merged.py`):
> 
> ```
>                       ...
> 385                # star_base with CO core and the comp has a He core
> 386                elif (star_base.co_core_mass > 0 and comp.co_core_mass == 0):
> 387                    pass # the central abundances are kept as the ones of star_base
>                       ...
> ```
> 
> whereas for HMS-HMS mergers, we do not apply the same rule. Should we?
> 
> The following binary evolution from our test suite contains an HMS-HMS merger where the base star has a non-zero CO core mass while the companion has no CO core.
> 
> ```
> star_1= SingleStar(**{'mass': 9.845907 , 'state': 'H-rich_Core_H_burning','metallicity':1,
>             'natal_kick_array':  [0.0, 4.190728383757787, 1.1521129697118118, 5.015343794234789]})
> star_2 = SingleStar(**{'mass': 9.611029, 'state': 'H-rich_Core_H_burning','metallicity':1,
>                 'natal_kick_array': [0.0, 0.0, 0.0, 0.0]})
> binary = BinaryStar(star_1, star_2, **{'time': 0.0,  'state': 'detached',  'event': 'ZAMS',
>                         'orbital_period':  3.820571,'eccentricity': 0.0}, properties = sim_prop)
> ```
> 
> (The branch that I'm working on the logic in is `sg_merger_copiedattributes`. After we sort this out, if the changes there look good and the logic looks correct, we could merge it in to this branch. I've tested it with the test suite, but noticed that this binary produces a different outcome compared to `manos_merger_copiedattribues` because of this issue and how I currently wrote the logic.)

This is strange. A HMS-HMS merger should **not** have a CO core or He core defined in either component, so the same rule shouldn't be applied.
It's concerning that something with a CO core is being put into the HMS-HMS category. A few clarifying questions: Is the stellar state being updated correctly during the evolution when entering the step? Is it still burning central hydrogen? And how large is the "CO core"?

posydon/binary_evol/DT/step_merged.py
sgossage: working on this here: [`sg_merger_copiedattributes`](https://github.com/POSYDON-code/POSYDON/blob/sg_merger_copiedattributes/posydon/binary_evol/DT/step_merged.py)

sgossage:
> It's concerning that something with a CO core is being put into the HMS-HMS category. A few clarifying questions: Is the stellar state being updated correctly during the evolution when entering the step? Is it still burning central hydrogen? And how large is the "CO core"?

@maxbriel this is the debugging output prior to merging:

```
Before Merger:
_____________

M1 = 4.212749495765598
M2 = 10.95210897562295
he_core_mass1 = 0.6365034990860114
he_core_mass2 = 2.019788928314233
co_core_mass1 = 0.10058851260570133
co_core_mass2 = 0.0

star_1.center_h1 = 0.03464608832551165
star_2.center_h1 = 0.016979750448212434
star_1.center_he4 = 0.8990294555960494
star_2.center_he4 = 0.9691717446871981
star_1.center_c12 = 0.05079687439526389
star_2.center_c12 = 9.913536056802633e-05
star_1.surface_he4 = 0.3547279661505059
star_2.surface_he4 = 0.32219608838969765
```

Star 1 is classed as `H-rich_Core_H-burning` by our criteria (`log_LH = 3.4`, `surface_h1 = 0.6`). I was able to trace it back to IF interpolation giving the star the 0.1 Msol CO core in `step_HMS_HMS`. The masses and age from IF interpolation seem reasonable, and it is classed as unstable MT (reasonable for that region of the grid), but somehow it determines a non-zero CO core mass, which is not reflected by our unstable MT models in the grids. 

In the grids, this system may lie near a boundary between unstable MT and unstable reverse MT. While the unstable MT models have no CO core, the unstable reverse MT models do. These should be in different interpolation classes though?

I noticed that it seems we only have an `unstable_MT` interpolation class, and no `unstable_reverse_MT`. Could this be related to the issue? (@philipp-rajah)

At q = 0.95, the closest neighbors to this system are in the red boxed region:

q0 95_region

and at q = 0.99:

q1_region

(The axes in the plot are showing you the linear values of Porb and mass, not log10 values).

astroJeff (approved):

sgossage (approved):

elizabethteng (approved):

PR #786, opened 2026-01-15
[enhancement] Maltsev prescription

ezapartas:
Implementing Maltsev explodablity criterion:

As an alternative to Patton20+Ertl16 criterion, now we have Patton20+Maltsev25 criterion 

For now, implementation only for M_CO <= 10 Msun (as in Patton20+Ertl16 criterion). Still needed to implement the case for M_CO > 10 Msun (where the MT history is needed right now).

To work it still needs:
- to include the Kepler_sc.dat table as in the POSYDON data of Patton, so that it is able to be read.
- to make new SN MODELS based on this prescription for initial-final interpolation


posydon/binary_evol/SN/step_SN.py
```diff
@@ -2343,6 +2384,153 @@ def Patton20_corecollapse(self, star, engine, conserve_hydrogen_envelope=False):
...
+                    if rem==2:  # successful SN with NS
+                        m_rem = M4
+                        f_fb = 0.0
+                        state = 'NS'
+                    else:  # successful SN but with fallback BH
+                        if conserve_hydrogen_envelope:
+                            m_rem = star.mass
+                        else:
+                            m_rem = star.he_core_mass
+                            f_fb = 0.99
```
maxbriel: Are this one and the state not supposed to be set in the context of conserve_hydrogen_envelope too?

posydon/binary_evol/SN/step_SN.py
```diff
@@ -2343,6 +2384,153 @@ def Patton20_corecollapse(self, star, engine, conserve_hydrogen_envelope=False):
...
+                if conserve_hydrogen_envelope:
+                    m_rem = star.mass
+                else:
+                    m_rem = star.he_core_mass
+                f_fb = 1.0
+                state = 'BH'
+
+            elif (CO_core_mass < 10.0) and (CO_core_mass > 2.5):
+                ff = self.explod_crit(Xi, sc, mu4M4, mu4)
+                if ff=0:
```
maxbriel: The syntax here is incorrect. is is supposed to be?

```suggestion
                if ff==0:
```

posydon/binary_evol/SN/step_SN.py
maxbriel: Same in the other sections of conserve_hydrogen_envelope

posydon/binary_evol/SN/step_SN.py
```diff
@@ -2343,6 +2384,153 @@ def Patton20_corecollapse(self, star, engine, conserve_hydrogen_envelope=False):
 
         return m_rem, f_fb, state
 
+    def Maltsev25_corecollapse(self, star, engine, conserve_hydrogen_envelope=False):
```
maxbriel, line 2390: @astroJeff @sgossage I'm starting to feel that the `step_SN` file is becoming quite large. I think it's time to split this into separate files.

@ezapartas Nothing for you to do on this comment.

dimsour94 (commented):
I ran a population of 5,000 systems using this branch, including the Maltsev central entropy file with nearest-neighbour and interp_values set to False, and it worked well. The only remaining open issue is the treatment of SNe forming fallback BHs: we currently assume a fallback fraction of 0.99, but we still need to determine what the appropriate fallback fraction should be.

posydon/binary_evol/SN/step_SN.py
```diff
@@ -2343,6 +2386,153 @@ def Patton20_corecollapse(self, star, engine, conserve_hydrogen_envelope=False):
...
+        return m_rem, f_fb, state
+
+    def NS_vs_fallbackBH(self, comp_val, mco_val, M4_val, mu4M4_val):
+        a, b = 1.75, -0.044  # eq. (8) of [8]_
+        # conditions for guaranteed NS formation (eq. 7)
+        if comp_val <= 0.04 or (comp_val < a*mu4M4_val + b and comp_val <= 0.4) or M4_val/mco_val > 0.6:
+            rem = 2
+        else:
+        # stochastic determination of the remnant type (NS versus fallback-BH)
+            rand_number = random.uniform(0,1)
```
maxbriel: Change to rng from stepSN

sgossage:
@ezapartas @dimsour94 is this ready for review?

In principle, things look good outside of the TODO points raised in the PR's description. The current `v2.3` branch also needs to be merged in to this.

maxbriel:
I've updated this PR with the unaddressed comments:

- Merger v2.3 into this branch
- I fixed the setting of `f_fb` and `state` in the case `conserve_hydrogen==True`. 
- I made sure the `k1` and `k2` parameters in `explod_crit` are passed into it from the main function, instead of redefining them. Additionally, there was a typo in the `k2` parameter (0.421 instead of 0.420).
- Changed to self.RNG instead of random.
- Some additional cleanup of comments and indentations.

One problem I am seeing with this PR is the backwards compatibility. It forces the user to always have the new `Kepler_sc_table.dat` file in their `PATH_TO_POSYDON_DATA`, even if they are not using the Matlsev prescription.
**Do we want this?**

Next steps: 
1. Update unit tests
2. Add `Kepler_sc_table.dat` to POSYDON_auxilliary_data.
3. Post-process grids for new SN_models to be included  and retrain the interpolators

sgossage:
Primary next steps are establishing how we want to include the new auxiliary data, doing so, and amending unit tests to account for the new data.

posydon/binary_evol/SN/step_SN.py
```diff
@@ -2341,6 +2388,166 @@ def Patton20_corecollapse(self, star, engine, conserve_hydrogen_envelope=False):
...
+        return m_rem, f_fb, state
+
+    def NS_vs_fallbackBH(self, comp_val, mco_val, M4_val, mu4M4_val):
+        a, b = 1.75, -0.044  # eq. (8) of [8]_
+        # conditions for guaranteed NS formation (eq. 7)
+        if comp_val <= 0.04 or (comp_val < a*mu4M4_val + b and comp_val <= 0.4) or M4_val/mco_val > 0.6:
+            rem = 'NS'
+        else:
+            # stochastic determination of the remnant type (NS versus fallback-BH)
+            rand_number = self.RNG.uniform(0,1)
```
astroJeff, line 2507: This will cause problems for reproducibility. Not sure we can have randomness in our binary evolution. Is this necessary?

posydon/binary_evol/SN/step_SN.py
```diff
@@ -2341,6 +2388,166 @@ def Patton20_corecollapse(self, star, engine, conserve_hydrogen_envelope=False):
...
+            if mu4M4_val > mu4M4_crit2:
+                ff = 0
+            elif mu4M4_val < mu4M4_crit1:
+                ff = 1
+        # final fate classification based on reversed Ertl criterion
+            elif k1 + k2*mu4M4_val - mu4_val > 0:
+                ff = 0
+            else:
+                ff = 1
+        return ff
```
astroJeff, line 2547: We should return a bool instead of an integer here. After all, we are trying to determine if the explosion was successful or not.

astroJeff (approved):

Issue #794, opened 2026-01-29
[BUG]: CO_contact unusual time values reported


bruno-vizzone:
**Describe the bug**
Binaries that reach DCO and trigger the contact event (`CO_contact`) display unusual time values on the binary history (close to zero values). Issue is not present for binaries that do not merge or not enter DCO at all.

**To Reproduce**
Run any population that will result in DCO mergers.

**Expected behavior**
Displaying the time at `CO_contact`.

**POSYDON Version:**
Version 2.2.4

**Additional context**
I believe the problem is on the file `posydon/binary_evol/DT/double_CO.py`. On line 68, when a binary triggers the contact event, `binary.time` is assigned the value `self.res.t[-1]` which corresponds to the time length of the last iteration to solve the DCO evolution IVP. This is the value reported on the binary history at the `CO_contact` event. Instead, it should be assigned the final value of `t0` which is defined in line 84 and updated on each iteration of solving the DCO evolution. The following changes to the file seem to have worked for me:

68  – `            binary.time = self.res.t[-1]`
68  +`            binary.time = self.dco_formation_time_plus_inspiral`

113 +`        self.dco_formation_time_plus_inspiral = t0`

Thanks for looking into it! Love your work :)

astroJeff:
Thank you for pointing this out - I see the problem. We have on line 106:

106 `            t0 += res.t[-1]`

But this doesn't seem to get propagated to the binary object. A different suggestion:

68 -`            binary.time = self.res.t[-1]`
68 +`            binary.time += self.res.t[-1]`

@sgossage can you take a look at this?

sgossage:
I reopened this because the last solution [PR#795](https://github.com/POSYDON-code/POSYDON/pull/795) not fully address this issue.

sgossage:
@bruno-vizzone this issue is now properly solved in the latest [release v2.2.6](https://github.com/POSYDON-code/POSYDON/releases/tag/v2.2.6) of `main`.

PR #795, opened 2026-02-07
[fix] Change solver interval to align with the evolution time

maxbriel:
Related to issue #794.

The time from the solver was not correctly added to the `binary.time`.
The solver here runs from 0 to max_time-binary.time.
This solution adds the final solver time to the binary.time

Alternative solution is to change the solver from binary.time -> max-time. This would be more in-line with the normal detached step.

### DIFF
--- a/posydon/binary_evol/DT/double_CO.py
+++ b/posydon/binary_evol/DT/double_CO.py
@@ -65,7 +65,6 @@ class DoubleCO(detached_step):
         #                         f"{self.res.message}")
         # contact event triggered
         if self.res.status == 1:
-            binary.time += self.res.t[-1]
             binary.eccentricity = 0.0
             binary.state = "contact"
             binary.event = "CO_contact"
@@ -92,7 +91,7 @@ class DoubleCO(detached_step):
                 res = solve_ivp(self.evo,
                                 events=self.evo.ev_contact,
                                 method="BDF",
-                            t_span=(0, self.max_time - t0),
+                            t_span=(t0, self.max_time),
                             y0=[a, e,
                                 secondary.omega0, primary.omega0],
                             rtol=1e-10,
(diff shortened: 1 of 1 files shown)


sgossage:
I think what's going on is actually that the CO step solver uses a bad time interval. Right now it always starts at 0 and spans up to the max time minus the current binary.time. It should start at the current binary.time and go until the max time (similar to how it is done for normal stars). I suggest the following change.

Right now we have:

```
        t0 = binary.time
        ...

        while status == -1 and n < 6:
            try:
                res = solve_ivp(self.evo,
                                events=self.evo.ev_contact,
                                method="BDF",
                            t_span=(0, self.max_time - t0),
                            ...)

            ....
            t0 += res.t[-1]
```

and we should use


```
        t0 = binary.time
        ...

        while status == -1 and n < 6:
            try:
                res = solve_ivp(self.evo,
                                events=self.evo.ev_contact,
                                method="BDF",
                            t_span=(t0, self.max_time),
                            ...)

            ....
            t0 += res.t[-1]
```

posydon/binary_evol/DT/double_CO.py
```diff
@@ -65,7 +65,7 @@ def __call__(self, binary):
         #                         f"{self.res.message}")
         # contact event triggered
         if self.res.status == 1:
-            binary.time = self.res.t[-1]
+            binary.time += self.res.t[-1]
```
sgossage:
```suggestion
```

I think we can get rid of this line (68) because we already update the time and time history in `step_detached` with the method `update_after_evo`.

maxbriel:
That's good. Either we change the interval or how we add the time. I'm happy to go with the solver change.

sgossage (approved):

astroJeff (approved):

PR #798, opened 2026-02-09
[enhancement] Synthetic population lazy imports

maxbriel:
importing from the synthetic population file often took longer than a few seconds, when loading for the first time
i.e. `from posydon.popsyn.synthetic_population import Population`.

This is mostly due to the amount of other parts of the code that are loaded.

This brings the loading down from 1870ms to 737ms
Tested with `python -X importtime -c "from posydon.popsyn.synthetic_population import Population"`

The imports are still not great. Part of this seems to be how much we import from other parts of posydon.
For example, posydon.popsyn.synthetic_population imports `popsyn.binarypopulation`, and `posydon.binary_evol.binarystar`,etc, which is quite a large "time sink". See the attached image for the import time usage.

Focus on the top-left of the below images. These show what is taking the longest to import.

original import time:
importtime

importtime from this rework.
importtime


sgossage:
LGTM except I think @astroJeff had the suggestion to move some of the lighter weight imports like `Pwarn` back to the top.

posydon/popsyn/synthetic_population.py
```diff
@@ -2102,7 +2134,7 @@ def calculate_cosmic_weights(self, SFH_identifier, model_weights, MODEL_in=None)
         #].values
 
         # speed of light
-        c = const.c.to("Mpc/yr").value  # Mpc/yr
+        c = clight * 1.0227121650456949e-17 # cm/s to Mpc/yr
```
astro-abhishek: Putting in that 1.0227121650456949e-17 factor make for a very hardcoded outlook :| 

can we store this as a separate variable at the top instead? like this:

cm_per_sec_to_mpc_per_yr = 1.0227121650456949e-17 
c = clight * cm_per_sec_to_mpc_per_yr

posydon/popsyn/synthetic_population.py
```diff
@@ -2188,6 +2222,9 @@ def plot_efficiency_over_metallicity(self, model_weight_identifier, channels=Fal
         if model_weight_identifier is None:
             raise ValueError("Model weight identifier not provided!")
 
+        import posydon.visualization.plot_pop as plot_pop
+        from posydon.utils.constants import Zsun
```
astro-abhishek: Zsun seems to be used in a lot of places, we could keep this import at the top?

astro-abhishek:
@maxbriel since a lot of imports are now lazy, just checking if you’ve seen any new runtime errors?

sgossage (dismissed):

sgossage:
I've approved, but I think some of the comments above may still need to be addressed regarding a few imports moved back to the top.

posydon/popsyn/synthetic_population.py
astroJeff: We should also change the name so it isn't just `c`, instead something like `clight_Mpc_per_yr`

posydon/popsyn/synthetic_population.py
astroJeff: Agreed. Importing constants shouldn't add much overhead

astroJeff:
This may speed things up, but it obfuscates the code structure. Let's address this with future development in a careful and systematic way

PR #802, opened 2026-02-10
[enhancement] MESA grid running rewrite

maxbriel:
This PR revamps how POSYDON builds the MESA run directories.
The goal of this PR is to create a more manageable and modular infrastructure.
This includes a cleaner POSYDON-MESA-INLISTS repository, and a more obvious configuration file setup.

So far I've kept the output of this rewrite the same as what it was before. This means that additional files and folders are still created, i.e. a binary folder is still present for a single star run. 
I want to focus on specifically reproducing our previous setup, because the `posydon-run-grids` is a separate function with different requirements, and rewriting the logic in `posydon-setup-grids` is already quite the challenge.

At this moment (10/02/2026), I have had to reduce the functionality from reading the posydon inlists. I've hard-coded the read-in of the He-star formation, because this drastically simplified the logic in generating the inlist chain. This means that we cannot arbitrarily add "inlist_stepX" to the "star_formation" folder in the POSYDON repository. **Note: "star_formation" here means pre-main evolution steps; not actual star formation!**
We can reintroduce this if needed.

**I've also been unable to test the dynamic-grid setup.**

Currently, the code reproduces similar grid run setups files as those on yggdrasil. 
There are only a few more tasks to complete:
- [ ] Add submission script
- [ ] Add mk file + run make files


sgossage:
This is a large task and may benefit from being subdivided amongst core-devs.

PR #803, opened 2026-02-10
[docs/enhancement/fix] Natal kick info + fixes a table

maxbriel:
1. Fixes an issue with the "Additional Output" table rendering on the website.
2. Adds the unit + new values to the "Additional Output" table for the natal_kicks.
3. Flips the angles `phi` and `theta` in the list to represent the order the actual list used to be. $\phi$ used to be at index 1, and $\theta$ at 2. See here: https://github.com/POSYDON-code/POSYDON/blob/dd0429ba697800ba2554f084fc7d45ebdd4ff1fe/posydon/binary_evol/SN/step_SN.py#L1496-L1506

### DIFF
--- a/docs/_source/components-overview/pop_syn/single_star.rst
+++ b/docs/_source/components-overview/pop_syn/single_star.rst
@@ -119,7 +119,7 @@ Additional scalar properties are added during the evolution depending on which s
       * mean anomaly (radians)
 
   * - ``natal_kick_velocity``
-    - The natal kick velocity in km/s.
+    - The magnitude of the natal kick velocity in km/s.
   * - ``natal_kick_phi``
     - The natal kick azimuthal angle phi in radians.
   * - ``natal_kick_theta``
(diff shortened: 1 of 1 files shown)


docs/_source/components-overview/pop_syn/single_star.rst
```diff
@@ -103,20 +103,29 @@ The star properties are defined as follows
...
+      | This has been replaced with the individual properties below.
+      | ``natal_kick_array`` contains:
+
+      * velocity (km/s)
+      * azimuthal angle phi (radians)
+      * polar angle theta (radians)
+      * mean anomaly (radians)
+
+  * - ``natal_kick_velocity``
+    - The natal kick velocity in km/s.
```
astro-abhishek: minor suggestion: _magnitude_ of the natal kick velocity in km/s.

astro-abhishek:
I think this would also mean updating the `params.ini` file for future population runs?

maxbriel:
@astro-abhishek I've implemented your suggestion. 
The population_params.ini was updated before, when we made the change to the new properties. I must have forgotten to update it here at the time

astro-abhishek (approved):

PR #806, opened 2026-02-17
[fix] Update OS matrix in install_extras.yml

maxbriel:
Removed macOS 13 from the OS matrix in the workflow, because it's no longer supported.

### DIFF
--- a/.github/workflows/install_extras.yml
+++ b/.github/workflows/install_extras.yml
@@ -13,7 +13,7 @@ jobs:
 
     strategy:
       matrix:
-        os: [macos-13, macos-latest, ubuntu-latest]
+        os: [macos-latest, ubuntu-latest]
         python-version: ['3.11']  # can add more if we want to support
 
     steps:


sgossage (approved):

.github/workflows/install_extras.yml
```diff
@@ -13,7 +13,7 @@ jobs:
 
     strategy:
       matrix:
-        os: [macos-13, macos-latest, ubuntu-latest]
+        os: [macos-latest, ubuntu-latest]
```
astro-abhishek, line 16: Was there a reason macos-13 was explicitly mentioned before? Just want to make sure there are no code or dependencies that only work on that version.

astro-abhishek, line 16: This might be a bit overkill, but dpcumenting the supported ubuntu/mac versions might be helpful later.

maxbriel: I assume this was added at the time because it was the latest macOS version available, but Github has now deprecated it as a runner. So we can no longer use it.
@elizabethteng set up this CI at the time. Do you remember why you added it?

elizabethteng: correct. Safe to remove now.

elizabethteng (approved):

PR #810, opened 2026-02-19
[fix] CO step solver time handling

### LINKED ISSUE #794: [BUG]: CO_contact unusual time values reported

sgossage:
This builds off of [PR#809](https://github.com/POSYDON-code/POSYDON/pull/809) from @maxbriel.

The primary difference is that this PR contains a new `get_time_after_evo` and altered `combined_sol` function to manage combining the compact object step's multiple `solve_ivp` attempts, which occur at increasingly smaller time resolutions.

The DCO solver currently works by making several `solve_ivp` calls at progressively smaller timescales.

```
        ...

        t0 = binary.time
        n = 0
        status = -1
        while status == -1 and n < 6:
            try:
                res = solve_ivp(self.evo,
                                events=self.evo.ev_contact,
                                method="BDF",
                                t_span=(0, self.max_time-t0),
                                ...)
            except Exception as e:
                ...

            
            t0 += res.t[-1]
            n += 1
            ...

```

The time span `(0, self.max_time-t0)` is reduced on each `while` loop pass. Typically the first iteration spans the greatest interval and subsequent iterations span shorter intervals. The time scale varies widely between intervals, for example it can be on the order of `1e9` yr on the first pass and `1e-8` yr on the final pass as it resolves the DCO inspiral and contact event.

Each iteration produces an interpolant (accessed as `res.sol`) from `solve_ivp` that is valid over the time span `(0, res.t[-1])` that it was solved for. The value `res.t[-1]` represents the final time point in the ODE solution. These interpolants are subsequently used to determine the binary separation and orbital period of the DCO, but each will only provide appropriate solutions for times input in the range `(0, res.t[-1])`.

In principle, to get appropriate solutions, we could offset any time input to these interpolants by the appropriate `t0`. Based on the fact that `res.t + t0` represents the actual age of the binary. So e.g., `res.t[-1] + t0 = t_final` and thus `res.t[-1] = t_final - t0`, inputting some other time (`t_input`) less than `t_final` should translate to the correct age needed for the interpolant. However, the difference `t_input - t0` becomes an issue when the difference needs to be on the scale of `1e-8` and input times are on the scale of `1e9`, which is fairly typical. Due to machine precision, this difference can not be resolved on this scale and the difference `t_input - t0` effectively is rounded to 0 always, always providing the wrong time input for the inteprolant.

This PR attempts to remedy this by shifting the input times to the appropriate timescale before using them with the interpolants.

### DIFF
--- a/posydon/binary_evol/DT/double_CO.py
+++ b/posydon/binary_evol/DT/double_CO.py
@@ -50,7 +50,9 @@ class DoubleCO(detached_step):
 
     def __call__(self, binary):
 
+        prior_time_hist = binary.time_history
         super().__call__(binary)
+        time_hist_after_evo = binary.time_history
 
         binary.V_sys = binary.V_sys_history[-1]
 
@@ -62,7 +64,8 @@ class DoubleCO(detached_step):
         # contact event triggered
         if self.res.status == 1:
             binary.time = self.res.t[-1] + self.res.time_sols[-1]
-            binary.time_history[-1] = self.res.t[-1] + self.res.time_sols[-1]
+            if len(time_hist_after_evo) - len(prior_time_hist) > 1:
+                binary.time_history[-1] = self.res.t[-1] + self.res.time_sols[-1]
             binary.eccentricity = 0.0
             binary.state = "contact"
             binary.event = "CO_contact"
--- a/posydon/binary_evol/DT/double_CO.py
+++ b/posydon/binary_evol/DT/double_CO.py
@@ -50,9 +50,9 @@ class DoubleCO(detached_step):
 
     def __call__(self, binary):
 
-        prior_time_hist = binary.time_history
+        prior_tlen = len(binary.time_history)
         super().__call__(binary)
-        time_hist_after_evo = binary.time_history
+        tlen_after_evo = len(binary.time_history)
 
         binary.V_sys = binary.V_sys_history[-1]
 
@@ -64,7 +64,7 @@ class DoubleCO(detached_step):
         # contact event triggered
         if self.res.status == 1:
             binary.time = self.res.t[-1] + self.res.time_sols[-1]
-            if len(time_hist_after_evo) - len(prior_time_hist) > 1:
+            if tlen_after_evo - prior_tlen >= 1:
                 binary.time_history[-1] = self.res.t[-1] + self.res.time_sols[-1]
             binary.eccentricity = 0.0
             binary.state = "contact"
--- a/posydon/binary_evol/DT/double_CO.py
+++ b/posydon/binary_evol/DT/double_CO.py
@@ -132,7 +132,9 @@ class DoubleCO(detached_step):
         else:
             output_solution = CombinedSolution()
             output_solution.t = np.concatenate([t0+t.t for t, t0 in zip(sol, time_sol)])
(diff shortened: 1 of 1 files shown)


sgossage:
Example of the binary

```
STAR1 = SingleStar(**{'mass':33.964274,
                        'natal_kick_array':[0.0, 0.0, 0.0, 0.0]})
STAR2 = SingleStar(**{'mass': 28.98149,
                        'natal_kick_array':[0.0, 0.0, 0.0, 0.0]})

                    **{'time': 0.0, 'state': 'detached', 'event': 'ZAMS', 'orbital_period':82.370989, 'eccentricity': 0.0},
                    properties = sim_pop)
```

with `dt = 2e9`.

image

posydon/binary_evol/DT/double_CO.py
```diff
@@ -8,7 +8,10 @@
 ]
 
 
+from decimal import Decimal
```
astroJeff: It doesn't look like `Decimal` is used. Can you check?

sgossage: thanks @astroJeff , I've removed it now.

sgossage:
Here's an example with `n_o_steps_history = 3`, which now follows the old behavior (from `main` [d39be4d](https://github.com/POSYDON-code/POSYDON/commit/d39be4d8397ad4371cf434c0f4d9ecf5013d26a6)). In this case it will store 3 history steps from each solver attempt:

image

sgossage:
an example with `dt = 1e9` (in the older `main` [d39be4d](https://github.com/POSYDON-code/POSYDON/commit/d39be4d8397ad4371cf434c0f4d9ecf5013d26a6) this feature did not work with `step_dco`).

image

sgossage:
Comparison between this branch (left) and `main`  [d39be4d](https://github.com/POSYDON-code/POSYDON/commit/d39be4d8397ad4371cf434c0f4d9ecf5013d26a6) (right):

image

as noted by @maxbriel in the previous PR, there is a small time difference between the branches. Other than this, binaries in the test suite show identical histories, in these cases on the order of 10 years or less (37.74300003051758 years here).

astroJeff (approved):

maxbriel (approved):

Issue #442, opened 2024-11-21
Clean out duplicated functions and variables


mkruckow:
We have different functions and variables which have duplicates. We should clean that up to always use one version.
Here some examples

- `posydon.utils.gridutils.kepler3_a` vs. `posydon.utils.common_functions.orbital_separation_from_period`
- `posydon.utils.gridutils.T_merger_P` vs.  `posydon.utils.common_functions.inspiral_timescale_from_orbital_period`
- `posydon.utils.gridutils.T_merger_a` vs.  `posydon.utils.common_functions.inspiral_timescale_from_separation`
- variables are often hard coded in functions instead of imported from where they are defined, this happens often in `step_*` vs. `flow_chart`

PR #815, opened 2026-03-04
[fix/feat] Remove duplicate functions in `grid_utils`

maxbriel:
We had some duplicate functions in `grid_utils.py`, which were also present in our `common_functions.py`. 

This PR removes the duplicate functions and replaces them with those in `common_functions`.

Related to issue #442, though I'm not sure if all parts of that issue are addressed with the PR (i.e. the flow chart bulletpoint).


sgossage (approved):

astroJeff:
How often is the `beta_gw` function called? Since it's just a one-line equation, does it really need to be spun off into its own function?

astro-abhishek (approved):

maxbriel:
I don't think it's actually used in the codebase. I've just moved it for this PR, since it was present before, and I didn't want to just remove non-duplicate functions.

PR #818, opened 2026-03-08
[v2.3] Lazy loading `PSyGrids` to save RAM and per step `TrackMatcher` objects

sgossage:
Cherry picked from [PR#778](https://github.com/POSYDON-code/POSYDON/pull/778/changes#top).

**PSyGrid lazy loading and TrackMatcher/GRIDInterpolator recycling:**
- [x] `PSyGrid` data is lazy loaded by default. There's a new `LazyHDF5` class that helps manage this. (see `psygrid.py`)
- [x] `TrackMatcher` objects are created as needed by steps. (Things have been moved/restructured between `step_detached.py` and `simulationproperties.py`.)
- [x] likewise, `GRIDInterpolator` objects are created during step loading and passed to `TrackMatcher` instances as needed. (We only need one of these objects per metallicity, per grid type, so now each step's `TrackMatcher` reuses any existing ones if appropriate. Likewise, things have been moved/restructured between `step_detached.py` and `simulationproperties.py`.)
- [ ] May want a new section in .ini file to take some global `GRIDInterpolator` settings, like grid paths.

RAM usage goes from ~ 6GB to ~3GB as is, with slow accumulation during population synthesis. This exists anyways outside of this PR and at least partially has to do with storing binary evolution data as population synthesis progresses.

Looking into memory accumulation during population synthesis and whether this can be reduced further.

+ `step_mesa` has some heavy calls via `get_final_MESA_step_time`
+ `step_detached` also has some during matching

>[!NOTE]
> **Evolution step default kwarg handling:**
> In order to achieve this in a streamlined way (specifically the per step `TrackMatcher` assignment), the evolution steps have been standardized a bit more than they were. 
> + Each step now contains a `DEFAULT_KWARGS` global property that defines the default settings for that step; something similar was already in `step_SN` and has been adapated to all other step classes. 
> + Now, we can check e.g., `step_class.DEFAULT_KWARGS` to see what is supposed to be defined in that evolution step in a standard way. 
> + For now, this is now used to check whether a step needs a `TrackMatcher` or a `metallicity` assignment. Only `detached_step` and `StepCEE` use `TrackMatcher` objects at the moment.
> + In `SimulationProperties`, the new method `check_step` is run when loading a step and it checks for critical kwargs and makes sure they are defined, or passed on if pre-existing. This is particularly important for kwargs that interface with `BinaryPopulation` and may be passed on from it, like `metallicity`. Now that `SimulationProperties` sets up `TrackMatcher` objects as needed, these are also validated and passed on to steps here.
> + Doc strings have been updated and cleaned accordingly. This is mostly what has changed in other evolution step files.
>
> This standardizes the evolution step design a bit and also means that if someone needs to add additional properties to a class in the future, one just needs to modify this set of default kwargs and they are automatically instantiated in the class's ``__init__`` method and may be more readily checked for/setup via `simulationproperties.py` when loading the steps.


sgossage:
This is a comparison with a single metallicity, 10,000 binary population evolved with lazy loading on (blue) and off (orange), performed with this same branch. The overall RAM usage is about 50% less with lazy loading enabled, as well as slightly faster run times. The current `main` branch (v2.2.6, in green) is shown for comparison.

image

The slower accumulation of RAM during population synthesis is the same regardless.

sgossage:
Running 1 million binaries with a dump rate of 250, distributed over 1000 jobs (i.e., 1000 binaries per task), this is how `main` compares to this branch on Quest:

| Metric            | `main` | `this_branch` | Change              |
|------------------|--------------------|----------------------|---------------------|
| Wall time        | ~20 min            | ~18 min              | `this_branch` faster by ~10%       |
| Avg task runtime | ~12 min            | ~10.5–11 min         | `this_branch` faster by ~10–15%    |
| Peak RAM   | ~7 GB              | ~3.2 GB              | `this_branch` less memory by ~50%              |
| Total CPU time   | ~200 CPU-hours     | ~180 CPU-hours       | `this_branch` less CPU time by ~10%              |

astroJeff (approved):

PR #819, opened 2026-03-09
[fix/enhancement] Remove forced lower mass ratio limit 

maxbriel:
This PR contains PR #797 !

This PR removes the lower mass ratio limit that's forced by `initial_sampling`.
Additionally, it adds a warning when setting up a population synthesis run. if q<0.05 is sampled.


posydon/CLI/popsyn/setup.py
```diff
@@ -86,6 +87,22 @@ def setup_popsyn_function(args):
     validate_ini_file(args.ini_file)
 
     synpop_params = binarypop_kwargs_from_ini(args.ini_file)
+    # warn if mass ratio q = M2/M1 could fall below 0.05
+    if synpop_params['secondary_mass_scheme'] == 'flat_mass_ratio':
+        q_min_possible = (synpop_params['secondary_mass_min']
```
astro-abhishek, line 92: If I understand correctly, the `q_min_possible` should return a theoretically lowest case value, not a minimum from our actual sampling, right?

maxbriel: Yes. Our sampling technically should never be below the theoretical minimum. I wanted to warn the user that their settings are allowing the sampler to return values below the grid values :)

sgossage:
As @maxbriel, pointed out, this PR currently has a failed assertion check in the unit tests that needs to be looked at more closely.

sgossage:
The unit test for `independent_sample.py` fails because the `q_min` previously comes out to ~0.02, which would be forced to the previous minimum of 0.05, leading to the value 9.328252086070083 for `m2[0]`.

Now that the lower limit has been removed, the value of `m2[0]` differs from the hardcoded value. The solution is to update the hardcoded value.

I'll push a fix.

sgossage:
The other failing unit test came from:

```python
        def get_pdf_for_m1(m1):
            m1 = np.atleast_1d(m1)

            # This line here produced the error
            minimum = np.max(
                [kwargs['secondary_mass_min'] / m1, np.ones(len(m1))*0.05],
                axis=0)

            maximum = ...

            ...

            return q_dist
```

in `norm_pop.py`. I made the lower limit 0.0, instead of 0.05 to fix this one. Without that, some of the weights come out to 0, which fails the `np.allclose()` check (line 789) in `test_same_sample_space` from `test_norm_pop.py`.

Should we enforce 0 as a lower limit in `independent_sample.py` also?

```python
    q_min = np.maximum(secondary_mass_min / primary_masses, 0.0)
```

maxbriel:
@sgossage Thank you for looking into the failing tests and making fixes! 

You make a good suggestion about about the lower mass ratio limit of 0.0. The distribution classes have a minimum check as well and will raise an error otherwise. So the question is if we want to force it if the user does something wrong to overwrite it with 0, or just let them figure it out? :) 
I have no preference either way

astroJeff (approved):

sgossage (approved):

sgossage (approved):

PR #822, opened 2026-03-16
[fix] RNG arguments for distribution samplers

sgossage:
This PR contains a bug fix taken from [PR#797](https://github.com/POSYDON-code/POSYDON/pull/797) that addresses the same issue.

We have the option to fix the RNG seed (setting a fixed `entropy`), but any RNG instances with a fixed seed are not passed to several functions, meaning that orbital period and formation times can not be generated with a fixed seed in the intended way. This gives the following functions an `rng` argument to fix that:

+ `posydon.utils.rejection_sampler`
+ `posydon.utils.inverse_sampler`
+ `posydon.utils.histogram_sampler`

In versions including and earlier than v2.2.6, This would affect any attempt to fix the RNG seed when the SFH scenario is not `burst` and when the `orbital_period_scheme` is `'Sana+12_period_extended'` or Moe and DiStefano 17.

>[!Warning]
> This bumps `main` to `v2.2.7`

### DIFF
--- a/posydon/unit_tests/utils/test_common_functions.py
+++ b/posydon/unit_tests/utils/test_common_functions.py
@@ -800,7 +800,7 @@ class TestFunctions:
                                r)
         assert np.allclose(totest.rejection_sampler(x_lim=np.array([0.0,\
                                                                      1.0]),\
-                                                     pdf=mock_pdf, size=5),\
+                                                     pdf=mock_pdf, size=5, rng=mock_rng),\
                            np.array([0.0, 0.25, 0.0, 0.0, 0.0]))
         assert np.allclose(totest.rejection_sampler(x=np.array([1.0, 0.0]),\
                                                        y=np.array([0.2, 0.8]),\
--- a/posydon/unit_tests/utils/test_common_functions.py
+++ b/posydon/unit_tests/utils/test_common_functions.py
@@ -849,6 +849,16 @@ class TestFunctions:
         class MockRNG:
             def uniform(self, low=0.0, high=1.0, size=1):
                 return np.linspace(low, high, num=size)
+            def choice(self, a, size=None, replace=True, p=None):
+                if isinstance(a, int):
+                    a = np.arange(a)
+                sample = []
+                for (v, q) in zip(a, p):
+                    sample += round(size*q) * [v]
+                if len(sample) < size:
+                    sample += (size-len(sample)) * [a[-1]]
+                return np.array(sample[:size])
+
         mock_rng = MockRNG()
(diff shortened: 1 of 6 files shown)


astroJeff (approved):
Everything looks good to me

astroJeff:
I approved, but note, there are a few other places where `np.random` is used: 

1. In the GRB beaming factor calculation (I am not sure if this is still used)
2. In the Bondi-Hoyle accretion rate (when calculating the orbital phase for solving Kepler's Eqn. This is, I think, an incorrect way to approach the problem - we should be calculating a secular average, but we should look into it further)
3. In a bunch of the interpolation and classification training scripts
4. In the Moe & Di Stefano initial distribution

@sgossage this last one, in particular, worries me. Can you look into it? I think it should be a quick change that can be added to this PR

sgossage:
thanks @astroJeff , I've updated the Moe & Di Stefano sampler now with the same fix.

Looking at `bondi_hoyle`, I think doing a secular average would be good (in which case we could potentially avoid `np.random` in these calls), but I'm noticing anyways that we probably want to pass any existing (potentially fixed seed) RNGs to the evolution steps too.

`step_SN` also uses `np.random` functions, which could be fixed this way. It would also give flexibility for any additional random functions in the future. I think we could fix draws from the Maxwellian (and lognorm) distribution to get reproducible kick draws -- potentially replacing the SciPy stats functions with NumPy equivalents. This will take a bit of retooling though.

For the training/interpolation, we could allow the same flexibility to be consistent. Those cases seem like `np.random` is used for cross validation, which probably shouldn't be controlled by `entropy` (?).

The GRB function is not used anywhere in the code, but could be called by a user -- we can add the ability to pass and RNG.

posydon/unit_tests/utils/test_common_functions.py
```diff
@@ -870,20 +884,19 @@ def mock_choice(a, size=None, replace=True, p=None):
             totest.histogram_sampler(x_edges=np.array([0.0, 1.0]),\
                                      y=np.array([0.4, 0.6]))
         # examples:
-        monkeypatch.setattr(np.random, "uniform", mock_uniform)
         monkeypatch.setattr(np.random, "choice", mock_choice)
```
maxbriel: I think this mock can be removed too. It's overwritten in the MockRNG class already.
```suggestion
```

maxbriel (commented):
This looks good to me. Similar to the implementation included in PR #797, except more focused and into `main`.

Only one line that can be removed before accepting.

maxbriel (approved):

PR #823, opened 2026-03-17
[enhancement] Add basic info on `model_weight` units

maxbriel:
This adds some essential information to the `model_weight` docstrings, such as its units.

In the future, I think the model weights require a more extensive documentation page, especially showing which methods are supported and how one can reweight populations with different options.

### DIFF
--- a/posydon/popsyn/norm_pop.py
+++ b/posydon/popsyn/norm_pop.py
@@ -305,7 +305,29 @@ def calculate_model_weights(pop_data,
                             M_sim,
                             simulation_parameters,
                             population_parameters):
-    '''reweight each model in the simulation to the requested population'''
+    """Reweight each model in the simulation to the requested population
+
+    Uses the PDF of the simulation and the PDF of the requested population to calculate
+    the weights for each model in the simulation to match the requested population.
+
+    Parameters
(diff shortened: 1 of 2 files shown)


posydon/popsyn/norm_pop.py
```diff
@@ -305,7 +305,29 @@ def calculate_model_weights(pop_data,
...
-    '''reweight each model in the simulation to the requested population'''
+    """Reweight each model in the simulation to the requested population
+
+    Uses the PDF of the simulation and the PDF of the requested population to calculate
+    the weights for each model in the simulation to match the requested population.
+
+    Parameters
+    ----------
+    pop_data : dict
+        Dictionary containing the population data
```
sgossage: I think we should add details about what the dictionary is comprised of (similar for `simulation_parameters` and `population_parameters` below).

maxbriel: Added some info. I think we should add more extensive documentation on the reweighting, but I haven't been able to get around to doing more than what has been done so far for the POSYDON school.

sgossage (approved):

astroJeff (approved):

PR #831, opened 2026-03-26
[v2.3] New grid paths section for `.ini` file

sgossage:
This adds a section to the `.ini` file that allows users to set the path to MESA grids in a centralized way.

It looks like this in the `.ini` file:
```bash
...

[grid_paths]
  # You shouldn't need to edit these paths unless you want to specify paths to
  # your own custom MESA grids. Leaving these as None will tell POSYDON to use
  # the default paths inside of your $PATH_TO_POSYDON_DATA environment variable.

  HMS_HMS_path = None
  # A string representing the path to your grid HDF5 files for HMS-HMS evolution.
  # If None, the default $PATH_TO_POSYDON_DATA/HMS-HMS/*_Zsun.h5 files will be used
  CO_HMS_RLO_path = None
  # A string representing the path to your grid HDF5 files for CO-HMS evolution.
  # If None, the default $PATH_TO_POSYDON_DATA/CO-HMS_RLO/*_Zsun.h5 files will be used
  CO_HeMS_path = None
  # A string representing the path to your grid HDF5 files for CO-HeMS evolution.
  # If None, the default $PATH_TO_POSYDON_DATA/CO-HeMS/*_Zsun.h5 files will be used
  CO_HeMS_RLO_path = None
  # A string representing the path to your grid HDF5 files for CO-HeMS evolution.
  # If None, the default $PATH_TO_POSYDON_DATA/CO-HeMS_RLO/*_Zsun.h5 files will be used
  single_HMS_path = None
  # A string representing the path to your grid HDF5 files for single star HMS evolution.
  # If None, the default $PATH_TO_POSYDON_DATA/single_HMS/*_Zsun.h5 files will be used. 
  # These are used by detached binaries and single stars in your population.
  single_HeMS_path = None
  # A string representing the path to your grid HDF5 files for single star HeMS evolution.
  # If None, the default $PATH_TO_POSYDON_DATA/single_HeMS/*_Zsun.h5 files will be used.

...
```

These paths are passed to the respective steps. The `single_*` paths are used by any steps that use `TrackMatcher` objects, the rest are used by the MESA grid steps.

### DIFF
--- a/posydon/binary_evol/MESA/step_mesa.py
+++ b/posydon/binary_evol/MESA/step_mesa.py
@@ -196,7 +196,7 @@ class MesaGridStep:
                 and self.interpolation_method != 'nearest_neighbour'):
             raise ValueError('Track interpolation is currently supported only '
                              'by the nearest neighbour interpolation method!')
-        
+
         z_str = convert_metallicity_to_string(self.metallicity)
         self.grid_name = os.path.join(self.grid_path, f"{z_str}_Zsun.h5")
 
@@ -1358,7 +1358,7 @@ class MS_MS_step(MesaGridStep):
                              'binary.event = %s and not H-rich_Core_H_burning '
                              '- H-rich_Core_H_burning - * - ZAMS'
                              % (state_1, state_2, event))
-        
+
     def __repr__(self):
         """Return the name of evolution step and settings."""
         return "MS_MS_step:\n" + \
@@ -1473,7 +1473,7 @@ class CO_HMS_RLO_step(MesaGridStep):
             self.binary.state = "detached"
             self.binary.event = "redirect_from_CO_HMS_RLO"
             return
-        
+
     def __repr__(self):
         """Return the name of evolution step and settings."""
         return "CO_HMS_RLO_step:\n" + \
@@ -1588,7 +1588,7 @@ class CO_HeMS_RLO_step(MesaGridStep):
             self.binary.state = "detached"
             self.binary.event = "redirect_from_CO_HeMS_RLO"
             return
-        
+
     def __repr__(self):
         """Return the name of evolution step and settings."""
         return "CO_HeMS_RLO_step:\n" + \
@@ -1684,7 +1684,7 @@ class CO_HeMS_step(MesaGridStep):
             self.binary.state = 'detached'
             self.binary.event = 'redirect_from_CO_HeMS'
             return
-        
+
     def __repr__(self):
         """Return the name of evolution step and settings."""
         return "CO_HeMS_step:\n" + \
@@ -1834,7 +1834,7 @@ class HMS_HMS_RLO_step(MesaGridStep):
             self.binary.state = "detached"
             self.binary.event = "redirect_from_HMS_HMS_RLO"
             return
-        
+
     def __repr__(self):
         """Return the name of evolution step and settings."""
         return "HMS_HMS_RLO_step:\n" + \
--- a/posydon/binary_evol/MESA/step_mesa.py
+++ b/posydon/binary_evol/MESA/step_mesa.py
@@ -127,8 +127,7 @@ class MesaGridStep:
     """Superclass for steps using the POSYDON grids."""
 
     DEFAULT_KWARGS = {'metallicity': None,
-                      'grid_name': None,
-                      'path': PATH_TO_POSYDON_DATA,
+                      'grid_path': None,
                       'interpolation_path': None,
                       'interpolation_filename': None,
                       'interpolation_method': 'nearest_neighbour',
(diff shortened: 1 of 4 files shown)


posydon/binary_evol/MESA/step_mesa.py
```diff
@@ -198,11 +197,14 @@ def __init__(self, **kwargs):
             raise ValueError('Track interpolation is currently supported only '
                              'by the nearest neighbour interpolation method!')
 
+        z_str = convert_metallicity_to_string(self.metallicity)
+        self.grid_name = os.path.join(self.grid_path, f"{z_str}_Zsun.h5")
```
maxbriel, line 201: I would prefer if we can not make the metallicity string hard-coded?

I feel like it should be possible to just link a specific file/grid_file without enforcing the naming convention, although I can be persuaded otherwise.

sgossage: I think this is a good idea too.

It is difficult with our current infrastructure though. We need to know the metallicity of the grid file and right now we get that info from the file name. If we had the metallicity stored in the HDF5 metadata, we could just read that, but we don't at the moment (as far as I can see), complicating things. I think this would be a good future goal: to store metallicity information in the HDF5 file when it is created. I think that would be the cleanest solution. In principle, this could also allow us to more easily define an arbitrary set of metallicities in the .ini to use instead of our pre-defined 8 as well.

An option with how things are now could be to give users the ability to provide their own mapping in the .ini file. That might look like a list of file names that a user can set, e.g., one for each of our 8 metallicities. Or, a dictionary mapping a file name to a metallicity. It gets a bit clunky I think, but I'll try to see if something works, or feel free to let me know if you have something in mind.

astro-abhishek: Just a comment: since we don't have a `grid_name` any more, we have to make sure that the grid type is included in the `grid_path` (e.g., HMS-HMS/).

posydon/binary_evol/MESA/step_mesa.py
```diff
@@ -213,14 +215,13 @@ def __init__(self, **kwargs):
 
             # Set the interpolation path
             if self.interpolation_path is None:
-                self.interpolation_path = os.path.join(self.path,
-                    os.path.split(self.grid_name)[0],
+                self.interpolation_path = os.path.join(self.grid_path,
```
astro-abhishek, line 218: I get the idea, but now the `interpolation_path` depends on the `grid_path` which might mean being more careful in case of user-specific grids and their interpolators.

posydon/binary_evol/simulationproperties.py
```diff
@@ -453,9 +522,18 @@ def check_step(self, metallicity, RNG, step_name, step_tup, verbose=False):
             if metallicity is None:
                 Pwarn(f"{step_name} not assigned a metallicity. "
                     "Defaulting to Z = Zsun (solar).",
-                    "MissingValueWarning")
+                    "ReplaceValueWarning")
                 metallicity = 1.0
             step_kwargs['metallicity'] = float(metallicity)
+            print(step_name, step_kwargs['metallicity'])
```
astro-abhishek: I would suggest to keep this print statement within a verbose check?

posydon/binary_evol/MESA/step_mesa.py
sgossage: yeah, as things are right now a user will need to define the full path to their custom grid in the .ini using one of the designated variables:

```
  HMS_HMS_path = None
  # A string representing the path to your grid HDF5 files for HMS-HMS evolution.
  # If None, the default $PATH_TO_POSYDON_DATA/HMS-HMS/*_Zsun.h5 files will be used
  CO_HMS_RLO_path = None
  CO_HeMS_path = None
  CO_HeMS_RLO_path = None
  single_HMS_path = None
  single_HeMS_path = None
```

Or else, by default, the `grid_path` includes the grid type, e.g., for HMS-HMS the default is:

```python
...
"HMS_HMS_path": os.path.join(PATH_TO_POSYDON_DATA, "HMS-HMS")
...
```

in `SimulationProperties`.

posydon/binary_evol/MESA/step_mesa.py
sgossage: Yeah, right now the user's interpolation files are expected to follow the same convention as POSYDON, where we have

```
path/to/my/grid/
|__1e+00_Zsun.h5
|__interpolators/
    |__1e+00_Zsun.pkl
```

The `path/to/my/grid/` can be custom now, but it expects an `interpolators/` subdirectory to live there in order to find the interpolators. We can try to make this more flexible -- that might involve a separate user defined interpolator path, or else we can try to emphasize this and maybe create some warnings to guide users to follow this convention.

posydon/binary_evol/simulationproperties.py
sgossage: Thanks, I got rid of that print statement for now. The other `verbose` guarded print statements will show the metallicity. That print statement must have been leftover from debugging.

astroJeff (approved):

astro-abhishek (approved):

PR #834, opened 2026-04-15
[feat] add `max_concurrent_jobs` and `exclude` to popsyn CLI

maxbriel:
Adds two additional options to the population synthesis CLI interface.
I originally implemented these on the #mb_model_paper branch, but had forgotten to add them to main version.

`--max_concurrent_jobs` allows the user to set a maximum number of jobs to run concurrently. This is per job array.

`--exclude` allows the user to exclude specific nodes/cpus from the population synthesis run. This is for all job arrays created. This was needed because there were some "bad" nodes on yggdrasil for a while.

### DIFF
--- a/bin/posydon-popsyn
+++ b/bin/posydon-popsyn
@@ -71,6 +71,16 @@ if __name__ == '__main__':
         '--account',
         help='the account you would like to use',
         default=None)
+    setup_parser.add_argument(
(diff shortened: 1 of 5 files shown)


sgossage:
as discussed at the dev meeting 4/23/26, this needs testing

astroJeff (approved):

PR #837, opened 2026-04-16
[enhancement] Add power law mass ratio options for reweighting

maxbriel:
Allows the user to request a mass ratio distribution following a (requested) power law slope.
Usage example:

```
params = {
    'secondary_mass_scheme': 'power_law_mass_ratio',
    'mass_ratio_slope': -1.0,  # alpha in q^alpha
    'q_min': 0.0,
    'q_max': 1.0,
    ...
}
```

### DIFF
--- a/posydon/popsyn/distributions.py
+++ b/posydon/popsyn/distributions.py
@@ -669,7 +669,7 @@ class PowerLawPeriod():
 
         return pdf_values
 
-    def rvs(self, size=1, rng=None):
+    def rvs(self, size=1, n_points=1000, rng=None):
         """Draw random samples from the power law period distribution.
 
         Parameters
@@ -691,7 +691,6 @@ class PowerLawPeriod():
         from posydon.utils.common_functions import inverse_sampler
 
         # Create discretized PDF for inverse sampling
-        n_points = 1000
         logp_grid = np.linspace(np.log10(self.p_min), np.log10(self.p_max), n_points)
         pdf_values = self.power_law_period(logp_grid)
 
--- a/posydon/popsyn/distributions.py
+++ b/posydon/popsyn/distributions.py
@@ -270,7 +270,7 @@ class PowerLawMassRatio:
         pdf_values[valid] = self.power_law_mass_ratio(q[valid]) * self.norm
         return pdf_values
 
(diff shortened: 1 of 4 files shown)


sgossage (dismissed):

posydon/popsyn/distributions.py
```diff
@@ -151,6 +151,151 @@ def rvs(self, size=1, rng=None):
...
+        -------
+        ndarray
+            Random samples from the distribution.
+        """
+        if rng is None:
+            rng = np.random.default_rng()
+
+        from posydon.utils.common_functions import inverse_sampler
+
+        n_points = 1000
```
sgossage: Should we make `n_points` an argument vs. hard coded?

posydon/popsyn/distributions.py
```diff
@@ -151,6 +151,151 @@ def rvs(self, size=1, rng=None):
...
+    ValueError
+        If q_min or q_max are not in valid range, or if q_min >= q_max.
+
+    Examples
+    --------
+    >>> dist = PowerLawMassRatio(alpha=-1.0, q_min=0.1, q_max=1.0)
+    >>> pdf_value = dist.pdf(0.5)
+    """
+
+    def __init__(self, alpha=0.0, q_min=0.05, q_max=1.0):
```
sgossage, line 180: Not strictly necessary for this PR, but it looks like this class has some overlap with `FlatMassRatio` and we could create a parent class that this class and that class are children of, modifying just the relevant parts.

posydon/popsyn/distributions.py
maxbriel: I've made it into an argument, although it's more for setting up the pdf values for doing the inverse sampling :)

posydon/popsyn/distributions.py
maxbriel: I agree. We can rework the FlatMassRatio to inherit from the PowerLawMassRatio. I suggest we do not do that in this PR

maxbriel:
@sgossage I've implemented your suggestion about the `n_points`.  For now we can keep the `FlatMassRatio` class and do a refactor at a later date

astroJeff (approved):

PR #844, opened 2026-05-07
[fix] suppress warnings for overflow & divide by zero from unbound binaries

maxbriel:
Wraps calculations into `np.errstate` to suppress `overflow` and `divide by zero` warning that come from unbound systems due to high kick velocities.


astroJeff:
I really don't like this solution. We should be checking for whether the binary is disrupted before we calculate all the post-SN orbital parameters. However, I looked through `step_SN.py` again, and, unfortunately, doing the right thing would require a fairly substantial re-write of the module. This fix ought to work for now

astroJeff (approved):

astroJeff (approved):

PR #845, opened 2026-05-09
[v2.3] Postprocessing LazyHDF5 support

sgossage:
The postprocessing expects an `ndarray` when adding fields to a `PSyGrid`, such as when postprocessing our data. This adds support so that it is compatible with `LazyHDF5` objects, which is what things like `psygrid.final_values` exist as now for RAM efficiency.


posydon/utils/gridutils.py
```diff
@@ -140,11 +142,15 @@ def add_field(a, descr):
     True
 
     """
-    if a.dtype.fields is None:
-        raise ValueError("'a' must be a structured numpy array")
+    if a.dtype.fields is None or len(a.dtype.fields) == 0:
+        raise ValueError("'a.dtype.fields' must not be empty or None.")
     b = np.empty(a.shape, dtype=a.dtype.descr + descr)
     for name in a.dtype.names:
         b[name] = a[name]
```
maxbriel, line 149: Is this loop still needed if we overwrite the whole thing later?

sgossage: I think so, because this loop handles copying over all of the old data from `a` to the new object `b`.

At least where it is used in `psygrid.py`, and as far as I could see, if we don't have this then we'd wind up with the new objects data columns being empty.

I might be overlooking a place where we do a full overwrite though.

maxbriel (approved):

posydon/utils/gridutils.py
astroJeff: Agreed. Resolving...

astroJeff (approved):

PR #846, opened 2026-05-15
Installation script

maxbriel:
A script to simplify the installation of POSYDON, setup the environment, and download the data.

The idea is that the user only has to run the command below, which downloads and starts the script. The script guides the user through the POSYDON installation.

```
bash <(curl -sSL https://raw.githubusercontent.com/POSYDON-code/POSYDON/main/install.sh)
```

> [!IMPORTANT]
> Requires `conda` to already be installed! I can adapt the code to also check the operating system + install `conda` if not install. **Is that's something we want?**

Co-written with Claude Sonnet 4.6

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -539,10 +539,6 @@ show_final_instructions() {
     else
         echo -e "2. ${BOLD}Start using POSYDON:${NC}"
     fi
-    echo -e "   ${CYAN}python${NC}"
-    echo -e "   ${CYAN}>>> import posydon${NC}"
-    echo -e "   ${CYAN}>>> print(posydon.__version__)${NC}"
-    echo ""
 
     echo -e "${BOLD}Environment Variables:${NC}"
     echo "  The following environment variables are set in your conda environment:"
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -0,0 +1,592 @@
...
+        echo -n "Data directory path: "
+        read -r DATA_PATH
+
+        # Expand tilde to home directory
+        DATA_PATH="${DATA_PATH/#\~/$HOME}"
+
+        if [[ -z "$DATA_PATH" ]]; then
+            print_error "Path cannot be empty!"
+            continue
+        fi
```
sgossage, line 224: I think it could be good to check for pre-existing datasets and/or if $PATH_TO_POSYDON_DATA already exists and give an option to use that.

sgossage:
> Requires conda to already be installed! I can adapt the code to also check the operating system + install conda if not install. Is that's something we want?

It looks like we have a warning that guides users on what to do if `conda` is not detected, which I think is fine. If we implement an automated `conda` install, things could become complicated with OS dependencies and future-proofing, which I'm not against, but I think it is reasonable to exclude for now.

install.sh
maxbriel: Great idea! I've implemented a check if PATH_TO_POSYDON_DATA already exists and to ask if the user wants to use that for installation.
If accepted, it uses the PATH_TO_POSYDON_DATA that exists.
If rejected, it requests a path like before.

sgossage:
One potential issue I'm seeing is that if you have a previously installed editable version of POSYDON via `pip`, that install location gets picked up instead of the `conda` install location targeted by this script.

A solution is to avoid `pip` installation altogether and not mix packages, but maybe good to safeguard against this anyways. E.g., there could be a check like

```bash
IMPORTED=$(conda run -n "$ENV_NAME" python -c "import posydon; print(posydon.__file__)")

if [[ "$IMPORTED" != "$ENV_PREFIX"* ]]; then
    print_warning "A different POSYDON installation is shadowing the conda package:"
    echo "  Imported from: $IMPORTED"
fi
```

and maybe instead of checking for the `PATH_TO_POSYDON` from `posydon.__file__`, we do something like

```bash
>> ENV_PREFIX=$(conda run -n "$ENV_NAME" python -c "import sys; print(sys.prefix)")
>> PATH_TO_POSYDON=$(dirname "$(dirname "$(find "$ENV_PREFIX" -path "*/site-packages/posydon/__init__.py" | head -1)")")
```

sgossage:
> One potential issue I'm seeing is that if you have a previously installed editable version of POSYDON via `pip`, that install location gets picked up instead of the `conda` install location targeted by this script.
> 
> A solution is to avoid `pip` installation altogether and not mix packages, but maybe good to safeguard against this anyways. E.g., there could be a check like
> 
> ```shell
> IMPORTED=$(conda run -n "$ENV_NAME" python -c "import posydon; print(posydon.__file__)")
> 
> if [[ "$IMPORTED" != "$ENV_PREFIX"* ]]; then
>     print_warning "A different POSYDON installation is shadowing the conda package:"
>     echo "  Imported from: $IMPORTED"
> fi
> ```
> 
> and maybe instead of checking for the `PATH_TO_POSYDON` from `posydon.__file__`, we do something like
> 
> ```shell
> >> ENV_PREFIX=$(conda run -n "$ENV_NAME" python -c "import sys; print(sys.prefix)")
> >> PATH_TO_POSYDON=$(dirname "$(dirname "$(find "$ENV_PREFIX" -path "*/site-packages/posydon/__init__.py" | head -1)")")
> ```

Just as a side note, this warning will trigger if `install.sh` is run within the `PATH_TO_POSYDON` directory because Python will prepend the current working directory to the system path, and so doing 

```python
import posydon
```

winds up importing that version of POSYDON and not `*/site-packages/posydon` (where `conda` actually installs the package). The install goes fine either way and this warning outputs correctly.

sgossage (approved):

astroJeff (approved):

PR #848, opened 2026-05-18
[fix/enhancement] move SN_models out of `final_values`

maxbriel:
We've likely hit the maximum number of columns in a single HDF5 dataset for the `final_values` with the increased number of SN models.
As such, we need to separate out the SN_models to make sure we will not hit the same issue in the future. 

I've rewritten the `add_post_processed_quantities` to:
1. Use the faster `recfuntions` from `numpy` to add all columns in one go to the structure array, `final_values`. Originally, we were writing one column at the time. deleting the existing `final_values` and then writing the new `final_values` with 1 extra column. Repeating this process for each new column we wanted to add (several hundred times given the number of SN_MODELS we have). This is a massive I/O overhead.
2. Store the `SN_MODELS` in their own group (`grid/SN_MODELS`) in the HDF5 file. Each SN_MODEL has its own dataset + in the attrs the parameters for the model are stored.
3. `grid.compression_args` are not actually populated correctly when loading a grid with `grid(path_to_file)` as such when writing the "new" datasets to file, they were uncompressed. 

> [!CAUTION]
> Bullet point 3 means that **the `final_values` dataset in the release dataset are uncompressed!**. I verified that this is the case!

<details>
<summary>Screenshots of uncompressed final_values</summary>

`final_values`
image

`initial_values`
image
</details>

This PR now also changes the `IF_interpolator` and `interpolator`. If required, the `IF_interpolator` now uses `_get_grid_column` to get the `final_values` or `SN_models`. It only does this if it looped over the SN_MODELS before. 

Changes to `step_MESA` is just how we access the `SN_MODEL` data. It should be checked if this stored everything exactly the same as before. `step_SN` is able to access this correctly at the moment.

> [!WARNING]
> There is a bug I noticed with the `LazyHDF5` accessing the HDF5 file when adding a column. When adding something to the HDF5f file, we reload the file, e.g. `grid._reload_hdf5_file()`. This invalidates the `LazyHDF5` linking. We should fix this, if possible, but not in this PR.

### DIFF
--- a/bin/posydon-run-pipeline
+++ b/bin/posydon-run-pipeline
@@ -282,7 +282,7 @@ def calculate_extra_values(i, path_to_csv_file, verbose=False):
         star_2_CO = False
 
     # load grid with ORIGINAL compression and get extra colums
-    grid_ORIGINAL = PSyGrid(grid_ORIGINAL_path)
+    grid_ORIGINAL = PSyGrid(grid_ORIGINAL_path, lazy=False)
     MESA_dirs_EXTRA_COLUMNS, EXTRA_COLUMNS = post_process_grid(grid_ORIGINAL,
                                               index=None, star_2_CO=star_2_CO,
                                               verbose=verbose)
@@ -296,7 +296,7 @@ def calculate_extra_values(i, path_to_csv_file, verbose=False):
     shutil.copyfile(grid_path, processed_grid_path)
 
     # load processed grid and append values
-    grid = PSyGrid(processed_grid_path)
+    grid = PSyGrid(processed_grid_path, lazy=False)
     add_post_processed_quantities(grid, MESA_dirs_EXTRA_COLUMNS, EXTRA_COLUMNS)
     if verbose:
         print(f'Added processed quantities to grid: {processed_grid_path}')
@@ -357,7 +357,7 @@ def train_interpolators(i, path_to_csv_file, verbose=False):
 
     # load grid
     grid = PSyGrid(verbose=verbose)
-    grid.load(grid_path)
+    grid.load(grid_path, lazy=False)
 
     interp_method = [method, method]
     interp_classes = ["stable_MT", "unstable_MT"]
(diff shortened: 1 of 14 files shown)


maxbriel:
Here are the output files from the binaries that I ran with the fix and v2.3. I originally posted these on Slack and forgot to add them here:

[evolve_1Zsun_fix.log](https://github.com/user-attachments/files/28843199/evolve_1Zsun_fix.log)
[evolve_1Zsun_main_.log](https://github.com/user-attachments/files/28843202/evolve_1Zsun_main_.log)

bin/posydon-run-pipeline
```diff
@@ -409,24 +409,31 @@ def train_interpolators(i, path_to_csv_file, verbose=False):
     # to catch reverse mass transfer cases where the secondary undergoes
     # core collapse first
     for SN_MODEL_NAME in SN_MODELS.keys():
+        if SN_MODEL_NAME not in grid.SN_MODELS:
+            continue
+        sn_ds = grid.SN_MODELS[SN_MODEL_NAME]
         for i in range(1,2): #TODO: range(1,3):
```
sgossage, line 415: why not `range(1, 3)`?

bin/posydon-run-pipeline
```diff
@@ -409,24 +409,31 @@ def train_interpolators(i, path_to_csv_file, verbose=False):
     # to catch reverse mass transfer cases where the secondary undergoes
     # core collapse first
```
sgossage, line 410: We should keep this in mind as a to do item. (Not necessarily for this PR though.)

posydon/grids/psygrid.py
```diff
@@ -1495,9 +1500,9 @@ def load(self, filepath=None, lazy=True):
             new_dtype[dtype[0]] = dtype[1]
 
         if lazy:
-            initial_values = LazyHDF5(initial_values)
-            final_values = LazyHDF5(final_values, new_dtype)
-        else: # pragma: no cover
+                initial_values = LazyHDF5(initial_values)
+                final_values = LazyHDF5(final_values, dtype_set=new_dtype)
```
sgossage:
```suggestion
            initial_values = LazyHDF5(initial_values)
            final_values = LazyHDF5(final_values, dtype_set=new_dtype)
```

sgossage:
Overall this looks good and just needs tests updated.

sgossage (approved):

astroJeff (approved):

PR #855, opened 2026-06-23
[enhancement] save the reweighting population parameters to file

maxbriel:
This PR saves the population parameters used for reweighting to the hdf5 file.

Additionally this adds a function to load these parameters based on the `model_weight_identifier` stored in the file.


sgossage (approved):

sgossage:
The only other things is that perhaps we should update the docs alongside this.

We have some parts of the documentation that mentions reweighting, e.g.,

https://posydon.org/POSYDON/latest/tutorials-examples/population-synthesis/bbh_analysis.html

and

https://posydon.org/POSYDON/latest/components-overview/pop_syn/synthetic_population.html

where we might add some information about this.

PR #861, opened 2026-06-27
fix: disable LaTeX usetex in default mplstyle to avoid crash on systems without LaTeX

### LINKED ISSUE #619: [BUG]: LaTeX not installed error
**Describe the bug**
When doing plotting using the default posydon style sheet, the plotting can fail due to the style sheet using LaTeX. We do not have this as a requirement. If it is not installed, plotting will fail!
Specific Error:

```
RuntimeError: Failed to process string with tex because latex could not be found
```

**To Reproduce**
LaTeX should not be "installed" in the PATH. Install posydon in a new environment and try to plot anything with the default style sheet loaded. 

**Expected behavior**
Either fallback onto non-latex plotting or make LaTeX part of the installation instructions.

**POSYDON Version:**
Latest development version

**System Configuration (please complete the following information):**
- Operating System [e.g., OSX, Ubuntu, etc.]
- HPC scheduler (if applicable) [e.g., SLURM]

**Additional context**

molloyzak13:
POSYDON's default matplotlib style sheet sets `text.usetex: True`, which crashes plotting on any system that doesn't have LaTeX installed. The error is `RuntimeError: Failed to process string with tex because latex could not be found`.

The root cause is a one-liner in `posydon/visualization/posydon.mplstyle` — usetex is on by default but LaTeX was never listed as a dependency. If a user runs POSYDON in a clean environment and tries to plot with the default style, it breaks.

Set `text.usetex` to `False`. Matplotlib's built-in mathtext renderer handles the same LaTeX-like math syntax without needing an external texlive install, so plots look the same. Users who have LaTeX and want it can still flip it back in their own matplotlibrc.

Closes #619.

### DIFF
--- a/posydon/visualization/plot_pop.py
+++ b/posydon/visualization/plot_pop.py
@@ -5,7 +5,6 @@ __authors__ = [
 
 import os
 import shutil
-import warnings
 
 import matplotlib as mpl
 import matplotlib.pyplot as plt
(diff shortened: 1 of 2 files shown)


astroJeff (commented):
I have some concerns that this will cause some of our plotting scripts to break, as they are designed to require LaTex formatting of strings (e.g., axis labels). We ought to check our plotting scripts before adopting this change.

molloyzak13:
reply_target: POSYDON-code/POSYDON#861
Pushed an update that fixes the CI failures. On the plotting scripts concern: the fallback path only activates when LaTeX isn't available on the system, so the default rendering with LaTeX formatting should be unchanged if you have it installed. Happy to check specific plotting scripts if you can point me to which ones to look at.

molloyzak13:
reply_target: POSYDON-code/POSYDON#861
Pushed an update that fixes the CI failures. On the plotting scripts concern: the fallback path only activates when LaTeX isn't available on the system, so the default rendering with LaTeX formatting should be unchanged if you have it installed. Happy to check specific plotting scripts if you can point me to which ones to look at.reply_status: sent
submitted_url: https://github.com/POSYDON-code/POSYDON/pull/861#issuecomment-4941218024
reply_verdicts: mentor=VERDICT: APPROVE mia=VERDICT: APPROVE sam=VERDICT: APPROVE

maxbriel (dismissed):

sgossage (approved):

PR #862, opened 2026-07-04
fix: flatten time array in TransientPopulation.plot_delay_time_distribution (closes #856)

### LINKED ISSUE #856: [BUG]: TransientPopulation.plot_delay_time_distribution() error
Note: the bug report below was prepared by AI under my supervision

**Describe the bug**

`TransientPopulation.plot_delay_time_distribution()` raises a `ValueError` when called with model weights, because the delay-time array and weight array passed to `np.histogram()` have different shapes.

The issue appears to be that the method obtains the delay times using

```python
time = self.select(columns=["time"]).values
```

which returns a 2D array with shape `(N, 1)`, while the model weights are flattened:

```python
model_weights = self.model_weights(model_weights_identifier).to_numpy().flatten()
```

which gives shape `(N,)`.

Then the method calls

```python
h, bin_edges = np.histogram(time, bins=bins, weights=model_weights)
```

and NumPy raises

```text
ValueError: weights should have the same shape as a.
```

because `time.shape != model_weights.shape`, even though the number of entries is the same.

The same issue likely affects the metallicity-specific branch, where `time` is also extracted as a one-column DataFrame and converted to `.values`, giving shape `(N_met, 1)`, while `weights_for_met` is 1D.

**To Reproduce**

```python
import numpy as np

bins = np.logspace(6.5, 10.5)

BBH_mergers.plot_delay_time_distribution("default", bins=bins)
```

where `BBH_mergers` is a `TransientPopulation` object and `"default"` is an existing model weights identifier.

A minimal diagnostic showing the source of the mismatch is:

```python
time = BBH_mergers.select(columns=["time"]).values
weights = BBH_mergers.model_weights("default").to_numpy().flatten()

print(time.shape)
print(weights.shape)
print(time.size, weights.size)
```

This gives shapes like

```python
(N, 1)
(N,)
N N
```

so the arrays contain the same number of elements but have incompatible shapes for `np.histogram(..., weights=...)`.

**Expected behavior**

`plot_delay_time_distribution()` should plot the delay-time distribution without error when the number of delay times and the number of model weights are the same.

For example, the method should pass 1D arrays of matching shape to `np.histogram()`.

**Actual behavior**

The method raises:

```text
ValueError: weights should have the same shape as a.
```

with traceback ending at:

```python
h, bin_edges = np.histogram(time, bins=bins, weights=model_weights)
```

**Suggested fix**

Flatten the delay-time array before passing it to `np.histogram()`.

For the full-population branch, replace

```python
time = self.select(columns=["time"]).values
time = time * 1e6  # yr
```

with something like

```python
time = self.select(columns=["time"])["time"].to_numpy()
time = time * 1e6  # yr
```

or

```python
time = self.select(columns=["time"]).to_numpy().flatten()
time = time * 1e6  # yr
```

For the metallicity-specific branch, similarly replace

```python
time = time_df[time_df["metallicity"] == metallicity].drop(columns=["metallicity"]).values
time = time * 1e6  # yr
```

with something like

```python
mask = np.isclose(time_df["metallicity"].to_numpy(), metallicity)
time = time_df.loc[mask, "time"].to_numpy()
time = time * 1e6  # yr
```

It may also be safer to use `np.isclose` for metallicity comparisons rather than exact equality, since metallicities may be floating-point values.

A slightly more robust version would align the model weights with the transient table by index before applying the metallicity mask, e.g.

```python
time_df = self.select(columns=["metallicity", "time"])
weights_df = self.model_weights(model_weights_identifier)
time_df = time_df.join(weights_df)

mask = np.isclose(time_df["metallicity"].to_numpy(), metallicity)

time = time_df.loc[mask, "time"].to_numpy() * 1e6
weights_for_met = time_df.loc[mask, model_weights_identifier].to_numpy()
```

This avoids possible shape or ordering mismatches if the model weights table and transient table are not perfectly position-aligned.

**Environment**

* POSYDON version: 2.2.8 installed in a conda environment
* Python version: 3.11
* NumPy version: 1.26.4

**Additional context**

This can be worked around outside the method by manually flattening the delay-time array:

```python
import numpy as np
import matplotlib.pyplot as plt

bins = np.logspace(6.5, 10.5)

time = BBH_mergers.select(columns=["time"])["time"].to_numpy() * 1e6
weights = BBH_mergers.model_weights("default").to_numpy().flatten()

h, bin_edges = np.histogram(time, bins=bins, weights=weights)
h = h / np.diff(bin_edges)

fig, ax = plt.subplots()
ax.step(bin_edges[:-1], h, where="post", color="black")
ax.set_xscale("log")
ax.set_yscale("log")
ax.set_xlabel("Time [yr]")
ax.set_ylabel("Number of events/Msun/yr")
```

botbikamordehai2-sketch:
## What
`TransientPopulation.plot_delay_time_distribution()` raised a ValueError when calling `np.histogram()` with model weights because the `time` array obtained via `.select().values` was 2D (shape (N,1)) while `model_weights` was 1D (shape (N,)).

## Fix
Changed the line:
```python
time = self.select(columns=["time"]).values
```
to:
```python
time = self.select(columns=["time"]).to_numpy().flatten()
```
This ensures the `time` array is 1D, matching the shape of the weights.

Closes #856

### DIFF
--- a/posydon/popsyn/synthetic_population.py
+++ b/posydon/popsyn/synthetic_population.py
@@ -2244,7 +2244,7 @@ class TransientPopulation(Population):
         model_weights = self.model_weights(model_weights_identifier).to_numpy().flatten()
 
         if metallicity is None:
-            time = self.select(columns=["time"]).values
+            time = self.select(columns=["time"]).to_numpy().flatten()
             time = time * 1e6  # yr
 
             # Create histogram with model weights


maxbriel:
Already implemented in PR #858

Issue #829, opened 2026-03-26
Incomplete calculation of the Bondi-Hoyle accretion rate


astroJeff:
In common_functions.py we calculate the Bondi-Hoyle accretion rate, but this calculation is imperfect. In calculating it, we draw a random orbital phase. This is not the correct approach. Instead, we should be performing an orbit average - this has an analytic form. Note, it assumes some approximations (orbital periods are short).

Relevant code is in `bondi_hoyle` function in common_functions.py. 

Specific code starts around lines 631

tassos25:
PR #881 adds a new property-based `posydon.utils.xrb.bondi_hoyle_accretion_rate` function that uses the standard deterministic analytic orbital factor rather than drawing a single random phase. This provides a possible implementation path for the concern raised here, but #881 will **not** change `common_functions.bondi_hoyle`.

The appropriate replacement is still a physics question that deserves separate review: the analytic expression assumes constant wind/stellar properties over an orbit and uses a characteristic orbital velocity, whereas a direct phase average can retain the phase dependence of separation, orbital speed, wind speed, and the wind--orbit angle. We should agree which quantity POSYDON intends to model before replacing the established function.

I opened #888 for the broader legacy implementation audit and #887 for eventual consolidation of duplicated wind/BHL code. A later dedicated PR can resolve this issue with focused numerical comparisons and compatibility tests.

PR #863, opened 2026-07-06
ULXs luminosity functions

AArn80:
In this pull request, I submit the functions that I have developed to calculate the X-ray luminosity of XRBs. It includes two way of calculating these luminosities: using the King et al. (2001, 2009) prescriptions and a new set of GRRMHD prescriptions. This code was developed for my Master's thesis work under the supervision of @maxbriel and @tassos25 . Tassos Fragos asked me to create a pull request with my functions as well as a small notebook on how to use my functions. Here are the files:

1) To calculate the _Bondi-Hoyle-Lyttleton_ mass accretion rate, I am using the wrong version mentioned in the issue #829 . 
2) "**Luminosity_XLF_functions26-05-28.py**" contains all the functions used to calculate and plot the cumulative X-ray luminosity functions.
3) "**XLF_Guide_functions** "(.HTML or .ipynb version (I recommend to download the .ipynb version in order to have the correct layout)) contains some explanations and example on how to use the functions that I have implemented.

### DIFF
--- a/Luminosity_XLF_functions26-05-28.py
+++ b/Luminosity_XLF_functions26-05-28.py
@@ -4,18 +4,21 @@ __author__ = "Arnaud Aguet <arnaud.aguet@etu.unige.ch><arnaud552000@gmail.com>"
 
 
 import os
+import warnings  # type : ignore
 
-from scipy.optimize import newton                                                            # type : ignore
-import matplotlib.pyplot as plt                                                              # type : ignore
+import matplotlib.pyplot as plt  # type : ignore
 import numpy as np
-import pandas as pd                                                                          # type : ignore
-import warnings                                                                              # type : ignore
+import pandas as pd  # type : ignore
+from scipy.optimize import newton  # type : ignore
 
-import posydon                                                                               # type : ignore
-import posydon.utils.constants as const                                                      # type : ignore 
+import posydon  # type : ignore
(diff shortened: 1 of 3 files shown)


sgossage:
As far as organization goes, where should we place this ULX submodule?

We could have dedicated submodules within `popsyn` for example. `posydon.popsyn.ULX` that would hold these functions, and similarly for other populations. E.g., we also have `posydon.popsyn.GRB` currently.

We also should convert the documentation into our site's usual format.

sgossage:
I think that we should consider creating a class for `ULX` pop analysis. Currently, it looks like a lot of the functions being introduced are moreso helper methods that wouldn't be called outside of specific use cases, and are called in sequence by either filtering or plotting functions.

Organizing things into an overarching class could help make the scope, use cases, and dependencies more clear.

tassos25:
@sgossage I had a similar discussion with @maxbriel who made the same suggestion. In the context of the school, I made a small PR that included only a a small subset of functions that could be of more general use and thus I thought they would go better in a submodule in teh utils module. See PR #881 . But for the filtering and plotting functions, That would probably not be a good place. 

I will keep working on this piece of code the coming weeks, but I think I will break it down to smaller PRs and not try to review and edit the whole thing at once...

tassos25:
I converted the PR to draft for now.

PR #864, opened 2026-07-08
Warn before downloading data

sgossage:
If `$PATH_TO_POSYDON` data is malformed or an expected file is missing there, our code will try to download data from Zenodo to remedy the issue.

This adds a warning that says which file triggered this response and the fact that this is happening before doing so.


maxbriel (approved):
LGTM!

astroJeff (approved):

PR #865, opened 2026-07-09
Remove `self.binary` calls in `collapse_star`

sgossage:
The `StepSN` class does not have a `binary` property (should it?) so when `collapse_star` tries to use `self.binary` within the `StepSN` class an error occurs.

This PR relocates the code that handles setting a `BinaryStar` object as failed outside of `collapse_star` and into the evolution step where references to a `BinaryStar` object do exist.

I also noticed that all SN models require:

```
        "use_interp_values": False,
        ...
        "use_core_masses": False,
```

Should we not expose these options in the .ini file then? If you change them ever, you get a bad SN model, so seems like the option shouldn't be there for now.

### DIFF
--- a/posydon/binary_evol/SN/step_SN.py
+++ b/posydon/binary_evol/SN/step_SN.py
@@ -460,13 +460,21 @@ class StepSN(object):
         # CC1 and CC2 respectively.
         if binary.event == "CC1":
             # collapse star
-            self.collapse_star(star=binary.star_1)
+            model_err = self.collapse_star(star=binary.star_1)
+            if model_err is not None:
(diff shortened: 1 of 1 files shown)


maxbriel:
I think we should define a `self.binary` or pass the binary to the collapse and go that route. This would also be helpful for using the mass transfer as part of the Maltsev prescription


@sgossage I'm not sure I get the second part.
If you use  `"use_interp_values": False`, the `SN_MODELS` are not used and the calculation is done based on the profile or the masses, depending what's there. It should not return a "bad SN model". Does it do this?

astroJeff (approved):

PR #866, opened 2026-07-11
DCO step comparison tolerance

sgossage:
In step DCO, when translating $\tau(s)$ back to physical time, we use the `brentq` root-finding method to solve the equation $f(s) = \tau(s)−\tau_{target} = 0$ within a bounded interval $s\in\left[s_{lo},s_{hi}\right]$. The `brentq` method requires that the function values at the interval endpoints (i.e., $f(s_{lo})$ and $f(s_{hi})$ ) have opposite signs, to guarantee that a root exists in that interval.

If $f(s_{lo})$ and $f(s_{hi})$ have the same sign, then this error occurs:

```python
ValueError: f(a) and f(b) must have different signs
```

In DCO evolution, this error can occur when $\tau_{target}$ is numerically very close to an endpoint value, such as $\tau(s_{lo})$. In this case, the desired root is at (or extremely close to) the integration boundary. Due to floating point rounding, $\tau_{target}$ may lie slightly outside the range of $\tau(s_{lo})$ and $\tau(s_{hi})$, causing $f(s_{lo})$ and $f(s_{hi})$ to have the same sign even though $\tau_{target}$ is effectively at the boundary.

For example, if $\tau_{target} = -10.000000000000001$, $\tau(s_{lo}) = -10$, and $\tau(s_{hi}) = -5$. The method `brentq` will see 

$$f(s_{lo}) = \tau(s_{lo}) - \tau_{target} = 10^{-15}$$

and

$$f(s_{hi}) = \tau(s_{hi}) - \tau_{target} \approx 5$$

The method then sees $f(s_{lo})$ and $f(s_{hi})$ as having the same sign, `brentq` can not bracket a root, and fails.

To handle this case, the endpoint values are checked explicitly before calling `brentq`. We define a tolerance $\epsilon$ that is based on floating point precision and if

$$|\tau_{target}-\tau(s_{lo}) | < \epsilon$$

or

$$|\tau_{target}-\tau(s_{hi}) | < \epsilon$$

the corresponding boundary value $s_{lo}$ or $s_{hi}$ is used directly instead of calling `brentq`.

>[!NOTE]
> A real example triggering this error has these values:
> `f_lo = -1.4447446103803316e-05, f_hi = -1.6940658945086007e-21`
> `tau_target[i] = 1.4447446103803316e-05, tau_lo = 0.0, tau_hi = 1.4447446103803314e-05`

### DIFF
--- a/posydon/binary_evol/DT/double_CO.py
+++ b/posydon/binary_evol/DT/double_CO.py
@@ -69,16 +69,16 @@ class _SBrentqDenseOutput:
         result = np.empty((4, len(tau_target)))
         tau_lo = self.sol(self.s_lo)[1]
         tau_hi = self.sol(self.s_hi)[1]
-        
-        # An absolute tolerance for floating-point comparisons scaled 
+
+        # An absolute tolerance for floating-point comparisons scaled
         # with |tau_hi| when tau_hi >= 1, otherwise 1
         # (so the tolerance does not become unrealistically small).
         eps = 100 * np.finfo(float).eps * max(1.0, abs(tau_hi))
-        
+
         for i, tau in enumerate(tau_target):
             f_lo = tau_lo - tau
             f_hi = tau_hi - tau
-            
+
             if f_lo * f_hi > 0:
                 print(
                     f"FAILED at i={i}: "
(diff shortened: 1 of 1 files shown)


maxbriel:
I thought I had accounted for this error already, but clearly not! Thank you for finding it and catching it! <3
The fix looks good to me

maxbriel (dismissed):

maxbriel (approved):

astroJeff (approved):

PR #868, opened 2026-07-13
[enhancement] Update default parameters

maxbriel:
This PR changes the defaults of the population synthesis file to:

- Use the [Disberg & Mandel (2025)](https://ui.adsabs.harvard.edu/abs/2025ApJ...989L...8D) natal kick distribution.
- Changes the period distribution from Sana+12-extended (based on [Bavera et al. 2021](https://ui.adsabs.harvard.edu/abs/2021A%26A...647A.153B/abstract)) to a flat log-period distribution. This gives the best balance between LMC and SMC observations.

The change in period distribution is based on [Sana et al. (2025)](https://ui.adsabs.harvard.edu/abs/2025NatAs...9.1337S)), see the excerpt below.

image

Note that the boundaries of this study run from 1 day to $10^{3.5}$ days and found $76\pm8\%$ of O stars in the SMC are in binaries with $P<10^3.5$ days (~3'162 days). The current default period boundaries or binary fraction are not made consistent with this!


sgossage:
We should be sure to change the target branch to main, from v2.3.

This looks fine to me otherwise.

PR #877, opened 2026-08-10
[v2.3] Update urls + checksums + filenames

maxbriel:
Updates the urls and checksums of the new v2.3.0 grid release.


sgossage:
Looks good -- just noting that we may not be able to test it with our `urllib` calls while the record is in an unpublished draft state.

For the future, maybe we'd want to create a debugging mode that can access drafts on Zenodo via API personal access tokens.

sgossage (approved):

PR #878, opened 2026-08-11
[docs] Automatic docstring update

maxbriel:
This PR updates the doc strings in a bunch of sub-modules.
These were either missing, outdated, or misformed for the numpy style that we use.

Additionally, replaces all `'''` with `"""`, as is preferred for docstrings.

I made this into a stacked PR (A new things). It can be added to v2.3 after it has gone into main.


sgossage:
We should change the target branch to `main`, from `v2.3`.

PR #881, opened 2026-08-11
Add a minimal XRB utility module

### LINKED ISSUE #829: Incomplete calculation of the Bondi-Hoyle accretion rate

tassos25:
## Summary

- add one self-contained, vectorized X-ray-binary utility module: `posydon.utils.xrb`
- provide scalar and NumPy-array APIs for compact-object efficiencies, Eddington quantities, wind speed, deterministic orbit-averaged BHL capture, classical luminosity/beaming, and the Be-XRB period--luminosity relation
- document units, equations, physical assumptions, limitations, and literature sources in the module docstrings

This PR deliberately adds **one production module plus its dedicated test file, and changes no existing files**. In particular, it does not alter `common_functions.bondi_hoyle`, package initializers, the documentation tree, population selection, XLF construction, plotting, GRRMHD prescriptions, magnetic neutron-star corrections, or wind-disc criteria.

The reusable classical subset is a contained alternative to the much broader proposal in #863. The observational X-ray-band/bolometric correction remains outside the library.

## Validation

- 8 focused numerical tests passed, covering scalar and broadcast-array behavior, invalid values, efficiencies, Eddington quantities, luminosity/beaming transitions and floor, deterministic BHL behavior, and the Be-XRB relation
- `python -W error::SyntaxWarning -m py_compile posydon/utils/xrb.py`
- `python -m flake8 posydon/utils/xrb.py`
- `git diff --check`

## Follow-up work

- #887 tracks consolidation of duplicated wind/BHL utilities after review
- #888 tracks the legacy `common_functions.bondi_hoyle` defects and the open physics question associated with #829
- #886 tracks tutorial-style documentation generated around this API

The new BHL API is deterministic and implements an analytic orbit average, but this PR does not claim to close #829 because it leaves the legacy public path unchanged and its physical prescription still needs discussion.

### DIFF
--- a/posydon/unit_tests/utils/test_xrb.py
+++ b/posydon/unit_tests/utils/test_xrb.py
@@ -1,0 +1,191 @@
+"""Unit tests for :mod:`posydon.utils.xrb`."""
+
+import numpy as np
+import pytest
+
+from posydon.utils import xrb
+from posydon.utils import constants as const
+
+
+def test_black_hole_radiative_efficiency_scalars_and_arrays():
+    assert isinstance(xrb.black_hole_radiative_efficiency(0.0), float)
+    assert xrb.black_hole_radiative_efficiency(0.0) == pytest.approx(
+        0.057190958417936644
+    )
+    efficiencies = xrb.black_hole_radiative_efficiency(
+        np.array([-0.998, 0.0, 0.9, 1.0, 1.1, np.nan])
+    )
+    assert efficiencies[:3] == pytest.approx(
+        [0.03777362529209353, 0.057190958417936644,
+         0.15575299199446402]
+    )
+    assert efficiencies[3] == pytest.approx(1.0 - 1.0 / np.sqrt(3.0))
+    assert np.isnan(efficiencies[4:]).all()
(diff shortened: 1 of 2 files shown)


sgossage:
This module should probably go inside of `posydon.popsyn`. That's where `GRB.py` lives, and makes sense as an import `posydon.popsyn.XRB`, for example, as this contains population analysis tools for XRBs.

As with the ULX package, a number of these functions exist as helper functions to facilitate other things, and don't seem to be intended as standalone functions themselves. It could be beneficial to organize things into a class that contains these as internal methods, to distinguish the main functionality that users should interface with.

sgossage:
This also needs documentation to be written for it.

PR #889, opened 2026-08-14
Fix CARB_braking primary clip using m1 instead of m2

maxbriel:
Currently the primary uses m2 in the `dOmega_mb_pri` calculation. Others prescriptions use `m1`


sgossage (approved):

posydon/binary_evol/DT/magnetic_braking/prescriptions.py
sgossage:
```suggestion
            * (const.rsol**2 * const.secyer**4 / const.msol**2))
```

The units here also need to be corrected

sgossage:
```suggestion
            * (const.rsol**2 * const.secyer**4 / const.msol**2))
```

The units here also need to be corrected

TheLuckyPie (approved):

PR #893, opened 2026-08-17
Fix Single He star inlist layering

### LINKED ISSUE #876: [BUG]: inlist_step2 not created for single_HeMS
**Describe the bug**
When running a single_HeMS MESA grid, `inlist_step2` was not created during the `posydon-setup-grid` step, and then the HeMS models did not run fully.

**To Reproduce**
I copied [this `.ini` file](https://github.com/POSYDON-code/POSYDON-MESA-INLISTS/blob/development/r11701/running_scripts/single_HeMS_quest.ini) and used the following scenario:
`scenario = ['posydon', 'matthias_lbv_thermo_dedt_noAgeLimit-dcc42e1d6d89fc267826970643e7f4f1bf463171', 'CO-He_star']`
I then ran `posydon-setup-grid`. I initially got the following error:
```
We are going to dynamically fetch the posydon inlists based on your scenario
You have selected posydon as your source
checking if we have already cloned POSYDON-MESA-INLISTS for you
/home/oav5601/.conda/envs/posydon_env/bin/posydon-setup-grid:84: OverwriteWarning: 'git repository is already there, using that'
checking out branch: matthias_lbv_thermo_dedt_noAgeLimit
For posterity we are pulling (specifically needed if you already have the repo clone)
checking out commit/tag: dcc42e1d6d89fc267826970643e7f4f1bf463171
Based on system_type CO-He_star We are populating the posydon inlists in the following directory: /home/oav5601/.posydon_mesa_inlists/r11701/default_common_inlists/
Running Single Grid: Setting mesa_star1_extras to /home/oav5601/.posydon_mesa_inlists/r11701/default_common_inlists//binary/src/run_star_extras.f
Running Single Grid: Setting mesa_star2_extras to /home/oav5601/.posydon_mesa_inlists/r11701/default_common_inlists//binary/src/run_star_extras.f
Updating inifile values
Based on system_type CO-He_star We are populating the user inlists in the following directory: /home/oav5601/.posydon_mesa_inlists/r11701/CO-He_star/
We are making star2 so we can unset the zams file we were going to use for star2
These are the user inlists used in the single star grid: ['/home/oav5601/.posydon_mesa_inlists/r11701/CO-He_star/star1_formation/inlist_step1', '/home/oav5601/.posydon_mesa_inlists/r11701/CO-He_star/star1_formation/inlist_step2']
You want a single star He grid, this means that we need to make the inlist that will be used to evolve the system and make sure we layer on the line x_logical_ctrl(1)=.true.
/home/oav5601/.conda/envs/posydon_env/bin/posydon-setup-grid:356: OverwriteWarning: 'Replace /gpfs/home/oav5601/.posydon_mesa_inlists/special_single_star_user_inlist'
  File "/home/oav5601/.conda/envs/posydon_env/bin/posydon-setup-grid", line 1024, in <module>
    find_inlist_from_scenario(source=mesa_inlists['scenario'][0],
  File "/home/oav5601/.conda/envs/posydon_env/bin/posydon-setup-grid", line 372, in find_inlist_from_scenario
    mesa_inlists['star1_controls_special'].append(special_single_star_user_inlist)
    ~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^
KeyError: 'star1_controls_special'
```
This occurred because `mesa_inlists['star1_controls_special']` did not already exist, and I fixed it by replacing the line with `pass` since it did not get referenced again.
However, although `posydon-setup-grid` would now run without error, it did not create an `inlist_step2` file. Additionally, when I tried to run `./run_grid.sh`, it threw errors that the `Zbase` of my ZAMS model did not match the one in the `inlist` files, rather than copying the `Zbase` (and `initial_z` and `new_Z`) from my `grid.csv` file.

**Expected behavior**
I expected it to create an `inlist_step2` alongside steps 0 and 1.

**POSYDON Version:**
The error was with version 2.2.8.
Was able to fix it by using an older version, running `source /projects/b1119/.bash_shared_fixed_rerun_implicit`

**System Configuration (please complete the following information):**
Quest, Slurm

**Additional context**

sgossage:
The inlist layering in single star grids was modified in the past to avoid overwriting any provided `star1_controls_user`. However, that patch took an approach that was incompatible with subsequent star formation scenario inlist layering. In `construct_static_inlist`, 

```python
        for step in range(number_of_star1_formation_steps):
            star1_formation['step{0}'.format(step)] = {}
            star1_formation['step{0}'.format(step)]['inlist_file'] = os.path.join(working_directory, 'star1', 'inlist_step{0}'.format(step))
            for k, v in star1_formation_dictionary.items():
                star1_formation['step{0}'.format(step)][k] = v[step] if type(v) == list else v
```

will look for star formation settings in `mesa_inlist['star1_job_*']` and `mesa_inlist['star1_controls_*']`. The previous fix added star1_controls_special as a separate, single-item entry rather than adding the special inlist to the existing star1_controls_user sequence. As a result, the special inlist was not layered consistently with the star-formation scenario inlists, and the resulting sequence could be incomplete.

This PR instead keeps the existing star1_job_user and star1_controls_user dictionary items and removes the separate star1_controls_special entry. It prevents the original overwrite by modifying the line that caused the issue:

```python
    if system_type == "HMS-HMS" and mesa_inlists['single_star_grid']:
        ...
        mesa_inlists['star1_controls_user'] = special_single_star_user_inlist
```

changing this to

```python
        if system_type == "HMS-HMS" and mesa_inlists['single_star_grid']:
            ...
            if 'star1_controls_user' not in mesa_inlists or mesa_inlists['star1_controls_user'] is None:
                mesa_inlists['star1_controls_user'] = []

            mesa_inlists['star1_controls_user'].append(special_single_star_user_inlist)
```

This preserves any existing star-formation inlists while adding the special single-star control inlist, following the same layering approach already used for the CO-He_star scenario.

### DIFF
--- a/bin/posydon-setup-grid
+++ b/bin/posydon-setup-grid
@@ -323,14 +323,15 @@ def find_inlist_from_scenario(source, gitcommit, system_type):
             with open(special_single_star_user_inlist, 'wb') as f:
                 f.write(b'&controls\n\n')
                 f.write('\t{0} = {1}\n'.format("x_logical_ctrl(1)", ".true.").encode('utf-8'))
-
                 f.write(b'\n\n')
-
                 f.write(b"""
-        / ! end of star1_controls namelist
+                         / ! end of star1_controls namelist
+                         """)
+            if 'star1_controls_user' not in mesa_inlists or mesa_inlists['star1_controls_user'] is None:
+                mesa_inlists['star1_controls_user'] = []
 
-        """)
-            mesa_inlists['star1_controls_special'] = special_single_star_user_inlist
+            mesa_inlists['star1_controls_user'].append(special_single_star_user_inlist)
+
         elif system_type == "CO-He_star" and mesa_inlists['single_star_grid']:
             inlists_location = '{0}/{1}/{2}/'.format(inlists_dir, 'r11701', system_type)
             print("Based on system_type {0} "
@@ -358,18 +359,15 @@ def find_inlist_from_scenario(source, gitcommit, system_type):
             with open(special_single_star_user_inlist, 'wb') as f:
                 f.write(b'&controls\n\n')
                 f.write('\t{0} = {1}\n'.format("x_logical_ctrl(1)", ".true.").encode('utf-8'))
-
                 f.write(b'\n\n')
                 f.write(b"""
-        / ! end of star1_controls namelist
-
-        """)
+                         / ! end of star1_controls namelist
+                         """)
                 f.write(b'&star_job\n\n')
                 f.write(b"""
-        / / ! end of star_job namelist
-
-        """)
-            mesa_inlists['star1_controls_special'].append(special_single_star_user_inlist)
+                         / / ! end of star_job namelist
+                         """)
+            mesa_inlists['star1_controls_user'].append(special_single_star_user_inlist)
 
     # change back to where I was
     os.chdir(where_am_i_now)
--- a/bin/posydon-setup-grid
+++ b/bin/posydon-setup-grid
@@ -345,8 +345,12 @@ def find_inlist_from_scenario(source, gitcommit, system_type):
             # Find the user single star controls
             single_star_scenario = sorted(glob.glob(os.path.join(inlists_location, "star1_formation", "*step*")))
             print("These are the user inlists used in the single star grid: {0}".format(single_star_scenario))
-            mesa_inlists['star1_controls_user'] = single_star_scenario
-            mesa_inlists['star1_job_user'] = single_star_scenario
+            if 'star1_controls_user' not in mesa_inlists or mesa_inlists['star1_controls_user'] is None:
+                mesa_inlists['star1_controls_user'] = []
+            if 'star1_job_user' not in mesa_inlists or mesa_inlists['star1_job_user'] is None:
+                mesa_inlists['star1_job_user'] = []
+            mesa_inlists['star1_controls_user'].append(single_star_scenario)
+            mesa_inlists['star1_job_user'].append(single_star_scenario)
             print("You want a single star He grid, "
                   "this means that we need to make the inlist that will be used to evolve the system "
                   "and make sure we layer on the line "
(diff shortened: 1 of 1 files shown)


bin/posydon-setup-grid
```diff
@@ -323,14 +323,15 @@ def find_inlist_from_scenario(source, gitcommit, system_type):
...
                 f.write(b'\n\n')
-
                 f.write(b"""
-        / ! end of star1_controls namelist
+                         / ! end of star1_controls namelist
+                         """)
+            if 'star1_controls_user' not in mesa_inlists or mesa_inlists['star1_controls_user'] is None:
+                mesa_inlists['star1_controls_user'] = []
+
+            mesa_inlists['star1_controls_user'].append(special_single_star_user_inlist)
```
maxbriel, line 336: Are you sure that this is a list already? I thought that it was initialised as a string (with the filepath)

maxbriel (commented):
One question, but otherwise looks good

bin/posydon-setup-grid
sgossage: In the case of single HMS I'm sure, but actually we need to add something more to the single HeMS case.

In general, `mesa_inlists` is first defined in `main()` after parsing the .ini file.

Later in `find_inlist_from_scenario`, this dictionary gets modified in various ways. Focusing on the single stars though:

- In the single HMS case, that `if` statement above is the first check since reading the ini file as to whether or not `mesa_inlists['star1_controls_user]` exists or not (or if it is set to `None`). If it does not exist, we need to create it, because we need it to hold at least `special_single_star_user_inlist`. So, if isn't in the .ini file, it is initialized as an empty list here, then `star1_controls_user` is always appended to that list, or whatever pre-existing list there may have been.
- In the single HeMS case, there's a line, actually two, that force `star1_controls_user` to a specific set of inlists:

```python
348            mesa_inlists['star1_controls_user'] = single_star_scenario
349            mesa_inlists['star1_job_user'] = single_star_scenario
```

so that will overwrite anything that was set in .ini in the case of the single He star grid. 

I've now added a similar `if` statement in the single HeMS case now that checks if these `dict` items exist already, and if so, append `single_star_scenario`, and subsequently, append `special_single_star_user_linlist` as well.

maxbriel (approved):

PR #788, opened 2026-01-16
[enhancement] HMS+HMS Mergers rejuvenation according to Schneider+2016

ezapartas:
- This PR is changing the default option in the POSYDON .ini file of `HMS_HMS_merging_rejuvenation = True` to stronger rejuvenation of MS+MS mergers (according to Glebbeek+2023, Schneider+2016, following the recent Wang+2025 too). So it should be considered a change in POSYDON results for few types mergers

-This PR includes already the changes in #785 and builds on it. If accepted, that PR becomes unnecessary

PR #902, opened 2026-08-23
[enhancements] PR 788 into main (HMS+HMS merger)

maxbriel:
PR #788

### DIFF
--- a/posydon/binary_evol/DT/step_merged.py
+++ b/posydon/binary_evol/DT/step_merged.py
@@ -22,6 +22,7 @@ from posydon.binary_evol.singlestar import (
     convert_star_to_massless_remnant,
 )
 from posydon.utils.common_functions import check_state_of_star
+from posydon.utils.constants import Zsun, zams_table
 from posydon.utils.posydonerror import ModelError
 from posydon.utils.posydonwarning import Pwarn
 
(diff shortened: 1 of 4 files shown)


sgossage:
Overall, from the previous PR, this looks fine to me.

posydon/binary_evol/DT/step_merged.py
```diff
@@ -244,21 +250,77 @@ def merged_star_properties(self, star_base, comp):
...
+                        'rel_mass_lost_HMS_HMS must be either a float in [0, 1) '
+                        'or the string "Glebbeek+2013".'
+                    )
+
+
+                if self.HMS_HMS_merging_rejuvenation:  # according to Schneider+2016
+
+                    Z_div_Zsun_init = star_base.metallicity
+                    if Z_div_Zsun_init in zams_table.keys():
+                        Y_initial = zams_table[Z_div_Zsun_init]
```
sgossage, line 271: Maybe a slight inconsistency here. Just to be clear, this may be more conceptually inconsistent, but probably not numerically very impactful.

Later we are apparently scaling Y_He4, but this initial Y contains both He3 and He4. The value from the ZAMS table is 0.2703 (e.g., for solar anyways), which is the bulk initial Y value.

MESA takes this input and assigns initial He3 and He4 abundances according to their supposed primordial abundance ratios from Anders and Grevesse 1989, unless `initial_he3` is specified, which it is not in our ZAMS inlists.

The difference is pretty small though. A&G89 have the ratio of He3/He4 = 1.42e-4, and by mass,

Y3/Y4 = 1.42e-4 * 3/4 = 1.065e-4

So, for the solar case, initial Y4 comes out to `0.27027123463996966` and initial Y3 to `2.8765360029938486e-05`.

Later on, we're calculating

```
                    Y_average_merged = (star_base.total_mass_he4 + comp.total_mass_he4
                                        - rel_mass_lost_HMS_HMS * total_mass * Y_initial) / merged_star.mass
```

Using the bulk Y (`Y_initial`) and `total_mass_he4` (ignoring He3). This probably doesn't matter much, except in cases where He3 is significantly produced (e.g., low mass stars). To be consistent anyways, we should probably be using `total_mass_he3 + total_mass_he4` here. In our current grids, we only output `total_mass_he4` though. I think Schneider+ 2016 is meant to use bulk Y in Eq. 8 (not just He4)?

So, maybe not much we can do at the moment since we don't have `total_mass_he3`, but I wanted to flag this.

sgossage: Likewise, in the `parameters_to_mix` lists used in subsequent `if..else` blocks, should we add `center_he3` and `surface_he3`, since it forms a part of the bulk Y?

He3 likely remains small, maybe on the order of 1e-3 or 1e-4 at most, so probably negligible, as above.

maxbriel: @ezapartas what is your opinion on this?

PR #903, opened 2026-08-23
[enhancement] Add workflow for running validation binaries

maxbriel:
This adds a workflow for us to run the validation binaries across metallicity to test changes to the output. This runs on a self-hosted runner and can be ran by adding a comment with /run-baseline-check to a comment by one of us.

It runs the binaries, collects the output files, and adds them + a summary to the PR as a comment.
This still needs testing (after being merged into main).

Supersedes PR #873

### DIFF
--- a/.github/workflows/baseline-validation-on-comment.yml
+++ b/.github/workflows/baseline-validation-on-comment.yml
@@ -1,0 +1,341 @@
+name: Baseline Validation (Comment Triggered)
+
+on:
+  issue_comment:
+    types: [created]
+
+# Only one validation run per PR at a time; a new trigger comment supersedes
+# an in-flight run for the same PR.
+concurrency:
+  group: baseline-validation-${{ github.event.issue.number }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+  issues: write
+  pull-requests: write
+
+env:
+  # Marker used to find/update the single "sticky" comment on the PR.
+  STICKY_MARKER: "<!-- posydon-baseline-validation -->"
+
+jobs:
+  check-trigger:
+    # Only run if the comment is on a PR and contains the trigger command
+    if: github.event.issue.pull_request && contains(github.event.comment.body, '/run-baseline-check')
+    runs-on: self-hosted
+    outputs:
+      should_run: ${{ steps.check.outputs.should_run }}
+      pr_number: ${{ steps.get-pr.outputs.pr_number }}
+      pr_branch: ${{ steps.get-pr.outputs.pr_branch }}
+      pr_sha: ${{ steps.get-pr.outputs.pr_sha }}
+      base_branch: ${{ steps.get-pr.outputs.base_branch }}
+      base_sha: ${{ steps.get-pr.outputs.base_sha }}
+
+    steps:
+      - name: Check if user has permission
+        id: check
+        uses: actions/github-script@v7
+        with:
+          script: |
+            const { data: permission } = await github.rest.repos.getCollaboratorPermissionLevel({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              username: context.payload.comment.user.login
+            });
+
+            const hasPermission = ['admin', 'write'].includes(permission.permission);
+            core.setOutput('should_run', hasPermission);
+
+            if (!hasPermission) {
+              await github.rest.issues.createComment({
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                issue_number: context.issue.number,
+                body: `❌ @${context.payload.comment.user.login} does not have permission to trigger baseline validation.`
+              });
+            }
+
+      - name: React to comment
+        if: steps.check.outputs.should_run == 'true'
+        uses: actions/github-script@v7
+        with:
+          script: |
+            await github.rest.reactions.createForIssueComment({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              comment_id: context.payload.comment.id,
+              content: 'rocket'
+            });
+
+      - name: Get PR details
+        id: get-pr
+        if: steps.check.outputs.should_run == 'true'
+        uses: actions/github-script@v7
+        with:
+          script: |
+            const { data: pr } = await github.rest.pulls.get({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              pull_number: context.issue.number
+            });
+
+            core.setOutput('pr_number', pr.number);
+            core.setOutput('pr_branch', pr.head.ref);
+            core.setOutput('pr_sha', pr.head.sha);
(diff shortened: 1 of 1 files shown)


sgossage:
~@maxbriel, were you able to test this workflow or is that TBD?~

Nevermind, I see that it needs testing after merging into `main`.

.github/workflows/baseline-validation-on-comment.yml
```diff
@@ -0,0 +1,341 @@
...
+  pull-requests: write
+
+env:
+  # Marker used to find/update the single "sticky" comment on the PR.
+  STICKY_MARKER: "<!-- posydon-baseline-validation -->"
+
+jobs:
+  check-trigger:
+    # Only run if the comment is on a PR and contains the trigger command
+    if: github.event.issue.pull_request && contains(github.event.comment.body, '/run-baseline-check')
```
sgossage, line 25: I think this means that any comment containing the command could trigger the workflow to run (even it is just mentioned for example).

Maybe we want to make the parsing more strict. E.g., `startsWith(github.event.comment.body, '/run-baseline-check')` instead of `contains`?

.github/workflows/baseline-validation-on-comment.yml
```diff
@@ -0,0 +1,341 @@
...
+        if: always()
+        run: |
+          if [ "${{ steps.baseline-cache.outputs.cache-hit }}" != "true" ] && [ "${{ steps.generate-baseline.outcome }}" == "failure" ]; then
+            echo "Baseline generation failed"
+            exit 1
+          fi
+          if [ "${{ steps.validate.outcome }}" != "success" ] && [ "${{ steps.validate.outcome }}" != "failure" ]; then
+            echo "Validation step errored (not a diff failure)"
+            exit 1
+          fi
```
sgossage, line 340: What's an example instance where this is triggered?

Since we're using `continue-on-error` above, is it possible for the baseline/validation to still encounter an infrastructure error and continue, thus producing `steps.validate.outcome == "failure"` here and skipping this block?

In `validate_binaries.sh`, we report the exit codes

```
exit 0 (no errors, all passed)
exit 1 (something happened, could be differences detected or an infrastructural error)
```

I.e., I don't think we're distinguishing infrastructure failures from cases where differences are detected. Both produce `exit 1`. 

We may want to use a distinct exit code for differences detected. E.g., in `validate_binaries.sh`,

```
...
if [ $FAIL -gt 0 ]; then
    exit 10
fi
...
```
instead of `exit 1`.

sgossage: If we do this, we would need to capture the exit code, e.g., for the validation step:

```
- name: Run validation on PR branch
  id: validate
  continue-on-error: true
  working-directory: dev-tools
  env:
    PATH_TO_POSYDON_DATA: /data
    PR_BRANCH: ${{ needs.check-trigger.outputs.pr_branch }}
    BASE_BRANCH: ${{ needs.check-trigger.outputs.base_branch }}
  run: |
    set +e
    ./validate_binaries.sh "$PR_BRANCH" "$BASE_BRANCH"
    status=$?
    echo "exit_code=$status" >> "$GITHUB_OUTPUT"
    exit "$status"
```

Then we could use `steps.validate.outputs.exit_code` in the final step that writes the outcome in a comment to distinguish the three cases.

We could do

```
      - name: Update sticky comment with results
        if: always()
        uses: actions/github-script@v7
        with:
        ...
        const validationExitCode = Number('${{ steps.validate.outputs.exit_code }}');
        ...
       comment += '### Step 2: Binary Validation\n';

       if (validationExitCode === 0) {
           comment += '✅ Validation completed successfully — no differences found\n\n';
       } else if (validationExitCode === 10) {
           comment += '⚠️ Validation completed with differences (see below) — this is informational and does not block merging\n\n';
       } else {
           comment += `❌ Validation failed with exit code ${validationExitCode} — see the workflow logs for details\n\n`;
       }
```

instead of what we currently have there with `validationStatus`. Finally, for the step determining the final status:

```
- name: Set final status
  if: always()
  run: |
    if [ "${{ steps.baseline-cache.outputs.cache-hit }}" != "true" ] && [ "${{ steps.generate-baseline.outcome }}" == "failure" ]; then
      echo "Baseline generation failed"
      exit 1
    fi

    exit_code="${{ steps.validate.outputs.exit_code }}"

    if [ "$exit_code" -ne 0 ] && [ "$exit_code" -ne 10 ]; then
      echo "Validation step failed with infrastructure/error code $exit_code"
      exit 1
    fi

    echo "Workflow completed"
```

using the `exit_code` as well.

sgossage: In `validate_binaries.sh`, we may have another issue. When we run `compare_runs.py`, that Python script has three exit codes:

```
exit 0 (success, no differences)
exit 1 (infrastructural failure)
exit 2 (differences detected)
```

and then in `validate_binaries.sh`, when we run this, we do

```
    if python "$SRC_DIR/compare_runs.py" "$BASELINE_FILE" "$CANDIDATE_FILE" \
        $COMPARE_FLAGS \
        2>&1 | tee "$COMPARISON_FILE"; then
        echo "  PASS: No differences"
        echo "Z = ${Z} Zsun: PASS" >> "$SUMMARY_FILE"
        PASS=$((PASS + 1))
    else
        echo "  DIFFERENCES DETECTED — see $COMPARISON_FILE"
        echo "Z = ${Z} Zsun: DIFFERENCES DETECTED (see comparison_${Z}Zsun.txt)" >> "$SUMMARY_FILE"
        FAIL=$((FAIL + 1))
    fi
```

which does not distinguish between `exit 1` and `exit 2`; both are treated as differences detected.

We could do something like this instead there:

```
if python "$SRC_DIR/compare_runs.py" "$BASELINE_FILE" "$CANDIDATE_FILE" \
    $COMPARE_FLAGS \
    2>&1 | tee "$COMPARISON_FILE"; then

    ...

else
    status=${PIPESTATUS[0]}

    if [ "$status" -eq 2 ]; then
        echo "  DIFFERENCES DETECTED — see $COMPARISON_FILE"
        echo "Z = ${Z} Zsun: DIFFERENCES DETECTED (see comparison_${Z}Zsun.txt)" >> "$SUMMARY_FILE"
        FAIL=$((FAIL + 1))
    else
        echo "  ERROR: Comparison failed with exit code $status" >&2
        echo "Z = ${Z} Zsun: ERROR (comparison failed, exit code $status)" >> "$SUMMARY_FILE"
        exit "$status"
    fi
fi
```

so that the script actually is sent to fail when `exit 1` is returned by `compare_runs.py`. This then goes downstream in `validate_binaries.sh` to produce an `exit 1` code with an infrastructural error from `compare_runs.py` being detected consistently in `validate_binaries.sh` and `exit 10` being used to indicate that differences were detected.

PR #904, opened 2026-08-23
[enhancement] Add check for already downloaded data

### LINKED ISSUE #696: [BUG]: Downloaded data, redownloaded
**Describe the bug**
When downloading the data, we should check for the already existing data, such that the user doesn't redownload the same data, unless it is explicitly requested.

For example, I downloaded the `auxiliary` data and then downloaded `DR2_1Zsun`, which redownloaded the `auxiliary` data first. 

**To Reproduce**

```
get-posydon-data auxiliary
get-posydon-data DR2_1Zsun

```

**Expected behavior**
Skip the `auxiliary` data if already present/up-to-date.

**POSYDON Version:**
v2.1.1

maxbriel:
This checks if datasets are already downloaded to allow you to repeat the same command as before, e.g. 

```
get-posydon-data DR2
```
With the auxilliray data, 2Zsun, and 1Zsun already downloaded, gives: 

```
(posydon-debug) ➜  ~ get-posydon-data DR2
POSYDON data 'auxiliary' is already present, skipping.
POSYDON data 'DR2_grids_2Zsun' is already present, skipping.
POSYDON data 'DR2_grids_1Zsun' is already present, skipping.
Downloading POSYDON data 'DR2_grids_0.45Zsun' from Zenodo to /Users/max/Documents/POSYDON_popsynth_data/v2/v2.3.0
```

additional adds:
- downloads in progress are names <file>.tar.gz.part and are renamed on success. The leftover complete tarball is MD5-verified and extracted (similar to before). But if the file already exists, instead of raising a FileExistsError, a corrupt leftover is deleted and re-downloaded.
- a new `-f/--force` to re-download regardless of file existence is added.

### DIFF
--- a/posydon/unit_tests/utils/test_data_download.py
+++ b/posydon/unit_tests/utils/test_data_download.py
@@ -12,13 +12,14 @@ import posydon.utils.data_download as totest
 # aliases
 os = totest.os
 
-from contextlib import chdir
-from inspect import isclass, isroutine
-from shutil import rmtree
-from unittest.mock import patch
-
 # import other needed code for the tests, which is not already imported in the
 # module you like to test
+import hashlib
+from contextlib import chdir
+from inspect import isclass, isroutine
+from shutil import copyfile, rmtree
+from unittest.mock import patch
+
 from pytest import approx, fixture, raises, warns
 
 from posydon.utils.posydonwarning import ReplaceValueWarning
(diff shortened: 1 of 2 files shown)


astroJeff:
I have not gone through the code carefully, but I can confirm that the code checks for whether the downloaded data already exists

maxbriel:
This PR closes issue #696

sgossage (approved):

sgossage:
We should also assess the points raised in [PR#930](https://github.com/POSYDON-code/POSYDON/issues/930), as this PR addresses that issue for the most part. Depending on design goals (i.e., whether or not we should implement an automatic retry in case the Zenodo connection drops), it may address it fully.

My initial opinion is that we may actually want to avoid an automatic retry, and leave that to the discretion of the user to rerun the command. However, that does mean the data download needs more babysitting if left to run overnight, for example.

PR #859, opened 2026-06-25
[enhancement] Maltsev above 10Msun

maxbriel:
Implements the "rapid" Maltsev prescription, discussed in [Section 3.2.3](http://arxiv.org/abs/2503.23856), for systems outside the region of Patton & Sukhbold (2020) for $M_\mathrm{CO}>10M_\odot$ masses.

This implements the following boundaries based on the values in Tables 3,4,5, and 6:
- M_CO<M_NS_1                       -> Fallback BH / NS (In most cases M_NS_1 is lower than 10Msun.
- M_NS_1<M_CO<M_NS_2       -> NS formation
- M_NS_2 < M_CO < M_CO_3  -> Fallback BH / NS
- M_CO_3 < M_CO                     -> Direct BH formation 

M_NS_1, M_NS_2, and M_CO_3 are interpolated across metallicity following the linear formula in $\log$ metallicity space.

Using a 15% probability for fallback BHs (though 10% is also mentioned in Section 3.2.3, unclear why @ezapartas)

To-do:

- [x] Implement boundaries in M_CO from Maltsev+25
- [ ] Add categorisation based on MT type from the binary
- [ ] What to do with the NS mass?

PR #920, opened 2026-08-29
[enhancement] Rapid Maltsev+25

maxbriel:
This PR adds a "rapid" Maltsev+25 prescription. 
The prescription uses the M_CO and metallicity to determine:
1. The explodibility
2. NS or BH formation

The prescription uses the mass transfer type (Case A, B,C) to determine how the boundaries in $M_\mathrm{CO}$ shift.
While in our grids, we have the `cumulative_mt_cases` available, we do not train a classifier on them, because there are too many sub-categories to train properly.
I've also implemented the 3 extrapolation methods from [Willcox+2025](https://arxiv.org/pdf/2510.07573). This introduces `balanced`, `optimistic`, and `pessimistic` options. From discussions with Schneider, the balanced seems "best" compatible with more recent core-collapse models.

As a solution, I created a `first_mt_case`, which capture the first interaction a system has into `case_A`, `case_B`, `case_C`, and other none case termination_flags originally available in `cumulative_mt_cases`.

The `first_mt_case` is added to the grid when created or using the extra script, `scripts/backfill_first_mt_case.py`. This scripts adds the column to `final_values` and overwrites the original file. **Using this prescription in an IF run, requires retraining the interpolators!**

For the Maltsev+25 prescription, this information is added to the `star.first_mt_class` during `step_mesa` such that it is available during `step_SN` where only the star is passed into the function. There the `star.first_mt_class` is made compatible with the Maltsev+25 classes. 

> [!IMPORTANT]
> Since no remnant mass prescription is available for the neutron stars, the model assumes a NS of $1.4M_\odot$. This can be swapped out, but care should be taken when using it!

> [!NOTE]
> I have not made any changes for self-stripped stars. In the appendix of the paper, these are mentioned as having to be classified as Case B systems.

> [!WARNING]
> This PR requires retraining of the interpolators and/or additional post-processing of the grids (adding a column).


maxbriel:
Based on PR #859 comment, the probability between for fallback BH should be:

>this can be 10% in the end. It's ~10% between M_CO = M_CO_NS_2-M_CO_3 whereas it is 15% for all the range of >M_CO cores that make successful SN with a possibility of fallback BHs

sgossage:
I think `backfill_first_mt_case.py` may fit better inside `posydon.utils`, rather than creating a new `posydon.scripts`.

Although, I know that reorganizing `utils` and `common_functions.py` is an ongoing effort.

What is the envisioned use case for `posydon.scripts` vs. `posydon.utils`?

Rather than being a script that's called, should `backfill_first_mt_case` be a method of `PSyGrid`, for example?

posydon/binary_evol/MESA/step_mesa.py
```diff
@@ -742,6 +742,15 @@ def update_properties_NN(self, star_1_CO=False, star_2_CO=False,
...
         setattr(self.binary, f'mt_history_{self.grid_type}', mt_history)
 
+        # TODO: we can also add first_mt_case directly from the grid.
+        # add mt_class to the star
+        # first mass transfer episode (whichever star is the donor)
+        first_mt_case = cf.mt_class_from_cumulative(cumulative_mt_case)
+        setattr(self.binary, f'first_mt_case_{self.grid_type}',
+                first_mt_case)
+        for star in stars:
+            star.mt_class = first_mt_case
```
sgossage:
```suggestion
            star.first_mt_class = first_mt_case
```

Here and elsewhere, should we call it something like `first_mt_class`? Is there ever an instance where it is not the first MT case? If not, making this distinction could be good for clarity.

sgossage:
Why do we only care about the first MT case? Is it because of the comment in Maltsev+25, Appendix A.5.1? 

> Repeated MT episodes: In the case of repeated MT episodes,
in particular Case AB and Case BC donors, we suggest to proceed
as follows. For Case AB donors, whether MT class A or
B is assigned to the star, does not substantially change the final
CCSN outcome, since the pre–SN properties of Case A and
B progenitors are similar. For Case BC donors, we suggest to
use the first MT episode for the classification, because Case C
progenitors are closer to single stars, and the binary interaction
effects are likely more adequately accounted for by the critical
values we found for Case B donors.

If so, we should make a comment somewhere in a doc string or something that this is the rationale (just referring to this section in the Appendix, for example, not the whole text).

posydon/utils/common_functions.py
```diff
@@ -1749,6 +1750,65 @@ def cumulative_mass_transfer_flag(MT_cases, shift_cases=False):
     )
 
 
+def mt_class_from_cumulative(cumulative_mt_case=None, star_index=None):
```
sgossage:
```suggestion
def first_mt_class_from_cumulative(cumulative_mt_case=None, star_index=None):
```

In line with the other suggestion I made in `step_mesa` to clarify that this is derived from the first mt case.

Or, maybe we could generalize this and have a `first` and `last` mode or something. I guess we don't have a use case for the `last` MT case at the moment (or do we in plotting?).

That way, we could store both, rather than just the first. (I can imagine someone noticing we have one and not the other, in case they wanted to access it in a similar fashion for some reason.)

maxbriel: Changed to first_mt_class throughout! 

We already sort of keep the final interaction, as mt_history, but it's not super clear because it combined `termination_flag_1` and `termination_flag_2` together. If the need arises, we can include the last_mt_case as well.

maxbriel:
> I think `backfill_first_mt_case.py` may fit better inside `posydon.utils`, rather than creating a new `posydon.scripts`.
> 
> Although, I know that reorganizing `utils` and `common_functions.py` is an ongoing effort.
> 
> What is the envisioned use case for `posydon.scripts` vs. `posydon.utils`?
> 
> Rather than being a script that's called, should `backfill_first_mt_case` be a method of `PSyGrid`, for example?

**In principle, this script isn't necessary. When we re-post-process the grids, the extra column will be present. 
It's something I included to allow you to use the current grids + validation/random grids to train it while we wait for this.
Additionally it's in a folder outside the `posydon` module, so it shouldn't be accessible inside the modules, e.g. `posydon.scripts` won't work. I would even say that we shouldn't include it when merging, it's just useful for testing :)** 


> Why do we only care about the first MT case? Is it because of the comment in Maltsev+25, Appendix A.5.1?
> 
>     Repeated MT episodes: In the case of repeated MT episodes,
>     in particular Case AB and Case BC donors, we suggest to proceed
>     as follows. For Case AB donors, whether MT class A or
>     B is assigned to the star, does not substantially change the final
>     CCSN outcome, since the pre–SN properties of Case A and
>     B progenitors are similar. For Case BC donors, we suggest to
>     use the first MT episode for the classification, because Case C
>     progenitors are closer to single stars, and the binary interaction
>     effects are likely more adequately accounted for by the critical
>     values we found for Case B donors.
> 
> If so, we should make a comment somewhere in a doc string or something that this is the rationale (just referring to this section in the Appendix, for example, not the whole text).

**Yes. That's the reason and from discussion with @ezapartas and @dimsour94. I've added a reference to the appendix.**

posydon/grids/SN_MODELS.py
```diff
@@ -517,6 +518,78 @@
...
+        "use_interp_values": False,
+#        "use_profiles": True,
+        "use_core_masses": False,
+        "Maltsev25_MCO_extrapolation_mode": "balanced",
+#        "allow_spin_None" : False,
+#        "approx_at_he_depletion": False,
+    },
+    "SN_MODEL_v2_32": {
+        "mechanism": "Maltsev+25-MCO-rapid",
+        "engine": "M16",
```
maxbriel, line 579: Change to None

posydon/grids/SN_MODELS.py
```diff
@@ -517,6 +518,78 @@
...
+        "use_interp_values": False,
+#        "use_profiles": True,
+        "use_core_masses": False,
+        "Maltsev25_MCO_extrapolation_mode": "balanced",
+#        "allow_spin_None" : False,
+#        "approx_at_he_depletion": False,
+    },
+    "SN_MODEL_v2_31": {
+        "mechanism": "Maltsev+25-MCO-rapid",
+        "engine": "M16",
```
maxbriel, line 561: Here too

posydon/grids/SN_MODELS.py
```diff
@@ -517,6 +518,78 @@
...
+        "use_interp_values": False,
+#        "use_profiles": True,
+        "use_core_masses": False,
+        "Maltsev25_MCO_extrapolation_mode": "balanced",
+#        "allow_spin_None" : False,
+#        "approx_at_he_depletion": False,
+    },
+    "SN_MODEL_v2_30": {
+        "mechanism": "Maltsev+25-MCO-rapid",
+        "engine": "M16",
```
maxbriel, line 543: Here

posydon/grids/SN_MODELS.py
```diff
@@ -517,6 +518,78 @@
 #        "use_profiles": True,
         "use_core_masses": False,
 #        "allow_spin_None" : False,
+#        "approx_at_he_depletion": False,
+    },
+    "SN_MODEL_v2_29": {
+        "mechanism": "Maltsev+25-MCO-rapid",
+        "engine": "M16",
```
maxbriel, line 525: And here

posydon/grids/SN_MODELS.py
```diff
@@ -36,6 +36,7 @@
     "use_core_masses": True,
     "allow_spin_None" : False,
     "approx_at_he_depletion": False,
+    "Maltsev25_MCO_extrapolation_mode": "balanced",
```
maxbriel, line 39: Adding this parameter requires it to become part of all SN_MODELS for the check_SN_models. This is not what we want. Other prescriptions should work without requiring this input parameter.

posydon/CLI/popsyn/setup.py
```diff
@@ -38,6 +42,20 @@ def check_SN_MODEL_validity(ini_file, verbose_on_fail=True):
...
+    # be identifiable. Report them separately, because otherwise the model
+    # matching below only reports that nothing matched.
+    missing = missing_SN_MODEL_parameters(step_SN_MODEL)
+    if len(missing) > 0:
+        mechanism = step_SN_MODEL.get('mechanism',
+                                      DEFAULT_SN_MODEL['mechanism'])
+        Pwarn(f"The mechanism '{mechanism}' requires {missing} to be set in "
+              "the [step_SN] section of the ini file.",
+              "IncompletenessWarning")
+        return False
```
sgossage, line 57: These lines are currently missed in the unit tests:

```python
    if len(missing) > 0:
        mechanism = step_SN_MODEL.get('mechanism',
                                      DEFAULT_SN_MODEL['mechanism'])
        Pwarn(f"The mechanism '{mechanism}' requires {missing} to be set in "
              "the [step_SN] section of the ini file.",
              "IncompletenessWarning")
        return False
```

DavidRAguileraDena (dismissed):
Commented by mistake!

PR #924, opened 2026-09-01
[Enhancement] direct-collapse no kick

maxbriel:
This allows you to force all direct-collapse BHs to receive no kick.


posydon/popsyn/population_params_default.ini
```diff
@@ -395,6 +395,8 @@
     # float (0,inf)
   mean_kick_ECSN = None
     # float or None
+  no_kick_direct_collapse = False
+    # True, False. If True, no kick for BHs formed by direct collapse (f_fb = 1)
```
sgossage, line 399:
```suggestion
    # True, False. If True, no kick for BHs formed by direct collapse (fallback fraction, f_fb = 1)
```

DavidRAguileraDena (approved):
LGTM!

I have a question, though, just to make sure I understand the use case of the option.

One of the conditions to set Vkick = 0 if no_kick_direct_collapse = True is to have f_fb=1 (which, of course, makes sense, to make sure you have direct collapse!). As far as I can tell, f_fb is only ever used to calculate the kick normalisation, and except for in the Fryer mechanisms and in the fallback SN case in the Maltsev mechanism (where it's set to f_fb = 0.99), it's only ever set to either 1.0 for BHs, 0.0 for NSs.

So the only time where setting no_kick_direct_collapse = True changes the output is when you have kick_normalisation = 'one' or 'one_over_mass'? Because, with the other normalisation options, you would have the kick set to 0 anyway for direct collapse BHs, right?

sgossage (approved):

PR #925, opened 2026-09-09
Fix box scaling for IllustrisTNG SFRD

sashalvna:
The default IllustrisTNG data file scaled the SFRD by the incorrect box volume for TNG100-1 (100^3 Mpc^3 instead of the correct 110.7^3 Mpc^3, [https://www.tng-project.org/about/](https://www.tng-project.org/about/)). This leads to about a factor of 1.36 shift in normalization of the TNG SFRD and rates calculated using it. 

I have rewritten the "SFR" key in the IllustrisTNG data file to remove the scaling by the box size (working with Max Briel to get the TNG data file updated in POSYDON_data), and moved the scaling step into the IllustrisTNG class in star_formation_history.py. This makes the scaling step more explicit and more flexible for users to use it with other IllustrisTNG boxes or different cosmology assumptions. I also added the default TNG100-1 values for calculating the box volume into the data file as well, but these could be added as parameters to the class instead.

Happy to add/change any comments or data file column names for more clarity! I added extra comments to make it more clear what each data column is and what each step does in the scaling. 

IllustrisTNG_SFRD_comparison
IllustrisTNG_SFRD_comparison_diff


sgossage:
In order to pass the unit tests, we should also update `mock_illustris_data` inside of `TestIllustrisTNG`, inside of `posydon/unit_tests/popsyn/test_star_formation_history.py`.

The mock data there needs to be modified to contain the new `BoxSFR` key. It could be the same sort of mock data that `SFR` uses right now.

I.e., this dictionary needs a `BoxSFR` item:

```python
        ...
        mock_data = {
            "SFR": np.linspace(0.1, 1.0, num_redshifts)[::-1],  # SFR decreases with redshift
            "redshifts": np.linspace(0.0, 9.0, num_redshifts)[::-1],  # Redshifts from 0 to 9
            "mets": np.logspace(-4, -1, num_metallicities),  # Metallicities from 1e-4 to 1e-1
            "M": np.ones((num_redshifts, num_metallicities))  # Equal mass in all bins for simplicity
        }
        ...
```

sashalvna:
Updated the tests to match, and on my side the star formation history tests all pass. Let me know if there's anything else I should change.

sgossage:
LGTM

We should have a strategy in place upon release for the new data release and any relevant notifications to users (if necessary).

sgossage (approved):

DavidRAguileraDena (approved):
I am not particularly knowledgeable in cosmology, but changes LGTM! Also, I found this useful/interesting: https://www.tng-project.org/data/forum/topic/674/which-volume-should-i-use/

PR #926, opened 2026-09-11
Update Bondi-Hoyle implementation

### LINKED ISSUE #829: Incomplete calculation of the Bondi-Hoyle accretion rate

### LINKED ISSUE #888: Audit and fix legacy bondi_hoyle implementation bugs
## Context

While extracting reusable XRB calculations for #881, several suspicious behaviors were identified in `posydon.utils.common_functions.bondi_hoyle`. They should be investigated and fixed in a dedicated PR rather than folded into the new-module PR.

## Items to investigate

1. A variable named `he_core_mass` is populated from `he_core_radius` and then implicitly converted to a Boolean.
2. The slow-wind assignment for extended H-rich giants (`beta = 0.125`) is followed by mass-based assignments that appear to overwrite it for every valid positive stellar mass.
3. In the `Kudritzki+2000` branch, `f_m` is assigned as a scalar inside a loop, so an array calculation can apply the last system's factor to all systems.
4. An unsupported wind-scheme string can reach the calculation with `f_m` undefined instead of raising a clear `ValueError`.
5. The random single-phase treatment and the appropriate orbit-averaged alternative require the physics review already discussed in #829.
6. The wind-disc criterion mixes velocities calculated in SI with `const.clight` in cgs through an implicit `0.01` conversion, which should be made explicit and tested.

## Acceptance criteria

- Confirm the intended prescriptions from the cited literature before changing behavior.
- Separate definite implementation bugs from choices of physical approximation.
- Add scalar and vector regression tests for each confirmed bug.
- Test H-rich giants, He-rich donors, mixed-temperature arrays, invalid schemes, circular/eccentric systems, and wind-disc outcomes.
- Keep the repair separate from #881.

sgossage:
- Fix assignment of Hurley+2002 `beta` for giants. Subsequent mass-based assignments would overwrite the value given to giants. So, now we apply the value to giants last.
- Fix `f_m` assignment in Kudritski+2000 scheme. Previously we assigned `f_m` as a scalar, meaning only one value was used throughout, when it should have been an array.
- Now we raise a `ValueError` for an invalid scheme option.
- Fix mean motion calculation. It used stellar radius and not orbital separation (probably inconsequential, since it was only used to sample random orbital phases).
- Adds an `orbit_averaged` calculation that can be enabled via a bool of the same name supplied to the function based on Bondi & Hoyle 1944. Should still implement an orbit averaged calculation for the disk criterion.

At least two outstanding issues (may or may not be very severe, but worth considering):

1. The disk criterion remains instantaneously calculated and tied to randomly sampled orbital positions, so it inconsistent with the orbit averaged calculation. We still rely on and RNG until we produce an orbit averaged disk criterion.
2. The disk criterion comes from [Sen et al. 2001](https://ui.adsabs.harvard.edu/abs/2021A%26A...652A.138S/abstract) and is derived assuming a black hole accretor and also that the wind speed greatly exceeds the orbital speed. These assumptions limit the applicability of the disk criterion to those regimes. Ideally, we generalize this. Or, we limit this criterion to the appropriate regimes explicitly.


sgossage:
A potential issue with the current implementation:

```python
    # eq. 10 in Sen, K. ,Xu, X. -T., Langer, N., El Mellah, I. , Schurmann, C.,
    # Quast, M., 2021, A&A
    if wind_disk_criteria:      # check if accretion disk will form
        eta = 1.0/3.0           # wind accretion efficiency between 1 and 1/3
        gamma = 1.0             # for non-spinning BH
        q = m / m_acc
        rdisk_div_risco = (
            (2/3) * (eta / (1 + q)) ** 2
            * (v / clight) ** (-2)
            * (1 + (v_wind / v) ** 2) ** (-4) * gamma ** (-1))
        for i in range(len(rdisk_div_risco)):
            if rdisk_div_risco[i] <= 1:         # No disk formed
                mdot_acc[i] = 10**-99.0
```

We run with `wind_disk_crtieria =True` always, and the calculation there only seems appropriate for BHs (it assumes `gamma = 1` and is based on ISCO. For a WD for example, this may always suppress accretion. Really, for WDs and neutron stars, we'd probably want something more like the radius of the compact object.

We should have a switch here to handle different CO types.

Before we get to where ISCO, etc. is involved, using the circularization radius as the disk radius (as in Sen et al.) we could write

$$\rm R_{circ}​=\frac{j^2}{G M_{acc​}}​$$

and

$$\rm j=\frac{1}{2}\eta\Omega_{orb}R_{acc}^2​$$

with 

$$\rm R_{acc}​=\frac{​2GM_{acc}}{v_{rel}^2}​​$$

so

$$\rm R_{circ}=4\eta^2\frac{G^3M^3_{acc}\Omega^2_{orb}}{v_{rel}^8}$$

Then, divide by the appropriate R depending on CO type. (To get something analytical like the orbit averaged equation from Bondi & Hoyle 1944, we could expand this in $e$ and do the same.)

sgossage:
also the sign in `v_rel` looks wrong:

```python
    v_rel = np.sqrt(v**2 + v_wind**2 + 2 * v * v_wind * k)  
```

but this should be from `v_rel = v - v_wind`, which would have `v_rel = np.sqrt(v**2 + v_wind**2 - 2 * v * v_wind * k)`.

posydon/utils/common_functions.py
```diff
@@ -642,10 +660,17 @@ def bondi_hoyle(binary, accretor, donor, idx=-1, wind_disk_criteria=True,
...
 
     # Bondi, H., & Hoyle, F. 1944, MNRAS, 104, 273
-    mdot_acc = alpha * ((G * m_acc * Msun)**2
+    if orbit_averaged:
+        mdot_acc = alpha / (2 * np.sqrt(1 - ecc**2))
+        mdot_acc *= ( (G * m_acc * Msun) / (sep * Rsun * v_wind**2) )**2
+        mdot_acc *= ( 1 + (G * (m_acc + m) * Msun) / (sep * Rsun * v_wind**2) )**(-3/2)
+        mdot_acc *= 10**lg_mdot
+    # instantaneous calculation randomly sampled around orbit
+    else:
```
TheLuckyPie, line 670: Shouldn't the following code (corresponding to lines 646-663) be placed under the else: command? If orbit_averaged: then the other calculations are unnecessarily performed?


    n = np.sqrt((G * (m_acc + m) * Msun) / ((sep * Rsun)**3))
    # random orbital period draws
    t0 = RNG.random(len(sep)) * 2 * np.pi / n
    # solve Kepler's equation for eccentric anomaly E
    E = newton(lambda x: x - ecc * np.sin(x) - n * t0,
               np.ones_like(sep) * np.pi / 2,
               maxiter=100)
    b = sep * Rsun * np.sqrt(1 - ecc**2)
    r_vec = np.array([sep * Rsun * (np.cos(E) - ecc), b * np.sin(E)])
    v_dir = np.array([-sep * Rsun * np.sin(E), b * np.cos(E)])
    r = np.linalg.norm(r_vec, axis=0)
    v_dir_norm = np.linalg.norm(v_dir, axis=0)

    k = np.einsum('ij,ij->j', r_vec, v_dir) / (r * v_dir_norm)  # cos(angle)
    v = np.sqrt(G * (m + m_acc) * Msun * ((2 / r) - (1 / (sep * Rsun))))  # m/s
    v_rel = np.sqrt(v**2 + v_wind**2 + 2 * v * v_wind * k)                # m/s
    v_rel = np.sqrt(v**2 + v_wind**2 - 2 * v * v_wind * k)                # m/s`

sgossage: Hey @TheLuckyPie, it should ideally be in the `else` block.

The reason it's not there yet is because when we calculate `wind_disk_criteria` below, there's still a reliance on `v`, which is randomly sampled still (relies on `r` and so also on `E`). So, a bunch of that stuff still needs to be calculated regardless (just not `v_rel` and the stuff specific to it).

However, ideally, we can create an orbit averaged version of the `wind_disk_criteria` too and separate the randomly sampled branch from the deterministic fully.

For now, I've gone ahead and moved what can be moved into the `else` block.

TheLuckyPie: Thanks for your answer. 

I also wanted to ask how often `wind_disk_criteria == False`? If there are many such cases, then what do you think about just putting the extra calculations under the following if condition just before the 
`if orbit_averaged:` block? Are the calculations performed here computationally expensive such that the code would benefit non-negligibly with this extra if statement?

``` 
if (not orbit_averaged) or wind_disk_criteria:
        # mean motion
        n = np.sqrt((G * (m_acc + m) * Msun) / ((sep * Rsun)**3))
        # random orbital period draws
        t0 = RNG.random(len(sep)) * 2 * np.pi / n
        # solve Kepler's equation for eccentric anomaly E
        E = newton(lambda x: x - ecc * np.sin(x) - n * t0,
                np.ones_like(sep) * np.pi / 2,
                maxiter=100)

        b = sep * Rsun * np.sqrt(1 - ecc**2)
        r_vec = np.array([sep * Rsun * (np.cos(E) - ecc), b * np.sin(E)])
        r = np.linalg.norm(r_vec, axis=0)
        v = np.sqrt(G * (m + m_acc) * Msun * ((2 / r) - (1 / (sep * Rsun))))  # m/s```

File: posydon/utils/common_functions.py [hunk 2]
```diff
@@ -628,24 +643,35 @@ def bondi_hoyle(binary, accretor, donor, idx=-1, wind_disk_criteria=True,
...
-    v_dir_norm = np.linalg.norm(v_dir, axis=0)
-
-    k = np.einsum('ij,ij->j', r_vec, v_dir) / (r * v_dir_norm)  # cos(angle)
     v = np.sqrt(G * (m + m_acc) * Msun * ((2 / r) - (1 / (sep * Rsun))))  # m/s
-    v_rel = np.sqrt(v**2 + v_wind**2 + 2 * v * v_wind * k)                # m/s
+
 
     # Bondi, H., & Hoyle, F. 1944, MNRAS, 104, 273
-    mdot_acc = alpha * ((G * m_acc * Msun)**2
+    if orbit_averaged:
```
TheLuckyPie, line 662: Furthermore, from my own searching, I found that this formulation is similar to:

Grzegorz Wiktorowicz et al 2021 ApJ 918 60: https://iopscience.iop.org/article/10.3847/1538-4357/ac0cf7
Equation (2), just without the $v_\text{orb.}$ term being expanded...

I don't know if this code's calculation understandability would be improved by citing this paper below the Bondi line and allowing the reader to find the exact equation used? 

Otherwise, I think the implementation is mathematically sound so I will approve it after hearing your response. Thanks.

posydon/utils/common_functions.py
sgossage: Right now we always call with `wind_disk_criteria == True`.

Another issue with us calling it all the time though is that the current criteria (from Sen et al. 2021) is specifically for accreting black holes, yet we apply it to NS and WD.

So, if we really want `wind_disk_criteria == True` always, then we should generalize the criteria to other compact objects. Besides deriving or finding an orbit averaged disk criteria, that would involve using a different radius than R_ISCO (e.g., just using the WD or NS radius maybe).

Turning off the `wind_disk_criteria` would be one way to bypass these complexities though, and I think if we do that, putting it all away into an `if` block like that sounds good.

The downside of going with that bypass is that we potentially lose some physical fidelity.

posydon/utils/common_functions.py
sgossage: Thanks, I think that will help with understandability for sure.

I got the equation from [Hurley et al. 2002](https://ui.adsabs.harvard.edu/abs/2002MNRAS.329..897H/abstract), so I've added a reference to that paper (eq. 6). Also [Boffin & Jorrisen 1988](https://ui.adsabs.harvard.edu/abs/1988A%26A...205..155B/abstract) eq. 6.

posydon/utils/common_functions.py
tassos25: @sgossage that is a good catch. The wind_disk_criteria criteria is **only** applicable to BHs as you said. For NSs for example you have a different mechanism of accretion where magnetic field are important and the disk criterion is not relevant. 

I don't think that the correct place to apply a "wind_disk_criteria" criterion is within the BHL mass accretion rate calculation. Rather, this should be applied at a later step by the user, to decide if that XRB, that is accreting at this rate, will be X-ray bright or not. Furthermore, this criterion is a bit outdated anyway. My suggestion is to remove this all together.

TheLuckyPie: @tassos25 @sgossage I have pushed a commit that removes all wind_disk_criteria components, including the calls for bondi_hoyle() in the rest of the code. I have also therefore moved unnecessary computation when `orbit_averaged == True` into the `else` block, and have updated the docstring to account for this change.

I also saw that in the test_common_functions.py script, there were a few 'assert' expressions that concern bondi_hoyle(). Will these need to be updated to account for this criteria's removal?

sgossage: thanks @TheLuckyPie , I went ahead and updated those unit test values. They are all different now because they're no longer zeroed by the disk criteria.

sgossage:
I did some testing of

1. The branch with `orbit_averaged = True`
2. The branch with `orbit_averaged = False` (instantaneous), but drawn 20,000 times and averaged. (Note: we only usually draw one random orbital position e.g., in `step_detached`, hence why `orbit_averaged = True` would be an improvement.)
3. versus an "exact" orbit average calculated by integrating the instantaneous formula using a grid of 1e6 points in eccentric anomaly, weighted by `1.0 - e*np.cos(E)`

You can see below that the orbit averaged case (closed form) starts to diverge from the others at high eccentricity:

image

It is off from the "exact" by about 24% at e  = 0.9.

It may be possible to do an integration (with fewer than 1 million points for example) and achieve higher accuracy than using the `orbit_averaged` formula without significant computational cost.

Issue #669, opened 2025-08-25
[BUG]: get-posydon-data "Killed" after downloading second .tar.gz file


berylha:
**Describe the bug**
When I run `get-posydon-data`, the code downloads the auxiliary files, then downloads `POSYDON_data_v2_grids_2Zsun.tar.gz`, then just states `Killed` and exits. I am instead downloading it all from [the Zenodo link](https://zenodo.org/records/15194708), which is working, but I wanted to send this in case it is an issue others may have too.

**To Reproduce**
I followed the steps given in the [Installation Guide](https://posydon.org/POSYDON/latest/getting-started/installation-guide.html).
I am trying to install POSYDON on Quest at Northwestern, and I ran the following:
```
conda create --name posydon_env python=3.11 -y
conda activate posydon_env
conda install -c posydon posydon
conda env config vars set PATH_TO_POSYDON=/home/oav5601/.conda/envs/posydon_env/lib/python3.11/site-packages/posydon
conda env config vars set PATH_TO_POSYDON_DATA=/projects/b1094/beryl/posydon_data
```
I then deactivated and re-activated the environment so the variables would be set, then ran
`get-posydon-data`, at which point I encountered the bug above.

**Expected behavior**
I had expected all of the DR2 data to be downloaded to my posydon_data folder.

**POSYDON Version:**
I am using POSYDON 2.0.0, installed with conda.

**System Configuration (please complete the following information):**
I am working on the Quest computing cluster at Northwestern, so the operating system is Red Hat Enterprise Linux 8.10.

sgossage:
Thanks @berylha, it could be that SLURM is killing the command to save resources. Is the `get-posydon-data` command being run on an interactive job (`srun`) or on a login node?

astroJeff:
Yes, the `killed` statement suggests this is a problem with Quest shutting down the download link, especially since the auxiliary files are being correctly downloaded. I've had problems like this in the past on clusters, where the connection will time out. One solution may be to run the command proceeded with `nohup` and with an `&` at the end:

`nohup get-posydon-data &`

This has the potential to let the download script run, even if your own connection to Quest times out. If the problem still persists, then it could have something to do with Quest wanting to limit the size of downloaded files (which makes no sense to me). @sgossage you haven't run into anything like this, right?

Just to check, are you running this on the head node?

sgossage:
[...]

Issue #930, opened 2026-09-15
[BUG]: Issues with get-posydon-data


TheLuckyPie:
**Describe the bug**
Running get-posydon-data resulted in a partially downloaded file, as the connection to Zenodo servers hosting the datasets was interrupted. This leaves behind a partially downloaded .tar.gz dataset file. This forces the user to restart the get-posydon-data, but due to line 159-160 in posydon/utils/data_download.py, an error is raised for an existing .tar.gz file, causing the script to always raise an error and crash. The only way to fix it is to manually delete the .tar.gz file.

Furthermore, re-running get-posydon-data after an interrupted download causes the _entire_ dataset to be redownloaded, instead of skipping the files that already exist.

The connection drop produces the following error:

> `File "/Users/anurag/.pyenv/versions/3.11.9/lib/python3.11/urllib/request.py", line 280, in urlretrieve raise ContentTooShortError(
urllib.error.ContentTooShortError: <urlopen error retrieval incomplete: got only 6064525812 out of 10241444651 bytes>`

And if the incomplete .tar.gz exists:

> `FileExistsError: POSYDON data already exists at /Users/anurag/Documents/POSYDON/posydon_data/POSYDON_data_DR2_grids_2Zsun.tar.gz.`

**To Reproduce**
Run get-posydon-data, and interrupt the script (ex. ctrl+c) such that the data folder contains the unzipped .tar.gz file for a dataset. Running get-posydon-data reproduces this error.

**Expected behavior**
If the connection to Zenodo is dropped (raising the ContentTooShortError), instead of stopping the script completely, the script should re-try the download.

If any error is raised, such as the FileExistsError, it should delete the partially downloaded .tar.gz file or give the user the option to delete/rename/skip instead of stopping the script completely.

Some behavior should exist to determine if a dataset already exists, skipping it when re-running get-posydon-data, or giving the user the option to re-download it.

**POSYDON Version:**
V2.3.2

**System Configuration:**
MacOS

PR #931, opened 2026-09-17
Fix data download crash

DavidRAguileraDena:
I experienced an intermittent crash during data download using using `get-posydon-data`. I got a cryptic "_Killed_" message  after the file was apparently done downloading, but not yet decompressed.

 I checked my computer's `journalctl` to find the cause. The error message was `Out of memory: Killed process (get-posydon-dat)`.

After some digging, I think the problem is that the MD5 check was loading the entire 10ish GB file at once. I modified `get-posydon-data` to split the MD5 check into 8MB chunks, and the problem seems to be fixed on my side.

This potentially fixes the problem described in [discussion #754](https://github.com/POSYDON-code/POSYDON/discussions/754), and [bug report 669](https://github.com/POSYDON-code/POSYDON/issues/669).


tassos25:
This is a useful addition to deal with big files, however I do not think it addresses issue #669 . I believe the the point there is that a failure to fully download of do the checksum for one file should don't kill the script, but stop more gracefully with options to overwrite or skip or whatever relevant the existing (partially downloaded) files.

sgossage:
I think this should address the core issue of issue [#669](https://github.com/POSYDON-code/POSYDON/issues/669). Although, it was not fully traced back in Beryl's case whether the command was killed there due specifically to the md5 checksum. It was likely due to OOM, as the command there was performed on a login node with limited computational resources (critically, RAM), and it the diagnosis that the checksum loads 10 GB into RAM would check out there.

This PR combined with [PR#904](https://github.com/POSYDON-code/POSYDON/pull/904) should build towards a more graceful handling of data download errors. That PR checks for and skips pre-existing data sets.

All together, I think this nearly resolves issue [#930](https://github.com/POSYDON-code/POSYDON/issues/930) too. The only missing thing would be the requested automatic retry in case the connection to Zenodo is dropped.

